PR #4, opened 2017-10-04
FOLIO-854 update readme for rename

dcrossleyau:
Follow rename stripes-util-notes to stripes-smart-components


dcrossleyau:
I gathered that, so relieving one task.

PR #9, opened 2017-10-11
Release v1.0.1

zburke:
Update `package.json` details from util-notes to stripes-smart-components. 

No, there wasn't a v1.0.0. Don't ask.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change history for stripes-util-notes
 
+## 1.0.1 (https://github.com/folio-org/stripes-smart-components/tree/v1.0.1) (2017-10-11)
+[Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v0.3.0...v1.0.1)
+
+* Change NPM module name to stripes-smart-components. Fixes STSMACOM-7.
+
 ## 0.3.0 (https://github.com/folio-org/stripes-util-notes/tree/v0.3.0) (2017-10-11)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v0.2.0...v0.3.0)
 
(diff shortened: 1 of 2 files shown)


MikeTaylor:
Though now I come to look more closely at the package file, `"type": "app"` is definitely wrong.
"

zburke:
> `"type": "app"` is definitely wrong.

Yes, but IIRC removing it altogether had other bad repercussions. Let me see if I can dig up the conversation.

zburke:
From @ttamj a few weeks back: https://folio-project.slack.com/archives/C210UCHQ9/p1506540609000400: 
> next up to a working build is including a stripes "type" property in the package.json for notes.  This is currently causing an assertion error in stripes-loader. 

@wafschneider and I encountered this at the beginning of the folio-testing Troubles at the end of September. Adding the `type` key allowed the build to proceed, so we didn't question it. I'll follow up with them both. `stripes-components` has no such key, so it must be possible to do without it.

MikeTaylor:
Right. There is some confusion here. I do remember that at one point people thought Notes would be a Stripes Module rather than merely an NPM package, so maybe this is a holdover from that. At any rate, it's none of stripes-loader's business what a non-module package may or may not elect to put in its package file.

ttamj:
Agreed.  Now that notes is a smart component, it shouldn't need a type property in `package.json`.  The loader logic expected a type only for modules specified in a platform's `stripes.config.js` tenant configuration.

zburke:
I'll remove the errant `"type": "app"` from `package.json` in release v1.0.2 since doing so within this PR would put this tag out of sync with the NPM package which has already been released. Ostensibly, this PR was going to be a formality as GitHub and NPM releases should always happen in lock step, but @MikeTaylor's sharp eyes caught the bug.

PR #14, opened 2017-10-30
adding EntryManager readme

zburke:
documentation!

### DIFF
--- a/lib/EntryManager/readme.md
+++ b/lib/EntryManager/readme.md
@@ -1,0 +1,26 @@
+# EntryManager
+Component for handling CRUD workflow
+
+## Description
+Many objects are managed in a workflow that includes a listing page, a
+detail page, and an add/edit page that also includes delete functionality.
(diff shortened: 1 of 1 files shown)


MikeTaylor:
Great that you've added this!

But GitHub is not properly formatting your table: see https://github.com/folio-org/stripes-smart-components/blob/ad238877a0b99ff1efb0e52fd4e9d2f2670cd6e9/lib/EntryManager/readme.md

Are you using the wrong Markdown flavour?

zburke:
Fixed; thanks.

MikeTaylor:
Beautiful!

Seeing this now for the first time, it occurs to me that we could simplify the API somewhat;
* `location` can be obtained using React Router's `withRouter` HOC.
* `history` likewise, I think.
* `stripes` can be taken from the context, as `<IfPermission>` does.

Good call on the example for `nameKey`, BTW. It would be pretty much impossible to write a comprehensible description of what this is, but the example makes it super-clear.

PR #15, opened 2017-10-30
simpler API. Fixes STSMACOM-13

zburke:
Implementing @mike's very sensible suggestions on #14: 
* `location` can be obtained using React Router's `withRouter` HOC.
* `history` likewise, I think.
* `stripes` can be taken from the context, as `<IfPermission>` does.


zburke:
Yes, `location` and `history` should remain as props, and when the CI ran `eslint` it caught the error. Nice!

PR #25, opened 2017-11-13
first draft of GitHub style @at-mention autocompletion. Refs STSMACOM-4

zburke:


zburke:
`MentionWrapper` doesn't actually handle `component={CustomComponent}`, it just uses a plain old `textarea`, but we need to use a `@folio/stripes-components/lib/TextArea`. So, this isn't actually functional at present as the current implementation provides the @mention functionality but does't allow the `<Note>` to be saved.

PR #26, opened 2017-11-13
facilitate additional fields in editable lists that are reactive components

wwelling:


wwelling:
depends on https://github.com/folio-org/stripes-components/pull/87

wwelling:
@MikeTaylor I believe I added you as a review only after I made the comment change commit. I also wanted review on the delete of the readonly metadata property and ensure additionalFields was acceptable to pipe through to a stripes-component component.

MikeTaylor:
I see, that makes more sense. For some reason, GitHub's "Files changed: 1" link was (at that time) only showing me the last commit, https://github.com/folio-org/stripes-smart-components/pull/26/commits/601d796263d880ca3901a4cd8f413234298ab4b5

I have no idea why, and what has now changed.

Anyway, I think this PR is fine. Go ahead and merge.

wwelling:
@MikeTaylor thanks for the review. The majority of the feature was fleshed out in the stripes-component and used in ui-users.

PR #31, opened 2017-11-23
Add generic `<SearchAndSort>` component. Fixes STSMACOM-21.

MikeTaylor:


MikeTaylor:
@JohnC-80 What is the deal with the lint error "Interactive elements should not be assigned non-interactive roles jsx-a11y/no-interactive-element-to-noninteractive-role"?

MikeTaylor:
The other two lint errors are addressed here: https://github.com/folio-org/eslint-config-stripes/pull/9

JohnC-80:
@MikeTaylor, seems it doesn't like the listItem role assigned to the anchor... We can look at wrapping these anchors within something else (a div with the listItem role, perhaps) but I imagine some tweaking of styles will be necessary for that. I can create an issue and take a look.

PR #34, opened 2017-11-23
Status of notes pane is now reflected in the URL

MikeTaylor:
The `notes` part of the query may be absent for no notes pane, `true`
to display it, or the UUID of a specific note.

Fixes STSMACOM-19.


MikeTaylor:
Hope this is OK, @zburke. I saw immediately how to do this through the magic of STRIPES-452, and it seemed churlish not to just do it.

PR #41, opened 2017-11-29
Stsmacom 31 react a11y lint

JohnC-80:
@MikeTaylor This might look odd at first pull until we get the text-color sorted out in the MCL rows (currently undergoing CSS overhaul) - but it fixes the lint issue.


MikeTaylor:
Thanks, @JohnC-80, appreciated!

PR #49, opened 2017-12-06
Added 'contentIsList' prop to results Pane

rasmuswoelk:
Added a single boolean prop that removes padding from PaneContent which is only relevant if the content is a MultiColumnList.

The update to the Pane itself will be published soon as well.


rasmuswoelk:
I'll submit an updated pull request for this.

PR #54, opened 2017-12-12
Search and Sort: Conditional render of detail pain

jeremythuff:
Resolves UIREQ-38:

This places logic in the render method of search and sort which conditions the inclusion of the details pain on the absence of `this.props.onSelectRow`. Now, if an embedded instance of search and sort has had the behavior of onSelectRow overwritten, the detail pane will not be initialized, and therefore not incur the errors that have been reported.


MikeTaylor:
LGTM

PR #62, opened 2017-12-19
Updating search and sort component to fit PaneHeader changes

rasmuswoelk:
In this pull request I have primarily been focusing on fitting the SearchAndSort component to use the new PaneHeader.

I also added a search icon in the results-pane which isn't active yet but should be in the future:

image

For now I have left the search bar in the PaneHeader but this will be deprecated in the future and replaced with a search field in the search/filter pane content like so:

image

(See http://ux.folio.org/prototype/en/inventory)


cherewaty:
This would be cleaner all squashed down into a single git commit.

PR #71, opened 2018-01-08
SearchAndSort: Added ability to toggle visibility of Filters pane

doytch:
- More work for [STCOM-164](https://issues.folio.org/browse/STCOM-164). 
- The magnifying glass in the header of the search results pane toggles the visibility of the filters/search pane. 
- When the pane is closed, a badge appears on the magnifying glass icon showing the number of active filters.

No animations on this pass per discussion on Slack.

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -354,7 +354,7 @@ class SearchAndSort extends React.Component {
 
     const { filterPaneIsVisible } = this.state;
     const filterCount = Object.keys(filters).length;
-    const toggleFilterPaneMessage = `${filterPaneIsVisible ? 'Hide' : 'Show'} Search and Filters pane.\n\n${filterCount} applied filter${filterCount > 1 ? 's' : ''}.`;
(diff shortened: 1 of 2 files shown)


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -392,22 +406,25 @@ class SearchAndSort extends React.Component {
...
-          <FilterGroups
-            config={filterConfig}
-            filters={filters}
-            onChangeFilter={this.onChangeFilter}
-            onClearFilter={this.onClearFilter}
-            onClearAllFilters={this.onClearAllFilters}
-            disableNames={disableFilters}
-          />
-        </Pane>
+        { filterPaneIsVisible ?
```
cherewaty: This requires completely re-rendering the filters when toggling it to visible. Would it be preferable to `display:none` or even just layer it underneath the results pane?

doytch: Thanks and I see your point. But given that this isn't something that redraws often, and will never be part of a change-state-redraw-change-state-redraw loop, I don't think it's a big issue. Profiling it (even in webpack dev server) didn't seem to show any red flags.

My argument for writing it this way is that it's understand and it's more idiomatically Reactish.

PR #88, opened 2018-02-01
Prevents EditableListForm from re-rendering before confirming success

aditya-matukumalli:
Prevents EditableListForm from re-rendering before confirming success from the backend


JohnC-80:
@MikeTaylor Since <EditableList> now accepts promises for its deletion and save actions, the parent components that make use of it have been updated to feed it functions that return promises.

PR #92, opened 2018-02-26
STCOM-164: Move SearchAndSort search field into an accordion

doytch:
- Moved the SearchField at the top of the Search & Filter pane into an accordion.
- The clear X by the accordion title resets the query index and the query.
- The clear X inside the SearchField resets just the query.
- Changed the top button's label back to "Reset all"

- Styling on that button has **not** been changed from the `link` styling. The various changes in the Button component re:styling have made getting back to the previous styling (grey with no border) not possible (AFAIK) without adding more CSS styles, which I thought we were trying to avoid. @JohnC-80 and @rasmuswoelk feel free to advise otherwise.

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.css
+++ b/lib/SearchAndSort/SearchAndSort.css
@@ -1,0 +1,9 @@
+@import '@folio/stripes-components/lib/variables.css';
+
+.resetButton {
+    background-color: transparent;
+    border: 1px solid transparent;
+    cursor: pointer;
+    color: var(--secondary);
+    padding: 4px;
+}
--- a/lib/SearchAndSort/SearchAndSort.css
+++ b/lib/SearchAndSort/SearchAndSort.css
@@ -1,9 +1,0 @@
-@import '@folio/stripes-components/lib/variables.css';
-
-.resetButton {
-    background-color: transparent;
-    border: 1px solid transparent;
-    cursor: pointer;
-    color: var(--secondary);
-    padding: 4px;
-}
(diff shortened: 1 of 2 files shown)


JohnC-80:
Thanks @doytch! I don't mind the blue text, myself. No high-powered opinions here, but it does set the action apart from the accordion headers and is consistent with the rare other text-only, borderless, background-less buttons that are around the UI. I think we'd definitely need a more compelling case to add something new. @filipjakobsen feel any different about it?

filipjakobsen:
Hi @JohnC-80.

I agree with you that this one button does not warrant building a component for. However, I spent a long time to design it in such a way that it does not steal too much attention from the rest of the layout, without becoming too inconspicuous, so it's very important for me that we make this particular button look the way I designed it. It's a special and unique use case and as such, we should just style it specifically rather than changing anything in Stripes Components.

CC @rasmuswoelk

JohnC-80:
@doytch it's possible to pass `buttonStyle="none"` to get base-line styling without any of the color/border changes it will render contents-only, no background, no border, and text in a normal weight. Then all that's left is to set the font weight to bold via inline styling...`style={{fontWeight: 'bold'}}` @filipjakobsen if there are multiple occurrences of this kind of button in designs, then I absolutely do recommend we create a built-in class.

filipjakobsen:
@JohnC-80, I agree. Currently this is a one-of-a-kind styling.

doytch:
Right on. Was trying to avoid inline-styles in `0e2dc09` since I thought they were persona non grata in the codebase but just popped in a second commit using your suggestion, @JohnC-80.

PR #100, opened 2018-03-06
Optionally derive some props from props.packageInfo

zburke:
Many values used in SearchAndSort are derived from values in the calling
module's `package.json` file. This update adds support for passing in
those values (via `props.packageInfo`) and deriving them, rather than
repeating that code in every single module that calls SearchAndSort.

Refs [STSMACOM-64](https://issues.folio.org/browse/STSMACOM-64)


MikeTaylor:
These ESLint errors are idiotic. Let's add `ignoreEOLComments` to eslint-config-stripes. See https://eslint.org/docs/rules/no-multi-spaces

PR #102, opened 2018-03-07
require props.packageInfo. Refs STSMACOM-64

zburke:
Now that all packages using `<SearchAndSort>` pass it their `packageInfo` values, we can deprecate the old props that were just derived from `packageInfo` before passing them as props. 

Refs [STSMACOM-64](https://issues.folio.org/browse/STSMACOM-64).


zburke:
Easy enough to revert and reimplement later; worth it?

MikeTaylor:
Well, I would have left at least a week between updating the apps and removing the support for the old-style API. But we'll see if you get away with it.

PR #126, opened 2018-03-27
STSMACOM-76

szkata:
Update EntryManager to use translations instead of hardcoded strings.

### DIFF
--- a/lib/EntryManager/EntryForm.js
+++ b/lib/EntryManager/EntryForm.js
@@ -74,10 +74,10 @@ class EntryForm extends React.Component {
         <Button
           id="clickable-save-entry"
           type="submit"
-          title={this.props.stripes.intl.formatMessage({id: "stripes-core.button.save"})}
+          title={this.props.stripes.intl.formatMessage({ id: 'stripes-core.button.save' })}
           disabled={(pristine || submitting)}
           marginBottom0
-        >{this.props.stripes.intl.formatMessage({id: "stripes-core.button.save"})}
+        >{this.props.stripes.intl.formatMessage({ id: 'stripes-core.button.save' })}
         </Button>
       </PaneMenu>
     );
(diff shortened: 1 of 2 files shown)


lib/EntryManager/EntryForm.js
```diff
@@ -113,11 +121,12 @@ class EntryForm extends React.Component {
             {selectedEntry &&
               <ConfirmationModal
                 open={confirmDelete}
-                heading={`Delete ${this.props.entryLabel}?`}
-                message={(<span><strong>{selectedEntry[this.props.nameKey] || `Untitled ${this.props.entryLabel}`}</strong> will be <strong>removed</strong>.</span>)}
+                heading={deleteButtonText}
```
doytch: We lost the `<strong>` formatting here.

szkata: So should I split up the translation to the sentence, the entry and the verb and apply the `<strong>` formatting in the js code?

doytch: Yeah that's how we're moving forward for the time being. It's not pretty, but it's the best solution we've got to situations like this.

rjruizes:
@szkata Can you make the requested changes so we can merge? I would like to use the changes you've made.

zburke:
Hang on. The discussion on #stripes today led to [STCOR-182](https://issues.folio.org/browse/STCOR-182) and [STRIPES-518](https://issues.folio.org/browse/STRIPES-518), which says the opposite: _don't_ break up strings and stash HTML in the translated values via `<FormattedHTMLMessage>`. 

I wouldn't bother refactoring the translations into fragments because that's work we'll just have to undo later. We could just merge this as-is and know that we're going to have to do a bit of refactoring when STRIPES-518 is finally ready, or we could hold of until STRIPES-518 is ready. Either is fine; your call.

MikeTaylor:
Agree with @zburke.

lib/EntryManager/EntryForm.js
ibendermonguz: As you also read on #stripes Slack channel we probably will have a safeFormattedHTMLMessage component which will allows using HTML elements in translations and will sanitize them. I would recommend to move the `<strong>` formatter in the translation and use the mentioned function when it will be available. Thoughts?

ibendermonguz:
@zburke: Created an issue not to forget about it: https://issues.folio.org/browse/STSMACOM-78

zburke:
Refs [STSMACOM-76](https://issues.folio.org/browse/STSMACOM-76).

PR #128, opened 2018-03-27
pass validate() as prop

rjruizes:
redux-form can accept `validate` as a prop on the the FormComponent, or as a stand-alone function. To have access to `this.props.stripes`, it needs to be the former. This PR enables that.


zburke:
Ah, nice, I didn't know about passing the `validate` function as a prop. I also ran into the problem of adding translation substitution strings there and ended up just using `<FormattedMessage>`, but this offers more flexibility.

PR #130, opened 2018-03-29
Added a more user friendly empty message

rasmuswoelk:
This PR adds a NoResultsMessage component that displays a message to the user depending on entered or missing search query and selected filters. It also shows a loading indicator when results are fetching as well as any request errors. Pretty much the same messages like before but a bit beautified.

This hopefully removes some confusion for users when interacting with the SearchAndSort - especially on large screens in the cases where the list has no results as default (and no message telling the user why) which left a large blank area until any filters were added.

**Note:** The wording might need to be changed.

## Screenshots

**Initial view on larger screens - showing the user that there are filters available**
image

**If the filter pane should be closed there's an easy and noticeable way of opening it again**
image

**No results found by search query**
image

**No results found with filters (no search query)**
image

**Loading indicator**
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 node_modules
 yarn.lock
+.DS_Store
+yarn-error.log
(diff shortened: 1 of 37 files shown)


rasmuswoelk:
This PR requires an update in stripes-components (PR is submitted) in order to work (if you want to test it).

PR #168, opened 2018-05-04
ControlledVocab: Switched from fill-available to stretch

doytch:
Just supporting the eventual value for the behaviour we want here. Doesn't really do anything currently, since only prefixed values are used by browsers.


funkymalc:
Sorry,  Mark - You were the unwitting guinea pig of a recent change to perform mandatory stripes platform build as part of the PR.   The webpack is currently broken due to an issue with the vendors module.   See latest comments on the coreteam channel on Slack.

funkymalc:
re-ran build for this PR after fixing issue with ui-vendors.  All good.

PR #179, opened 2018-05-30
ChangeDueDateDialog and LoanList components

doytch:
This will all be used for [UIU-497](https://issues.folio.org/browse/UIU-497).

- `ChangeDueDateDialog` is the entry point into this set of components.
- Keeps track of success/failure and renders either the "editing" or "success" view in its `<Modal>`.
- `ChangeDueDate` renders the `DueDatePicker` and new `LoanList` component, and manages the actual changing of the due date on the backend.
- `LoanList` lets me DRY the code a bit since the editing/success views render similar but different loan lists. It also manages the fetching of the loan policies since only the policy ID is in the loan objects.


lib/ChangeDueDateDialog/DueDatePickerForm.js
```diff
@@ -0,0 +1,42 @@
...
+            timezone="UTC"
+            {...timeProps}
+          />
+        </Col>
+      </Row>
+    );
+  }
+}
+
+export default reduxForm({
```
mkuklis: @doytch do we care about showing the "unsaved changes" popup in cases when somebody will want to leave the due date dialog before saving? If no then this is fine. If we do then we may need to replace this with `stripes-form`.

lib/ChangeDueDateDialog/ChangeDueDate.js
```diff
@@ -0,0 +1,294 @@
...
+          <Button buttonStyle="primary" onClick={this.changeDueDate}>
+            <FormattedMessage id="stripes-core.button.saveAndClose" />
+          </Button>
+        </Layout>
+      </div>
+    );
+  }
+}
+
+export default reduxForm({ form: 'cddd' })(ChangeDueDate);
```
mkuklis: @doytch I'm probably missing something here but do we need `reduxForm` here?

lib/ChangeDueDateDialog/DueDatePickerForm.js
doytch: Good catch. For whatever reason it never throws the "unsaved changes" popup for me anyway. 🤔 

Will change it anyway though! :D

lib/ChangeDueDateDialog/ChangeDueDate.js
doytch: Nope we don't. Refactoring artifact I missed.

lib/ChangeDueDateDialog/DueDatePickerForm.js
mkuklis: Cool. I think the `navigationCheck` has to be set to `true`.

PR #181, opened 2018-06-01
Remove @ mentions from Notes

cherewaty:
## Purpose
`@folio/react-githubish-mentions` has a questionable status on React 16.

## Approach
Since Notes is currently non-functional, @zburke suggested ripping out the use of `@folio/react-githubish-mentions` until a suitable replacement can be found/built: https://github.com/folio-org/react-githubish-mentions/pull/4

This replaces the previous at-mention-enabled `TextArea` with a bare `TextArea`. `AtMentionTextArea.js` will still be in the git history when Notes receives attention again.


cherewaty:
@zburke can you merge? (I don't have collab permissions on `stripes-smart-components`)

PR #193, opened 2018-06-13
Adding transitioning reset button to SearchAndSort

rasmuswoelk:
We have discussed how to make the reset button less annoying (can't come up with a better way of saying it).

A solution we came up with is to transition the button in and out but with a slight delay. This makes it possible to check/uncheck fast (or uncheck one checkbox and check another) without having the entire list shift position immediately which could cause misclicks and frustration for the user.

The component is a standalone component that could be used in other places in the application but for now, it lives within this repo.

## Before
jun-13-2018 15-39-21

## After
jun-13-2018 15-37-10

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.css
+++ b/lib/SearchAndSort/SearchAndSort.css
@@ -6,3 +6,8 @@
   .searchField {
     margin: 0 15px;
   }
+
+  /* reset button wrap - minor alignment adjustment */
+  .resetButtonWrap {
+    margin-left: -3px;
+  }
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -108,7 +108,7 @@ class SearchAndSort extends React.Component { // eslint-disable-line react/no-de
         other: PropTypes.shape({
           totalRecords: PropTypes.number.isRequired,
         }),
-        isPending: PropTypes.bool.isPending,
+        isPending: PropTypes.bool,
         successfulMutations: PropTypes.arrayOf(
           PropTypes.shape({
             record: PropTypes.shape({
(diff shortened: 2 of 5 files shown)


MikeTaylor:
I would so, so much rather have it just stay there. What do we gain by having it confusingly phase in and out of existence all the time?

doytch:
I feel like this is one of those scenarios where we think users interact with our applications with as much speed as we do as developers. I really don't see where someone would check and then uncheck something that quickly.

The use case I always think of here is where the user checks a filter, waits for the results, sees there are none (or too few), and then removes that filter. Think about the time to fetch the filtered results, display them, the user to parse and realise that there are no/few results, decide that they'll remove that filter, and finally clicking the mouse/keyboard again. 

That's a loooong process for most people, no? And the "Reset all" button will have appeared and shifted everything by then. And I don't think just increasing the delay is great because then the effect feels disconnected from the cause.

TL; DR - This is hard. I don't envy you.

rasmuswoelk:
@MikeTaylor The decision to have the button appear and disappear is not mine. I can't tell you exactly why we aren't doing as you are suggesting. One reason I could think of would be that we don't want unnecessary UI shown when it's not relevant for the user. I'm sure @filipjakobsen can come up with an even better explanation 🙂

@doytch You are _probably_ right that users won't interact with as much speed as developers but it's very likely that a user would misclick a checkbox and want to correct that error immediately. Without a delay, the layout shifts and that could cause an additional misclick. Switching between two checkboxes would also induce a shift. This can especially be a pain on touch devices.

Personally, I feel that this is a good and needed solution if we are going with a _disappearable reset-button_. If it's decided that the button should stay visible at all times then we of course wont be needing this additional delay/transition.

MikeTaylor:
@rasmuswoelk I hear what you're saying :-)

rasmuswoelk:
@JohnC-80 Thanks, John :) And you are correct - it's easy to switch back if it turns out to cause issues.

PR #201, opened 2018-06-25
Add support for param namespacing. Part of STCOM-300

mkuklis:
The purpose of this PR is to introduce a way for namespacing params used by `<SearchAndSort>` component. 

The implementation follows the document written by @MikeTaylor here:

https://github.com/folio-org/stripes-components/blob/master/util/parameterizing-makeQueryFunction.md

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -260,14 +260,9 @@ class SearchAndSort extends React.Component {
   }
 
   onChangeFilter = (e) => {
-    const { nsParams, parentMutator, initialResultCount, filterChangeCallback } = this.props;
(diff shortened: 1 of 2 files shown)


mkuklis:
Thank you for reviewing this @zburke. I will need to hold with this PR before we merge https://github.com/folio-org/stripes-components/pull/446

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -310,6 +314,12 @@ class SearchAndSort extends React.Component {
     this.transitionToParams({ layer: 'edit' });
   }
 
+  transitionToParams(values) {
```
MikeTaylor: Are we confident that this function binds `this` correctly?

PR #206, opened 2018-07-02
LocationInput: Fix onChange signature

doytch:
`Selection` doesn't [pass back the event, just the new value](https://github.com/folio-org/stripes-components/blob/master/lib/Selection/SingleSelect.js#L261).


mkuklis:
Thanks for catching this @doytch!

PR #226, opened 2018-08-14
UIU-516 Spike: Select a Password Strength Meter

oleksii-popov:
Added Password Strength component with readme.md
Added translations for EN
Added password strength plugin to package.json
Changelog updated correspondingly

### DIFF
--- a/lib/PasswordStrength/PasswordStrength.js
+++ b/lib/PasswordStrength/PasswordStrength.js
@@ -18,7 +18,7 @@ class PasswordStrength extends React.Component {
 
   componentDidUpdate(prevProps) {
     // in case password updated and provided
-    if (this.props.password !== prevProps.password && !!this.props.password) {
+    if (this.props.password && this.props.password !== prevProps.password) {
       const password = this.props.password;
       const result = this.strengthTester.check(password);
 
@@ -37,9 +37,8 @@ class PasswordStrength extends React.Component {
   }
 
   render() {
-    const password = this.props.password;
-    if (!password) {
-      return false;
+    if (!this.props.password) {
+      return null;
     }
 
     return (
--- a/lib/PasswordStrength/PasswordStrength.js
+++ b/lib/PasswordStrength/PasswordStrength.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import taiPasswordStrength from 'tai-password-strength';
+import { withStripes } from '@folio/stripes-core/src/StripesContext';
+
+class PasswordStrength extends React.Component {
+  static propTypes = {
+    dataSource: PropTypes.string.isRequired
+  };
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const password = this.props.dataSource;
+    if (!password) {
+      return false;
+    }
+
+    const strengthTester = new taiPasswordStrength.PasswordStrength();
+    strengthTester.addCommonPasswords(taiPasswordStrength.commonPasswords);
+    strengthTester.addTrigraphMap(taiPasswordStrength.trigraphs);
+    const result = strengthTester.check(password);
+    let output = 'weak';
+
+    switch (true) {
+      case result.strengthCode.indexOf('WEAK') >= 0:
+        output = 'weak';
+        break;
+      case result.strengthCode.indexOf('STRONG') >= 0:
+        output = 'strong';
+        break;
+      default:
+        output = 'reasonable';
+        break;
+    }
+
+    return (
+      <div>
+        <FormattedMessage id={'stripes-smart-components.password-strength.' + output}/>
+      </div>
+    );
+  }
+}
+
+export default withStripes(PasswordStrength);
--- a/lib/PasswordStrength/PasswordStrength.js
+++ b/lib/PasswordStrength/PasswordStrength.js
@@ -22,16 +22,12 @@ class PasswordStrength extends React.Component {
       const password = this.props.password;
       const result = this.strengthTester.check(password);
 
-      switch (true) {
-        case result.strengthCode.indexOf('WEAK') >= 0:
-          this.output = 'weak';
-          break;
-        case result.strengthCode.indexOf('STRONG') >= 0:
-          this.output = 'strong';
-          break;
-        default:
-          this.output = 'reasonable';
-          break;
+      if (result.strengthCode.indexOf('WEAK') >= 0) {
+        this.output = 'weak';
+      } else if (result.strengthCode.indexOf('STRONG') >= 0) {
+        this.output = 'strong';
+      } else {
+        this.output = 'reasonable';
       }
     }
   }
@@ -43,7 +39,7 @@ class PasswordStrength extends React.Component {
 
     return (
       <div>
-        <FormattedMessage id={'stripes-smart-components.password-strength.' + this.output} />
+        <FormattedMessage id={'stripes-smart-components.passwordStrength.' + this.output} />
       </div>
     );
   }
--- a/lib/PasswordStrength/PasswordStrength.js
+++ b/lib/PasswordStrength/PasswordStrength.js
@@ -2,47 +2,52 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
 import taiPasswordStrength from 'tai-password-strength';
-import { withStripes } from '@folio/stripes-core/src/StripesContext';
 
 class PasswordStrength extends React.Component {
   static propTypes = {
-    dataSource: PropTypes.string.isRequired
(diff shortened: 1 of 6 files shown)


lib/PasswordStrength/PasswordStrength.js
```diff
@@ -0,0 +1,48 @@
...
+
+    return (
+      <div>
+        <FormattedMessage id={'stripes-smart-components.password-strength.' + output}/>
+      </div>
+    );
+  }
+}
+
+export default withStripes(PasswordStrength);
```
cherewaty: It doesn't look like you're accessing anything on the `stripes` god object, so I don't think you need `withStripes`. You can just `export default PasswordStrength;`.

doytch:
Never really looked into this lib so I'm just firing in the dark, but is it necessary to re-create the strengthTester, add the common passwords and trigraph map in every render call? Could we just do this in the constructor once?

Also, theoretically, if we did do that, we could just run the `check()` calls in a `getDerivedStateFromProps` if the password has changed.

Finally, just to be pedantic, can we rename the prop to `password`, especially since you rename it as such in `render`? I'll never remember `dataSource` 😞

lib/PasswordStrength/PasswordStrength.js
oleksii-popov: @cherewaty  thanks for your comment. 
I'm new to stripes platform. And I did not find a way to pass stripes to this component other then <code>withStripes</code> (I found it in another module). If you will describe how to pass <code>stripes</code> to the component, I will really appreciate.

zburke: I think @cherewaty was suggesting that since you don't use `stripes` anywhere in this class, you don't need to wrap it with `withStripes`.

oleksii-popov: It is needed for translation strings.

oleksii-popov:
Added aria-live="polite"
Moved tester init and trigraphs to a constructor
Password strength check logic moved out of 'render'.
Removed 'withStripes'


@JohnC-80  @doytch please review
+ @VictorSoroka1 @TarasSpashchenko @OleksiiKuzminov @Igor-Gorchakov @natali-zaitseva

lib/PasswordStrength/PasswordStrength.js
oleksii-popov: Updated, please take a look. No more withStripes

oleksii-popov:
@MikeTaylor, @zburke can you please take a look.

lib/PasswordStrength/PasswordStrength.js
```diff
@@ -0,0 +1,53 @@
...
+    this.output = 'weak'; // set default password hint to weak
+  }
+
+  componentDidUpdate(prevProps) {
+    // in case password updated and provided
+    if (this.props.password !== prevProps.password && !!this.props.password) {
+      const password = this.props.password;
+      const result = this.strengthTester.check(password);
+
+      switch (true) {
```
MikeTaylor: The `switch (true)` definitely falls into the "too clever" category.

> "Everyone knows that debugging is twice as hard at writing a program in the first place.  So if you're as clever as you can be be when you write it, how will you ever debug it?" -- Kernighan & Pike, _The Elements of Programming Style_.

Please rephrase this as a regular sequence of `if`s such as programmers have been happily using since 1958.

lib/PasswordStrength/PasswordStrength.js
```diff
@@ -0,0 +1,53 @@
...
+          this.output = 'reasonable';
+          break;
+      }
+    }
+  }
+
+  render() {
+    const password = this.props.password;
+    if (!password) {
+      return false;
```
MikeTaylor: Does React use the `false` return-value? If not, it's more honest just to `return`.

lib/PasswordStrength/readme.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Introduction
+
+It is POC for password strength component. It requies additional styling, 
+tests and configuration in order to be completed.
+
+## Usage
+
+Include plugin via 
+`<PasswordStrength dataSource={DATA_STRING}></PasswordStrength>`
```
MikeTaylor: Shouldn't this be `password={DATA_STRING}`?

lib/PasswordStrength/readme.md
```diff
@@ -0,0 +1,27 @@
...
+import { formValues } from 'redux-form';
+
+getCurrentValues(key) {
+        const {stripes: {store}} = this.props;
+        const state = store.getState();
+        const values = getFormValues('Form')(state) || {};
+        return values[key] || false;
+}
+
+<PasswordStrength type="password-strength" dataSource={this.getCurrentValues("InputName")}>
```
MikeTaylor: Likewise, `password`?

lib/PasswordStrength/PasswordStrength.js
```diff
@@ -0,0 +1,53 @@
...
+    super(props);
+    this.strengthTester = new taiPasswordStrength.PasswordStrength();
+    this.strengthTester.addCommonPasswords(taiPasswordStrength.commonPasswords);
+    this.strengthTester.addTrigraphMap(taiPasswordStrength.trigraphs);
+    this.output = 'weak'; // set default password hint to weak
+  }
+
+  componentDidUpdate(prevProps) {
+    // in case password updated and provided
+    if (this.props.password !== prevProps.password && !!this.props.password) {
```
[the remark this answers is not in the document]
oleksii-popov: Thank you, updated

lib/PasswordStrength/PasswordStrength.js
oleksii-popov: As it is POC - I presented the main idea of working with strengthTester results.
In a final version, there should be more check (very weak, weak, reasonable, strong, very strong ).
I suggest leaving <code>switch</code> construction as it will make the code more readable in future versions.

lib/PasswordStrength/PasswordStrength.js
oleksii-popov: React render want to get something here. Refactored into <code>return null</code> and simplified.

lib/PasswordStrength/readme.md
oleksii-popov: Yes, I forgot to update attribute name in the .md. Nice catch.

lib/PasswordStrength/readme.md
oleksii-popov: It can be 'confirmed password' or 'change password' or 'new password'. Just want to avoid blind copy-paste. Developers should pay attention to what exactly they are working with.

lib/PasswordStrength/PasswordStrength.js
```diff
@@ -0,0 +1,52 @@
...
+
+  render() {
+    if (!this.props.password) {
+      return null;
+    }
+
+    return (
+      <div aria-live="polite">
+        <FormattedMessage id={'stripes-smart-components.password-strength.' + this.output} />
+      </div>
```
JohnC-80: It's best that the aria-live region is present already before a message is placed there... The early presence of the aria-live region is what will allow the assistive technology to monitor it for changes so that those can be announced. If the whole thing appears at once, it may not be picked up. You could have it render the empty `aria-live` div by default and then conditionally render the formatted message.

oleksii-popov: Then I can remove <code>aria-live="polite"</code> here and and add it into readme.md example. Am I correct?
Also, can you please suggest how do you test accessibility? I used "Screen Reader" and it works properly.

oleksii-popov: Updated

lib/PasswordStrength/readme.md
zburke: Looks like this still needs to be fixed.

translations/stripes-smart-components/en.json
```diff
@@ -82,5 +82,9 @@
   "deleteTag": "Remove tag",
   "enterATag": "Enter a tag.",
   "tagsTextArea": "Tag text area",
-  "addTag": "Add"
+  "addTag": "Add",
+
+  "password-strength.weak": "Your password is too weak.",
```
zburke: How about `passwordStrength.weak` etc to be consistent with the capitalization scheme elsewhere in this file?

oleksii-popov: Sure. Updated

lib/PasswordStrength/PasswordStrength.js
oleksii-popov: @MikeTaylor refactored into if/else.

oleksii-popov:
Looks like this piece of code approved. I would like to ask anyone, who has write access to this repository to merge PR.

MikeTaylor:
Done. Thanks for your work on this.

PR #233, opened 2018-08-28
Fix initial query string to keep searchField updated. Fixes STSMACOM-119

mkuklis:
https://issues.folio.org/browse/STSMACOM-119

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -207,13 +207,7 @@ class SearchAndSort extends React.Component {
     this.transitionToParams = this.transitionToParams.bind(this);
   }
 
-  componentDidMount() {
-    const query = this.queryParam('query');
-    // eslint-disable-next-line react/no-did-mount-set-state
-    this.setState({ locallyChangedSearchTerm: query });
-  }
-
-  UNSAFE_componentWillReceiveProps(nextProps) { // eslint-disable-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps) {  // eslint-disable-line react/no-deprecated
     const logger = this.props.stripes.logger;
     const oldState = makeConnectedSource(this.props, logger);
     const newState = makeConnectedSource(nextProps, logger);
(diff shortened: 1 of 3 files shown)


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -206,12 +206,6 @@ class SearchAndSort extends React.Component {
     this.transitionToParams = this.transitionToParams.bind(this);
   }
 
-  componentDidMount() {
```
MikeTaylor: Why where you able to straight-up remove this? Was it not doing anything for us?

MikeTaylor:
Eh? Why did you close this?

lib/SearchAndSort/SearchAndSort.js
mkuklis: Yes exactly. It was not doing anything for us. The component would get mount before we had any resources available.

mkuklis:
Oh wow I did not close this. That's crazy.

lib/SearchAndSort/SearchAndSort.js
MikeTaylor: Looks like you added it in https://github.com/folio-org/stripes-smart-components/commit/3a3c056a23019eba2a03d7927ea1d1ef07f89da8 :-)

mkuklis:
What a heck is going on here! I need a break.

lib/SearchAndSort/SearchAndSort.js
mkuklis: @MikeTaylor something weird is happening with github because I don't see it being added anymore
https://github.com/folio-org/stripes-smart-components/pull/233/files

mkuklis: @MikeTaylor ah never mind. You pointed to a commit from April which I thought was my commit from today and got confused. Yes that commit was wrong.

mkuklis: @MikeTaylor actually thank you for pointing that out. I will need to take a closer look now.

mkuklis:
@MikeTaylor thanks for pointing out this other commit. I had to make couple more changes to make sure https://issues.folio.org/browse/STSMACOM-81 still works fine.

PR #242, opened 2018-09-04
STCOM-328 Move EditableList

cherewaty:
## Purpose
`EditableList` doesn't meet the definition of a purely presentational component (https://issues.folio.org/browse/STCOM-298), since it's dependent on a redux store. It's used by more than one `folio-org` repo, so its most logical new home is in `stripes-smart-components`.

https://issues.folio.org/browse/STCOM-328

## Approach
Used `git-subtree` to preserve as much git history for these files as possible from `stripes-components`. `EditableList` was recently moved out of `lib/structures` - unfortunately `git-subtree` didn't preserve any history before that move.

## Next Steps
- [ ] Introduce deprecation warning in `stripes-components`
- [ ] Adjust import in `ui-users`
- [ ] Remove `EditableList` from `stripes-components` `v4.0.0` branch

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -4,7 +4,6 @@ import { FormattedMessage } from 'react-intl';
 
 import Paneset from '@folio/stripes-components/lib/Paneset';
 import Pane from '@folio/stripes-components/lib/Pane';
-import EditableList from '@folio/stripes-components/lib/EditableList/EditableList';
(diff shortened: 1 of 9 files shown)


MikeTaylor:
After this one, can we please hold off on similar "housekeeping" changes until after the rapidly approaching 2018q3 release? We need to make a ton of releases for that, and it's better to do that in a less fast-moving environment. Thanks.

cherewaty:
@MikeTaylor there's a couple more in this batch, but they should all be done this week.

PR #257, opened 2018-09-14
use current stripes-form

zburke:
Because stripes-form has not been released as a v1.0.x package,
this means ^0.8.0 is effectively ~0.8.0, meaning those packages
are locked at 0.8.x, not 0.x.x as intended. stripes-form 0.8.0
pulled in an old version of stripes-components which pulled in
an old version of React which brought darkness upon the land.

Refs [STRIPES-555](https://issues.folio.org/browse/STRIPES-555)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-smart-components
 
+## 1.8.0 (IN PROGRESS)
+
+* Update to stripes-form 0.9.0. Refs STRIPES-555.
+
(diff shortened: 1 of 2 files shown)


MikeTaylor:
This is fine as it goes, but surely it's also big red flag that we need to release stripes-form v1.0 ASAP?

zburke:
It is. Upgrading all our deps on stripes-form to `v0.9.0` seemed like the least invasive approach for managing this on a release branch, but I agree we should publish stripes-form `v1.0.0` pronto.

PR #263, opened 2018-09-18
Adjust lodash imports

cherewaty:
## Purpose
Lodash is notorious for delivering way more code to the bundle than might actually be needed.

## Approach
We should avoid importing all of lodash. Both of these are better options:
```
import get from 'lodash/get';
```
or 
```
import { get } from 'lodash';
```

I haven't confirmed that the second option above appropriately tree-shakes. The first is optimal with our stack.

## Learning
- https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark
- https://stackoverflow.com/questions/35250500/correct-way-to-import-lodash

### DIFF
--- a/lib/AddressFieldGroup/AddressEdit/AddressEdit.js
+++ b/lib/AddressFieldGroup/AddressEdit/AddressEdit.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import _ from 'lodash';
+import has from 'lodash/has';
 
 import { Field } from 'redux-form';
 import stripesForm from '@folio/stripes-form';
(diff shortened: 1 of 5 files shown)


MikeTaylor:
I'm not ecstatic about making the code worse for programmers to work with, for the convenience of the compiler. I see the argument for switching to `import { get, drop, flatten } from 'lodash'`, but a separate line for each imported function? Not so nice. And having each of those lines know precisely _where_ in the lodash library the specific function is defined? That goes in the opposite direction to the approach we're taking in stripes-core.

cherewaty:
https://www.npmjs.com/package/lodash-es works the way we want `stripes-*` repos to work.

We're definitely not at a kilobyte-shaving stage, so I don't plan to spend any time soon looking to see if the tree shaking works for `import { get, drop, flatten } from 'lodash'` (I also didn't change any existing imports using that syntax).

This PR's intent is simply to remove instances of `import _ from 'lodash';` - that's not a habit we want to get in.

MikeTaylor:
Agreed -- although I would characterise `import _ from 'lodash'` not so much as a habit we want to avoid getting into, but as one we need to get out of.

PR #271, opened 2018-09-19
Add stripes-core to devDependencies

cherewaty:
I noticed `eslint` errors popping up on some PRs. `stripes-core` will need to also be a `devDependency` for `eslint` to resolve the imports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
   },
   "devDependencies": {
     "@folio/eslint-config-stripes": "^3.2.0",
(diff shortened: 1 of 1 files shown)


cherewaty:
Only in cases where the module tries to import something from `stripes-core` (which I think applies to all the repos that are needing `stripes-core` moved to `peerDependencies`.)

cherewaty:
@zburke `withStripes` gets exported: https://github.com/folio-org/stripes-core/blob/master/index.js

PR #272, opened 2018-09-19
Simplify stripes-core imports

cherewaty:
Import from the root instead of digging into the directory structure. Follow-up to https://github.com/folio-org/stripes-smart-components/pull/271


mkuklis:
@cherewaty this one looks good too. I was about to release v1.8.3 but I will wait for this one.

PR #279, opened 2018-09-24
Turn on more ESLint rules

cherewaty:
Turns on `max-len` and `react/sort-prop-types`, which have both been very useful in `@folio/stripes-components`.

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -224,7 +224,8 @@ class ControlledVocab extends React.Component {
 
         // Check if the primary field has had data entered into it.
         if (!item[primaryField]) {
-          itemErrors[primaryField] = this.props.stripes.intl.formatMessage({ id: 'stripes-core.label.missingRequiredField' });
+          itemErrors[primaryField] =
+            this.props.stripes.intl.formatMessage({ id: 'stripes-core.label.missingRequiredField' });
         }
 
         // Add the errors if we found any for this record.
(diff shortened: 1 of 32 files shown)


JohnC-80:
Some of these components consist of sub-components and the prop-types just parrot the props of those sub-components... I could see how it would be better to have these separated in custom grouping than one big, flat alpha list - *BUT* I'm not sure there's even an established standard here for that.
That said, would really appreciate the eyes of others who spend more time here before introducing the `sort-prop-types`. @zburke, @doytch 
Additionally, there's gotta be a better way to compound things than explicitly parenting the necessary props since the checking will happen at the lowest level anyhow.

zburke:
If we're saying that `<Foo>` has subcomponents `<Bar>` and `<Bat>` and we're checking all the props at every level, we could refactor `<Foo>` to receive objects in `props.bar` and `props.bat` and spread them when it instantiates the subcomponents. 

I don't know if that would make things more or less clear, but it would allow you group the props by subcomponent and be less strict about type checking at the top level.

lib/ControlledVocab/ControlledVocab.js
```diff
@@ -32,58 +32,58 @@ class ControlledVocab extends React.Component {
...
-    preUpdateHook: PropTypes.func,
+    stripes: PropTypes.shape({
+      connect: PropTypes.func.isRequired,
+      formatDate: PropTypes.func.isRequired,
+      hasPerm: PropTypes.func.isRequired,
+      intl: PropTypes.shape({
+        formatMessage: PropTypes.func.isRequired,
+      }).isRequired,
+    }).isRequired,
+    validate: PropTypes.func,
```
zburke: Unfrozen caveman programmer say, it's pedantic to comment about comment placement given the sad state of comments in our codebase, but I expect block comments to come before the code they describe as it was originally:
```
/* stuff about validate */
validate: PropTypes.func, 
```
This arrangement makes the block appear to refer to `visibleFields`. I would personally prefer a blank line above the comment and below the line it refers to in order to remove all doubt, but ESLint may disagree. Personally, I find our rules on whitespace suffocating, but I like four character tabs and closing braces on empty lines and `-Wall` and `-pedantic` because I'm a caveman; it's just the way I think.

cherewaty:
Rebased last commit to pull in new props on `SearchAndSort`. Prop comments should now all live with their original siblings.

PR #280, opened 2018-09-24
Simplify stripes-components imports

cherewaty:
## Approach
Prefers
`import { Badge, Button, Icon } from '@folio/stripes-components';`

over
```
import Badge from '@folio/stripes-components/lib/Badge';
import Icon from '@folio/stripes-components/lib/Icon';
import Button from '@folio/stripes-components/lib/Button';
```

## Notes
This exposed a few things with `stripes-components`:
- `injectIntl` is specifically not exported, as it's only intended for use with `stripes-components`. Whether the components here need that version (that can forward refs appropriately) or just the `react-intl` upstream version can be addressed in another PR.
- `stripes-components/lib/variables` isn't exported yet.
- `filters2cql` isn't exported - maybe that needs to live here instead? Can be addressed later.
- `NavListItem` isn't exported - that one seems easy to fix.


lib/AddressFieldGroup/readme.md
```diff
@@ -3,7 +3,7 @@
 
 import
 ```js
-import AddressList from '@folio/stripes-components/lib/AddressFieldGroup/AddressList';
+import AddressList from '@folio/stripes-smart-components';
```
VictorSoroka1: Should not this be?

```
import { AddressList } from '@folio/stripes-smart-components';
```

cherewaty: Thanks for the catch. Fixed.

PR #287, opened 2018-10-02
[STSMACOM-122] Setup BigTest Infrastructure

wwilsman:
## Purpose

[STSMACOM-122](https://issues.folio.org/browse/STSMACOM-122)

## Approach

Initially, I started setting up BigTest for this repo similar to the `stripes-components` setup, which uses a `Harness` to apply necessary contexts. The difference would be all of the components here should have available all of the contexts found in `stripes-core`.

Instead of having to maintain and update these contexts when `stripes-core` changes are made, we can utilize the `setupApplication` testing helper defined there, which sets up the entire `stripes-core` app for testing.

During the setup hook for `stripes-smart-components`, a dummy module is created and navigated to so we can mount the smart components under test. Using the `mount` helper for these tests replaces the dummy app so that it renders the component.

I chose to write some examples for the `PasswordValidationField`, because it seems like a simpler smart component that also requires `stripes.connect` to fetch validation rules. While mocking the route for rules, I noticed that there were issues where `stripes-connect` required `response.url` to be defined, but due to pretender mocking fetch with xhr, there is no `response.url`. The fetch polyfill used sets `request.url` based on the presence of `X-Request-URL`. I couldn't find a way to hook into _every_ response to set that header appropriately, so it is included directly in the `new Response` headers returned by the route handler. *This will have to be done any time `stripes-connect` requires a response*, or we might have to work around it in `stripes-connect` itself.

## TODO

- [ ] Update the Jenkins file to use the necessary karma flags


wwilsman:
Replaced with #297

PR #294, opened 2018-10-03
STRIPES-558 Remove notes components

cherewaty:
https://issues.folio.org/browse/STRIPES-558


cherewaty:
Will revisit when I'm back from vacation after Oct. 29. Targeted for Nov. 13 `@folio/stripes` release.

PR #295, opened 2018-10-03
Narrow use of UserLink and UserName

cherewaty:
`<UserLink>` and `<UserName>` could be very dangerous tools - they hook up to the store with `stripes-connect` for very small amounts of data that can generate a lot of server calls. 

Instead that data fetching should be handled by route-level container components in each UI module. `<ControlledVocab>` is the only consumer of `<UserLink>` and `<UserName>` in the `folio-org`, so this PR couples those and removes `<UserLink>` and `<UserName>` from exports.


zburke:
You're right about this being a dangerous performance goblin. 

> that data fetching should be handled by route-level container components in each UI module

Can you elaborate on this? I don't follow.

cherewaty:
Will revisit when I'm back from vacation after Oct. 29.

zburke:
`<ControlledVocab>` has already been refactored to avoid `<UserLink>`. Close this in favor of #375? We can still remove it altogether in v2.0.0.

cherewaty:
Closed in favor of #375. Thanks @zburke.

PR #296, opened 2018-10-03
Narrow use of LoanList

cherewaty:
`<LoanList>` hooks up to the store with `stripes-connect`. As often as possible, that data fetching should be handled by route-level container components in each UI module. `<ChangeDueDateDialog>` is the only consumer of `<LoanList>` in the `folio-org`, so this PR couples those and removes `<LoanList>` from exports.


cherewaty:
Will revisit when I'm back from vacation after Oct. 29.

PR #307, opened 2018-10-08
STCOM-354 Improve SearchAndSort behavior

cherewaty:
Adds full-width "Search" button, requires submit action to start searching, and removes dismiss icon from Search & Filter pane.

Fixes https://issues.folio.org/browse/STCOM-354 and https://issues.folio.org/browse/STSMACOM-132


cherewaty:
There's a very high chance this change will break some integration tests.

zburke:
I can't believe I'm just now finding out about https://designsystem.digital.gov/. 

You're right that some tests will break. Assign me a Jira ticket with before/after logs of the test runs. I'll fix 'em.

cherewaty:
Thanks @zburke 😁 will do.

PR #312, opened 2018-10-12
STCOR-267 Use react-intl directly instead of stripes.intl

cherewaty:
## Purpose
Retrieving the intl object from the `stripes` god object (`stripes.intl`) is a deprecated pattern.

## Approach
Sets some repo-wide conventions:
- Don't use `stripes.intl`
- When using `injectIntl()` that injects an `intl` prop, destructure `intl`
- Don't try to alias `formatMessage()` and its siblings; instead use destructuring to achieve brevity
- Always specify what parts of `props.stripes` are being used in the prop type validation


zburke:
Wow. This was a lot to work through. Thanks for the monster clean-up job!

PR #323, opened 2018-10-31
STCOM-391 Use plus-sign icon instead of + character

cherewaty:
Stops screen readers from announcing "plus": https://issues.folio.org/browse/STCOM-391

### Before
screen shot 2018-10-31 at 11 39 48 am

### After
screen shot 2018-10-31 at 11 39 01 am

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -7,6 +7,7 @@ import PropTypes from 'prop-types';
 import Route from 'react-router-dom/Route';
 import { withRouter } from 'react-router';
 import { FormattedMessage, injectIntl, intlShape } from 'react-intl';
+import { FormattedXmlMessage } from 'react-intl-formatted-xml-message';
 import { withModule } from '@folio/stripes-core/src/components/Modules';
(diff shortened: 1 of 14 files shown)


cherewaty:
Thanks for the additional context @julianladisch. I added a commit that I think will get us the flexibility we need. We'll definitely have more cases of icons needing to be localizable.

With [`react-intl-formatted-xml-message`](https://github.com/ephys/react-intl-formatted-xml-message/), the icon will appear in the translations file.

The usage in JSX:
```
<FormattedXmlMessage
  id="stripes-smart-components.new"
  tags={{ 'icon': (<Icon />) }}
/>
```

In a language that wants a `plus-sign`, the translation key:
`"new": "<icon icon='plus-sign'>New</icon>",`

In a language that wants some other icon we've defined in https://github.com/folio-org/stripes-components/blob/master/lib/Icon/icons.js:
`"new": "<icon icon='some-other-icon-key'>New</icon>",`

In a language where we don't want an icon at all:
`"new": "New",`

I'll put a test on the `<Icon>` component in `stripes-components` that it's handling RTL languages appropriately when it receives a child string.

julianladisch:
Great, thanks!

PR #330, opened 2018-11-05
STSMACOM-145: Update text and position of "+ Add address" button on User record

MaksymDryha:
According to [this](https://issues.folio.org/browse/STSMACOM-145) story, "Add address" button should be aligned to the left so that the label text and position for this type of button follow the latest general FOLIO style guidelines:

user-details-4

### DIFF
--- a/lib/AddressFieldGroup/AddressEdit/AddressEdit.css
+++ b/lib/AddressFieldGroup/AddressEdit/AddressEdit.css
@@ -53,7 +53,6 @@
 .addressFormList {
   list-style: none;
   padding: 0;
-  margin-bottom: 0;
 }
 
 .addressFormListItem {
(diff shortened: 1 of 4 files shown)


cherewaty:
@MaksymDryha we just figured out a new strategy for how to localize icons in buttons in https://github.com/folio-org/stripes-smart-components/pull/323

You should be able to replace:
```
<FormattedMessage id="stripes-smart-components.address.addAddress" />
```
with 
```
<FormattedXmlMessage id="stripes-smart-components.address.addAddress" tags={{ 'icon': (<Icon />) }} />

```

and then update the translation string to be:
```
"address.addAddress": "<icon icon='plus-sign'>Add address</icon>"
```

(you'll have to add imports for `Icon` and `FormattedXmlMessage`)

MaksymDryha:
@cherewaty Thanks! I updated the PR according to the information that you provided.

PR #331, opened 2018-11-06
Replace all formatMessage() with <FormattedMessage />

cherewaty:
## Purpose
Finishes the work from https://github.com/folio-org/stripes-smart-components/pull/329. Applied to `stripes-components` in https://github.com/folio-org/stripes-components/pull/696.

Aligns with `stripes` i18n documentation at https://github.com/folio-org/stripes/blob/master/doc/i18n.md, in preferring the `react-intl` component API instead of its functions with `injectIntl()`.

## Approach
- Exclusively use `<FormattedMessage>`, `<FormattedDate>`, and `<FormattedTime>` from `react-intl`. 
- Discontinue using `intl.formatMessage()`, `intl.formatDate()`, and `intl.formatTime()`.
- I had to leave two `injectIntl()`s - one is in the deprecated `NoteRenderer`, and the other is in `DueDatePicker`. At some point, `DueDatePicker`'s date handling should be re-evaluated to see if can just use `<FormattedDate>`. Not a priority, and not a big deal if it continues to use `injectIntl()`.

## Risk
`stripes-smart-components` doesn't have the kind of test coverage it needs, so this has a high chance of minor regressions. The most likely is `[Object object]` showing up in the UI instead of the internationalized string. Issues that arise should be quick to remedy.

### DIFF
--- a/lib/ChangeDueDateDialog/DueDatePicker.js
+++ b/lib/ChangeDueDateDialog/DueDatePicker.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import moment from 'moment-timezone';
-import { injectIntl } from 'react-intl';
+import { injectIntl, intlShape } from 'react-intl';
 
 import DueDatePickerForm from './DueDatePickerForm';
 
@@ -12,6 +12,7 @@ class DueDatePicker extends React.Component {
       date: PropTypes.string,
       time: PropTypes.string,
     }),
+    intl: intlShape,
     onChange: PropTypes.func,
     timeProps: PropTypes.object
   }
--- a/lib/ChangeDueDateDialog/DueDatePicker.js
+++ b/lib/ChangeDueDateDialog/DueDatePicker.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import moment from 'moment-timezone';
-import { injectIntl, intlShape } from 'react-intl';
+import { injectIntl } from 'react-intl';
 
 import DueDatePickerForm from './DueDatePickerForm';
 
@@ -12,7 +12,6 @@ class DueDatePicker extends React.Component {
       date: PropTypes.string,
       time: PropTypes.string,
     }),
-    intl: intlShape,
     onChange: PropTypes.func,
     timeProps: PropTypes.object
   }
--- a/lib/EditableList/EditableListForm.js
+++ b/lib/EditableList/EditableListForm.js
@@ -2,6 +2,7 @@ import isEqual from 'lodash/isEqual';
 import cloneDeep from 'lodash/cloneDeep';
 import uniqueId from 'lodash/uniqueId';
 import React from 'react';
+import { FormattedMessage } from 'react-intl';
 import stripesForm from '@folio/stripes-form';
 import { FieldArray } from 'redux-form';
 import PropTypes from 'prop-types';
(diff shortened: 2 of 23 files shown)


lib/EditableList/EditableListForm.js
```diff
@@ -377,7 +373,6 @@ class EditableListForm extends React.Component {
             size="small"
             id={`clickable-edit-${this.testingId}-${item.rowIndex}`}
             onClick={() => this.onEdit(item.rowIndex)}
-            title={formatMessage({ id: 'stripes-components.editThisItem' })}
```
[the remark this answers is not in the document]
cherewaty: We're moving away from `title`, yes: https://issues.folio.org/browse/STSMACOM-124. In this particular case the button was plenty explanatory, so no `aria-label` is really needed. The first rule of ARIA is don't use ARIA 😎

lib/ChangeDueDateDialog/DueDatePicker.js
```diff
@@ -12,7 +12,6 @@ class DueDatePicker extends React.Component {
       date: PropTypes.string,
       time: PropTypes.string,
     }),
-    intl: intlShape,
```
wwilsman: Should we keep this prop-type since the component still uses `intl.timeZone`?

lib/LocationLookup/LocationForm.js
```diff
@@ -72,48 +66,62 @@ class LocationForm extends React.Component {
...
-              component={Selection}
-              label={this.translate('location')}
-              placeholder={this.translate('selectLocation')}
-              name="locationId"
-              id="locationId"
-              onFilter={this.filter}
-              dataOptions={[{ label: '\u00A0', value: '' }, ...locationOpts]}
-            />
+            <FormattedMessage id="stripes-smart-components.ll.selectLocation" />
+            {placeholder => (
```
wwilsman: This doesn't look correct. Shouldn't this be a child of `FormattedMessage`?

lib/Notes/NoteRenderer.js
```diff
@@ -125,34 +125,42 @@ class NoteRenderer extends React.Component {
...
+                      bsRole="toggle"
                     >
-                      <FormattedMessage id="stripes-components.button.delete" />
+                      <Icon icon="down-caret" color="rgba(150, 150, 150, .5)" />
                     </Button>
-                  </IfPermission>
-                </DropdownMenu>
+                  )}
+                </FormattedMessage>
+                <FormattedMessage id="">
```
wwilsman: I think since both of these messages are `editOrDeleteNote`, this `FormattedMessage` without an `id` probably isn't meant to exist.

lib/Notes/NotesForm.js
```diff
@@ -48,33 +46,44 @@ class NotesForm extends React.Component {
...
-          aria-label={formatMessage({ id: 'stripes-smart-components.notesTextArea' })}
-          fullWidth
-          id="note_textarea"
-          component={TextArea}
-          onKeyDown={(e) => { handleKeyDown(e, handleClickSubmit); }}
-          rows={textRows}
-        />
+        <FormattedMessage id="stripes-smart-components.enterANote">
+          {placeholder => (
+            <FormattedMessage id="">
```
wwilsman: Missing the `notesTextArea` message id here

lib/EditableList/EditableListForm.js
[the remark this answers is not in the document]
cherewaty: Haha wow no that's me being dense. That still needs an `aria-label`. I'll fix that.

cherewaty:
Commit added that resolves https://issues.folio.org/browse/STSMACOM-124

PR #333, opened 2018-11-08
Update renderable props to type node

cherewaty:
Props that are simply rendered like `label` or `paneTitle` should accept any renderable node, not just a string, so we can pass in `react-intl` `<FormattedMessage>`, etc.

Prevents warnings like:
```
Warning: Failed prop type: Invalid prop `sections[0].label` of type `object` supplied to `Settings`, expected `string`.
```

### DIFF
--- a/lib/EntryManager/EntryForm.js
+++ b/lib/EntryManager/EntryForm.js
@@ -9,7 +9,7 @@ class EntryForm extends React.Component {
   static propTypes = {
     deleteDisabled: PropTypes.func.isRequired,
     deleteDisabledMessage: PropTypes.string.isRequired,
-    entryLabel: PropTypes.string.isRequired,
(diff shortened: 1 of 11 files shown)


lib/EntryManager/EntryForm.js
```diff
@@ -148,7 +148,7 @@ class EntryForm extends React.Component {
 
             {selectedEntry &&
               <ConfirmationModal
-                id={`delete${this.props.entryLabel.replace(/[^a-zA-Z0-9]/g, '').toLowerCase()}-confirmation`}
+                id={`delete${this.props.nameKey.replace(/[^a-zA-Z0-9]/g, '').toLowerCase()}-confirmation`}
```
cherewaty: Heads up @zburke this id was reliant on a translated string, so I changed it to a stable value. Likely will cause an integration test failure.

PR #338, opened 2018-11-10
UIU-679 Update label propTypes

cherewaty:
## Purpose
While working on https://issues.folio.org/browse/UIU-679, @Godlevskyi noticed that passing in `<FormattedMessage>` nodes to some `<ControlledVocab>` props would cause incidents of `[Object object]` making their way to the outputted UI.

For `labelSingular` in particular, `<ControlledVocab>` would attempt to transform the given already-localized string to lower case. This probably will not scale to languages with different capitalization rules than English (like German). We shouldn't be doing any transformations of the strings provided by translators.

## Approach
- Accept `node` prop type for `label`, `labelSingular`, and `objectLabel`
- Stop attempting to change capitalization of already-localized strings
- Change confirmation modal id to use `nameKey` prop instead of a transformed `labelSingular`

## Open questions
- Was there a need for the `toLowerCase()` that I'm missing?

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -378,7 +378,7 @@ class ControlledVocab extends React.Component {
             />
           }
           <ConfirmationModal
-            id={`delete${this.props.nameKey}-confirmation`}
+            id="delete-controlled-vocab-entry-confirmation"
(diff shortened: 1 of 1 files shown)


cherewaty:
Thanks @zburke. ID updated to `delete-controlled-vocab-entry-confirmation`.

PR #347, opened 2018-11-20
Add stopPropagation to search form. Refs UIU-731

mkuklis:
It looks like this was yet another weird issue with redux-form. The `onSubmit` handler in SearchAndSort form was also executing the UserForm `onSubmit` handler here:

https://github.com/folio-org/ui-users/blob/master/src/UserForm.js#L197

This PR just replaces the `onSubmit` with `onClick` since we don't really need to submit the SearchAndSort form. This seems to fix the issue in UserForm but it's still not clear why the UserForm `onSubmit` would execute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-smart-components
 
+## 1.12.0 (IN PROGRESS)
+
+* Remove onSubmit from search form. Refs UIU-731. 
+
 ## [1.11.0](https://github.com/folio-org/stripes-smart-components/tree/v1.11.0) (2018-11-19)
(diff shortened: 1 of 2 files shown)


cherewaty:
Wouldn't this mean that pressing enter wouldn't submit the search form? Can we instead change some organization so the `UserForm` isn't nested inside of the search form?

mkuklis:
@cherewaty I don't think `UserForm` is currently nested inside search form. Let me double check that.

mkuklis:
@cherewaty I added `stopPropagation` instead and brought back `onSubmit`. This seems to be working fine. 

BTW this issue occurs via `ui-plugin-find-user`

PR #350, opened 2018-11-23
Replace actionMenuItems-prop with actionMenu-prop

rasmuswoelk:
**Issue:** https://issues.folio.org/browse/STSMACOM-147


cherewaty:
Since this is a breaking change, we should create a smoother migration path. Perhaps add `actionMenu` but simply deprecate `actionMenuItems` for now (with `deprecated` from `prop-types-extra`)?

PR #352, opened 2018-11-27
STSMACOM-133: Apply arrows for magnifying glass

anastasiaHahanova:
## Purpose
Regarding [STSMACOM-133](https://issues.folio.org/browse/STSMACOM-133), arrow icons should be applied to magnifying glass to indicate the state of the searching pane. The existence of badge should be taken into the account (when filters are applied). 

_Animation for pane will be applied later._

## Approach
- Add arrow icons to stripes-components [STCOM-407](https://github.com/folio-org/stripes-components/pull/731).
- Create `<SearchButton> `component, where the icons are applied.
- Add animation for the icons.

## Screenshots

Without animation

:

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -8,10 +8,14 @@ import { deprecated } from 'prop-types-extra';
 import Route from 'react-router-dom/Route';
 import { withRouter } from 'react-router';
 import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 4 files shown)


filipjakobsen:
I am not able to try this out, as I do not have the code running on my computer; but judging from the GIF attached, I suggest we keep the animation in the code, but comment it out or otherwise disable it, until the pane animates as well.

MikeTaylor:
Hi, @anastasiaHahanova. I'm not working on FOLIO any more, so I'm declining your request to review this PR. Good luck with it!

PR #353, opened 2018-11-28
Replacing actionMenuItems-prop with actionMenu-prop in EntrySelector

rasmuswoelk:


lib/EntryManager/EntrySelector.js
```diff
@@ -92,6 +92,41 @@ class EntrySelector extends React.Component {
...
+        onClick={button.onClick}
+        id={button.id}
+      >
+        <Icon icon={button.icon}>
+          {button.label}
+        </Icon>
+      </Button>
+    ));
+  }
+
```
cherewaty: It looks like "Duplicate" is the only option ever available - if that's the case, we could just return a `<Button>` for that, and don't have to map over `actions`. That would cut `getActionMenu()` in half.

PR #391, opened 2019-01-10
STCOM-439: Add multiselection filter

Yurii-Danylenko:
https://issues.folio.org/browse/STCOM-439

## Purpose
Create a reusable multiselection filter, based on MultiSelection component, to be used in SearchAndSort side panel.

## Approach
- create `onChangeHandler` which will raise filter name and selected values after a filter change.
- add a couple of props
- add tests only specific for this component

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,7 @@ export { default as PasswordValidationField } from './lib/PasswordValidationFiel
 export { default as ProxyManager } from './lib/ProxyManager';
 
 export { default as SearchAndSort } from './lib/SearchAndSort';
+export { default as MultiSelectionFilter } from './lib/SearchAndSort/components/MultiSelectionFilter';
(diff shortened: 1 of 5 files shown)


doytch:
Same as #392, do we need to export this from stripes-smart-components/index.js as well?

index.js
```diff
@@ -26,6 +26,7 @@ export { default as PasswordValidationField } from './lib/PasswordValidationFiel
 export { default as ProxyManager } from './lib/ProxyManager';
 
 export { default as SearchAndSort } from './lib/SearchAndSort';
+export { default } from './lib/SearchAndSort/components/MultiSelectionFilter';
```
doytch: Does this work? Shouldn't it be `export { default as MultiSelectionFilter } from './lib/...`?

PR #392, opened 2019-01-10
STCOM-440: Add checkbox filter

Yurii-Danylenko:
https://issues.folio.org/browse/STCOM-440

## Purpose
Create a reusable checkbox filter, based on Checkbox component, to be used in SearchAndSort side panel. Presents a set of checkboxes working like one filter.

## Approach
- create `onChangeHandler` which will raise filter name and selected values after a filter change.
- add a couple of props
- add tests only specific for this component

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,7 +26,11 @@ export { default as PasswordValidationField } from './lib/PasswordValidationFiel
 export { default as ProxyManager } from './lib/ProxyManager';
 
 export { default as SearchAndSort } from './lib/SearchAndSort';
+
+export { default as CheckboxFilter } from './lib/SearchAndSort/components/CheckboxFilter';
+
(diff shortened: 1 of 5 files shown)


doytch:
Do we need to export this from stripes-smart-components/index.js as well?

index.js
```diff
@@ -26,6 +26,7 @@ export { default as PasswordValidationField } from './lib/PasswordValidationFiel
 export { default as ProxyManager } from './lib/ProxyManager';
 
 export { default as SearchAndSort } from './lib/SearchAndSort';
+export { default } from './lib/SearchAndSort/components/CheckboxFilter';
```
doytch: Same comment as on #391; I was under the impression we need to `export { default as CheckboxFilter } ... `since otherwise we're making CheckboxFilter the default export, which we don't want.

PR #394, opened 2019-01-10
Replace mutator.query.update with history.push. Part of STCOM-302

mkuklis:
This PR fixes an issue with the unsaved changes reported in https://issues.folio.org/browse/STCOM-302

It replaces a `mutator.query.update` with `history.push`. This plays well with react-router and stripes-form and it should fix an issue recorded by Ann-Marie:

https://issues.folio.org/secure/attachment/15610/Get%20Stuck%20trying%20to%20cancel%20Continue%20Editing.mp4

Ideally we should try to move away from managing routing via `mutator.query.update` and replace it with `history` and `<Link>`.

### DIFF
--- a/lib/EntryManager/EntryWrapper.js
+++ b/lib/EntryManager/EntryWrapper.js
@@ -115,27 +115,20 @@ export default class EntryWrapper extends React.Component {
   }
 
   goTo(entry) {
-    const path = `${this.props.match.path}/${entry.id}`;
-    this.props.mutator.query.update({
-      _path: path,
-      layer: '',
-    });
-
+    const { history, match } = this.props;
+    const path = `${match.path}/${entry.id}`;
+    history.push(path);
     this.onSelect(entry);
   }
 
   hideLayer() {
-    this.props.mutator.query.update({
-      _path: this.props.location.pathname,
-      layer: '',
-    });
+    const { location, history } = this.props;
+    history.push(location.pathname);
   }
 
   showLayer(layer) {
-    this.props.mutator.query.update({
-      _path: this.props.location.pathname,
-      layer,
-    });
+    const { location, history } = this.props;
+    history.push(`${location.pathname}?layer=${layer}`);
   }
 
   showCalloutMessage(name) {


zburke:
Hmmmmmmm. Deep down, I imagined `mutator.query.update()` called `history.push()`. Doesn't it? 

Directly using `history` may well be the right way to fix this, and the right way to manage URL state everywhere, but it feels like the kind of change we should plan and document since avoiding deliberately avoiding `history` in favor of `mutator.query...` has been de rigueur for quite some time.

mkuklis:
I agree. Yes STCOM-302 is a bigger issue and this PR only takes care of the cases where `EntryManager` is used (on settings pages).

Unfortunately `mutator.query.update` is not going to work correctly here with the steps described in STCOM-302. 

What happens now is that when the form is marked as dirty and then we hit `X` to close it (to go back)
we modify the query's state via `mutator.query.update`. The `history.push` gets called:

https://github.com/folio-org/stripes-core/blob/master/src/locationService.js#L74

but then the history is blocked by stripes-form: 

https://github.com/folio-org/stripes-form/blob/master/lib/StripesFormWrapper.js#L25

The confirmation dialog appears and if the user chooses to continue and makes another change and then hits 'X' again nothing will happen because the query state has already been changed in the previous attempt to close the form. Eventually we always land here:

https://github.com/folio-org/stripes-core/blob/master/src/locationService.js#L67

It would be really nice to stop patching `locationService` and stop relying on `mutator.query` for navigation but of course this is going to be a big undertaking. I think `SearchAndSort` is another place where `mutator.query.update` is used heavily.

JohnC-80:
Agreed with @zburke that this is definitely an architectural discussion for the pattern to adopt going forward. `mutator.query` vs native browser history api is only a single aspect (but a big one!) of the "routing" struggle the platform faces. I think in the future we may see ourselves supplementing the browser api, but in the better way, allowing the browser location to drive state rather than vice versa. The instant the state/store is out of sync, problems arise and, as @mkuklis describes, LocationService gets patched(ugh). `mutator.query` certainly won't go as lightly in other places as it does here. Admittedly, it's a graceful API for updating a query string... but.... In my gut I feel like I ( and hopefully all of us) would love the app structure of the future when `<Layer>` contents get their own `<Route>` and paths are preferred over query strings for full-screen occupying segments of the user workflow. It's a worthy stripes-arch discussion with accompanying RFC JIRA. I'd be happy to write one up.

mkuklis:
Guys let's discuss this a bit more during our call next week. I spent a bit more time with ui-inventory and SearchAndSort last night and today and reviewed it a bit closer and I think we should be able to get rid of `mutator.query.update` in some areas where we interact with the forms. I think this change could happen gradually without breaking our current navigation setup.

JohnC-80:
RFC Posted - will update the description as we discuss there and in architecture meetings: https://issues.folio.org/browse/FOLIO-1694

PR #396, opened 2019-01-15
CheckboxFilter: Stopped rendering some props to DOM

doytch:
`Checkbox` passes its props down to the `<input type="checkbox">` and React complains when we tell it to render `dataOptions` and `selectedValues` to the DOM.

@Yurii-Danylenko, there may be a similar change needed for `MultiSelectionFilter` but I haven't looked yet and am just using `CheckboxFilter` right now. Feel free to look into it if you want, otherwise I'll find/fix it when I do get around to using those later this week or the week after next (am on vacation next week).

### DIFF
--- a/lib/SearchAndSort/components/CheckboxFilter/CheckboxFilter.js
+++ b/lib/SearchAndSort/components/CheckboxFilter/CheckboxFilter.js
@@ -46,12 +46,15 @@ export default class CheckboxFilter extends React.Component {
 
     return (
       dataOptions.map(({ value, label, disabled, readOnly }) => {
+        const name = typeof label === 'string' ? label : value;
+
         return (
           <Checkbox
             {...rest}
             data-test-checkbox-filter-data-option={value}
             key={value}
             label={label}
+            name={name}
             disabled={disabled}
             readOnly={readOnly}
             checked={selectedValues.includes(value)}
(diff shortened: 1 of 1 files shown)


doytch:
Looking a bit more, I see we're also passing `name` down as a prop. As a result, all the `<input>`s rendered by a `CheckboxFilter` will have the same `name` attribute. I wonder if we should instead be constructing a name in the `.map()` callback based on the `CheckboxFilter`'s name prop and the value of the checkbox we're rendering. Something like `${this.props.name} ${label}`. Or just passing `name={label}`? I'm not sure.

Opinions, @cherewaty and @JohnC-80 ?

Yurii-Danylenko:
Thanks, for the tweak! Sure, I'll make the same changes in the `MultiSelectionFilter` filter. Regarding to the names of the checkboxes, it's better to keep it unique. 'name' for form elements are like identificators. I think, `label` is a suitable prop to be assigned to `name`. Mark, as you have already created the PR, could you assign `label` to `name` prop?

doytch:
Will do, @Yurii-Danylenko!

doytch:
Added that little bit of logic. The one caveat is that `label` can be a node since it could be a react-intl `FormattedMessage`. In those cases, I've fallen back to just using the `value`. It's not a perfect solution but I don't see one here.

PR #404, opened 2019-01-22
Code clean up for SearchAndSort component. It is a basement for the f…

VictorSoroka1:
…urther splitting process.

Part of UIDATIMP-111.


mkuklis:
Ah I think your comment indicates further split. I missed that. Perfect!

PR #415, opened 2019-01-30
UICIRC-108: Change ActionMenu behaviour.

OleksandrAntonenko1:
# Screen shot
screen shot 2019-01-30 at 6 13 22 pm


maximdidenkoepam:
I am ok with this solution, but @mkuklis or @zburke should look at it before merge.

PR #419, opened 2019-01-30
STCOM-365 MCL ARIA roles

JohnC-80:
In conjunction with https://github.com/folio-org/stripes-components/pull/843


lib/EditableList/ItemView.js
```diff
@@ -2,14 +2,15 @@ import React from 'react';
...
-  <div className={css.editListRow} role="row">
+const ItemView = ({ cells, rowIndex }) => (
+  <div className={css.editListRow} role="row" aria-rowindex={rowIndex + 2}>
     {cells}
   </div>
 );
 
 ItemView.propTypes = {
   cells: PropTypes.arrayOf(PropTypes.object).isRequired,
+  rowIndex: PropTypes.number
```
VictorSoroka1: Should this propType be required?

JohnC-80: 👍 I'll make it so. For the most, this comes from MCL, but if it isn't there, the addition would fail...

JohnC-80: Made it so. Makes sense since the "addition" leans on it.

PR #429, opened 2019-02-08
Replace query resource with history push. Refs STCOM-302

mkuklis:
This PR does something very similar to what was done in https://github.com/folio-org/stripes-smart-components/pull/394 but in SearchAndSort.

It should solve the issue reported in https://issues.folio.org/browse/STCOM-302

It also takes us one step closer to not rely on https://github.com/folio-org/stripes-core/blob/master/src/locationService.js#L53 for routing.  

@VictorSoroka1 would you mind giving it a try in your copy of SearchAndSort and see if you can reproduce the STCOM-302 issue?

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -46,6 +46,8 @@ import NoResultsMessage from './components/NoResultsMessage';
 import ResetButton from './components/ResetButton';
 import SearchButton from './components/SearchButton';
 
+import buildUrl from './buildUrl';
+
 import css from './SearchAndSort.css';
 
 class SearchAndSort extends React.Component {
@@ -383,13 +385,14 @@ class SearchAndSort extends React.Component {
 
   transitionToParams = values => {
     const {
+      location,
+      history,
       nsParams,
-      parentMutator: { query },
     } = this.props;
 
     const nsValues = mapNsKeys(values, nsParams);
-
-    query.update(nsValues);
+    const url = buildUrl(location, nsValues);
+    history.push(url);
   };
 
   onNeedMore = () => {
@@ -415,6 +418,8 @@ class SearchAndSort extends React.Component {
         return;
       }
     }
+
+    console.log('on select row...');
 
     this.log('action', `clicked ${meta.id}, selected record =`, meta);
     this.setState({ selectedItem: meta });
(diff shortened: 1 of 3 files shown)


lib/SearchAndSort/buildUrl.js
```diff
@@ -0,0 +1,32 @@
+import queryString from 'query-string';
```
VictorSoroka1: @mkuklis, could you please export this file in `index.js` so it can be reused.

VictorSoroka1:
@mkuklis, will try on Monday, thank you.

VictorSoroka1:
@mkuklis, have tested code, works fine. The issue is gone. Thank you!

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -383,13 +385,14 @@ class SearchAndSort extends React.Component {
...
+      history,
       nsParams,
-      parentMutator: { query },
     } = this.props;
 
     const nsValues = mapNsKeys(values, nsParams);
-
-    query.update(nsValues);
+    const url = buildUrl(location, nsValues);
+    history.push(url);
```
VictorSoroka1: Could you please have an empty line after the definition of variables. I have such setup in my project and followed this pattern while cleaning up the S&S component. I also ensure the blank line before the `return` construction. This is only for readability purposes.

mkuklis: @VictorSoroka1 thank you for reviewing it. Yes of course I will change it. Also are you using Visual Studio Code? If so I would love to see where you can configure it.

VictorSoroka1: I use the following `eslint` rule:

```
    "padding-line-between-statements": [
      "error",
      { "blankLine": "always", "prev": ["const", "let"], "next": "*"},
      { "blankLine": "any", "prev": ["const", "let"], "next": ["const", "let"]},
      { "blankLine": "always", "prev": "*", "next": ["return"]}
    ],
```

I understand that this is not a part of that project so adding this may end up in a bunch of lint errors.

mkuklis: Thanks. Do you maintain a global .eslintrc?

VictorSoroka1: I am extending this project but add several rules on top of it in my project.

mkuklis: Thank you. I was able to set them up. Now I'm getting over 100 errors :) but I will fix this one.

VictorSoroka1: Thanks 😄.

PR #435, opened 2019-02-13
Fix search when SearchAndSort is used via plugin. Refs STCOM-302

mkuklis:
Unfortunately the previous PR #429 doesn't work well when SearchAndSort is used via plugin (for example when we perform users lookup from the checkout screen). This PR goes back to what we had previously (rely on query resource) for those case.

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -395,12 +395,21 @@ class SearchAndSort extends React.Component {
       location,
       history,
       nsParams,
+      browseOnly,
+      parentMutator: { query },
     } = this.props;
 
     const nsValues = mapNsKeys(values, nsParams);
     const url = buildUrl(location, nsValues);
 
-    history.push(url);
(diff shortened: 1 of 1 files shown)


zburke:
@JohnC-80 put me onto https://itnext.io/auxiliary-routing-with-react-e0a4eee36122 a while back. I wonder if we could do something similar for plugins, handing them some specific section of the query string and saying "This is yours, you can do whatever you want with it, but it's the only part you can have." The same strategy would maybe simplify some of the routing shenanigans in ui-users with sorting fees and fines etc.

mkuklis:
Yes that's a nice idea. I worry because we have some crazy cases with that plugin. For example we are loading users lookup inside ui-users. This is definitely a corner case but it's something we will have to think about.

I was also thinking about perhaps completely rewriting that users lookup and not relying on ui-users for it (it's probably more work).

zburke:
The way plugins use apps, loading them by path rather than by export, is completely sketchy. It might be more work to rewrite them but in the long run I think we will have to.

PR #453, opened 2019-03-06
UIDATIMP-156: change ViewMetaData component

OleksandrYehorovEPAM:
## Purpose
Make sure `ViewMetaData` component source fields for updated and created lines renders correctly for the record created by System

## Approach
Check the passed id from metadata and if it contains only zeros then assume that this is System. In that case component will not make unnecessary http request and will render system user wich has a default value "System" or can be passed as a property.

## Screenshots
Before:
image

After:
image

### DIFF
--- a/lib/ViewMetaData/ViewMetaData.js
+++ b/lib/ViewMetaData/ViewMetaData.js
@@ -1,15 +1,38 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { MetaSection } from '@folio/stripes-components';
-import {
-  withStripes,
-  stripesShape,
-} from '@folio/stripes-core';
 
 class ViewMetaData extends React.Component {
+  static manifest = Object.freeze({
+    createdBy: {
+      type: 'okapi',
+      records: 'users',
+      path: 'users?query=(id==!{metadata.createdByUserId})',
+      fetch: false,
+      accumulate: true,
+    },
+    updatedBy: {
+      type: 'okapi',
+      records: 'users',
+      path: 'users?query=(id==!{metadata.updatedByUserId})',
+      fetch: false,
+      accumulate: true,
+    },
+  });
+
   static propTypes = {
     metadata: PropTypes.object,
-    stripes: stripesShape.isRequired,
+    mutator: PropTypes.shape({
+      createdBy: PropTypes.shape({
+        GET: PropTypes.func.isRequired,
+        reset: PropTypes.func.isRequired,
+      }).isRequired,
+      updatedBy: PropTypes.shape({
+        GET: PropTypes.func.isRequired,
+        reset: PropTypes.func.isRequired,
+      }).isRequired,
+    }).isRequired,
+    systemId: PropTypes.string,
     systemUser: PropTypes.oneOfType([
       PropTypes.shape({
         id: PropTypes.string,
@@ -34,10 +57,10 @@ class ViewMetaData extends React.Component {
   };
 
   componentDidMount() {
-    this.fetchUsers();
+    this.getUsers();
   }
 
-  async fetchUsers() {
+  async getUsers() {
     const {
       metadata: {
         createdByUserId,
@@ -47,7 +70,7 @@ class ViewMetaData extends React.Component {
 
     // fetch once if users are identical
     if (createdByUserId === updatedByUserId) {
-      const user = await this.fetchUser(createdByUserId);
+      const user = await this.getCreatedByUser();
 
       this.setState({
         createdBy: user,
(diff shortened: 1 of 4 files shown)


lib/ViewMetaData/ViewMetaData.js
```diff
@@ -1,52 +1,133 @@
...
+        }),
       }),
-    }),
+      PropTypes.string,
+    ]),
   };
 
   static defaultProps = {
     metadata: {},
+    systemUser: 'System',
```
doytch: Can we pull this from a translation file via `FormattedMessage` instead? (And I guess update `propTypes` to support `node` as well.

OleksandrYehorovEPAM: This property is passed to `MetaSection` component so I took PropTypes from there. And `MetaSection` component will not handle `node` correctly.

doytch: In that case we could just update `MetaSection`'s proptypes to accept a `node` and tweak it's `renderUser` function, no?

We'll definitely need translations for "System" though. I just think that a `FormattedMessage` is the cleanest way of doing it when we don't really _have_ to get it into a string.

zburke: Agreed; we need translation here. That should be a trivial change in stripes-components, and then we'll need to get that released and bump the dep in the package.json here as well.

doytch: Alas, it's not completely trivial because of the type-checking (outside of proptype checking) that `MetaSection`'s `renderUser` function does, but definitely still not a burden. _Is_ a "gotcha" though.

OleksandrYehorovEPAM:
@zburke @doytch @VictorSoroka1 I added internationalization for the System user and I also included a fix for https://issues.folio.org/browse/STSMACOM-177 bug.

VictorSoroka1:
@OleksandrYehorovEPAM, please add a note to the changelog and create a readme file to describe the idea behind provided changes.

lib/ViewMetaData/ViewMetaData.js
zburke: Hmmmmm. Seeing the ripple effect of [providing i18n support in MetaSection](https://github.com/folio-org/stripes-components/pull/908), I'm increasingly wary of that approach as it clearly relies on their being a hard-coded identifier that corresponds to the "system" user, thus the need for "system" to be localized, rather than there being a user, just like any other, which one tenant could name "system" and another could name "система".

I feel badly that you are doing everything we are asking here, quickly and elegantly to boot, and I'm still hesitant to merge the code. I just really worry that by coding this up, we are implicitly endorsing an approach to handling the system user problem when in fact how to handle that problem is still undecided. 

I dunno; maybe I'm over reacting. @doytch, any thoughts? As you said above, "magic numbers ... yuck, but you gotta do what you gotta do."

zburke: (context for "the problem is still undecided": [RMB-320](https://issues.folio.org/browse/RMB-320))

VictorSoroka1: @zburke, please make a decision about a merge sooner as we need these changes anyways in `ui-data-import` module. In case you are not agree on these changes we simply create this component in our module. But I think this could be the case not only for `ui-data-import` app but for any app where a similar issue may occur. I do not think that resolving the same problem per component is a good solution.

doytch:
I hate the fact that we have magic numbers, and I'll be on anyone's side who wants to push back against it on a Folio level.

According to RMB-320, the decision about whether to use the magic number or simply make the user ID an optional field is still unresolved.

But if this functionality is needed _now_ in the front-end, I'm cool to merge this because if we move away from magic numbers on the backend, the proposed implementation in this PR can be adjusted fairly seamlessly.

VictorSoroka1:
@doytch, I agree with your concerns, it would be great to have that `RMB-320` resolved. It was raised in development chat with no outcome so far and seems there would not be an answer soon. The proposed implementation won't break when the issue will be resolved as now in order to handle that use case the `systemId` should be provided. I am inclined to have that field optional so we do not have that hardcoding `systemId` in that shared module and rather have it in our own module as each backend module may use another value for that. In case I find that provided changes here does not make sense due to the resolution of `RMB-320` our team would be happy to remove that functionality from that public component. And of course, we will track that `RMB-320` issue for that purpose.

OleksandrYehorovEPAM:
@zburke @doytch please let me know if you are okay with merging. If no I will just add that functionality in `ui-data-import` module. Please also note that this PR also includes a bugfix https://issues.folio.org/browse/STSMACOM-177 and as already mentioned it will not break the app when `RMB-320` will be done.

PR #454, opened 2019-03-08
STSMACOM-179 <SearchAndSortQuery>

JohnC-80:
Placing this here until after Q1 releases for visibility's sake.

# Problem
Since `<SearchAndSort>` was first created, we've been able to easily assemble the foundational pattern for most FOLIO UI modules... but when it came to modification, non-conventional requirements, and all around need for more control at the module level, `<SearchAndSort>` ultimately left much to be desired in terms of flexibility. 

## Approach
Bake `<SearchAndSort>` down to the primary needs of its namesake: facilitate the creation of queries that can be flexible in both the presentational controls and in the ultimate application of the query within the application.

`<SearchAndSortQuery>` maintains an internal state of query-building slices: `searchFields`, `filterFields`, `sortFields`. The keys for these slices are ultimately combined, mapped, manipulated into a usable query string.

It supplies flexible control for both getting the current query and setting a new one, as well as fine adjustment to how changes might affect the `query state`.

Render props are used to relay current query state and supply callbacks for any shape of search/filter/sorting components. At the same time, migration path has been paved so that it's not too difficult to carry FilterGroups right over.

## This PR doesn't actually break anything.

It modifies some of the building-blocks (`StripesConnectedResource`) that `<SearchAndSort>` uses and makes them generic for usage at the module level, but that's it as far as changes to existing code.

 In the very near future, the existing `<SearchAndSort>` could actually use this component internally.

Please have a look at the [Readme](https://github.com/folio-org/stripes-smart-components/blob/uiu-896-s-and-s-query/lib/SearchAndSort/SearchAndSortQuery.md) for a usage example and deeper documentation.

Additionally, you can see this component in action over on a PR for `ui-plugin-find-user` that removes the dependency on the `ui-users` module. [Check it out ](https://github.com/folio-org/ui-plugin-find-user/blob/765019f7184250ddbc8fad98becc468cd997a477/UserSearch/UserSearchContainer.js)


MikeTaylor:
I won't presume to actually review this, but I do want to say that I endorse the overall approach _provided that_, as John says, it doesn't break anything — that the old `<SearchAndSort>` API continues to be supported, ideally by a refactoring to use this new component.

PR #479, opened 2019-04-16
improve metadata display

zburke:
Complete reworking of the metadata display. The previous implementation
was so broken it's not even worth trying to describe how it's changed
except to say that it doesn't suck anymore.

This implementation gathers the list of updaters from the list of
entries and then makes a single query to the `users` endpoint to
retrieve all the records. It also checks permissions _before_ doing so,
in order to avoid attempting (and failing) to retrieve data the end-user
doesn't have permission to read.

Refs [STSMACOM-181](https://issues.folio.org/browse/STSMACOM-181), [STSMACOM-182](https://issues.folio.org/browse/STSMACOM-182), [STSMACOM-183](https://issues.folio.org/browse/STSMACOM-183), [UIU-873](https://issues.folio.org/browse/UIU-873), [UIIN-462](https://issues.folio.org/browse/UIIN-462).


lib/ControlledVocab/ControlledVocab.js
```diff
@@ -136,6 +153,30 @@ class ControlledVocab extends React.Component {
...
   }
 
+  componentDidUpdate() {
+    // build a list of the ids of users who have updated CV items so
+    // we can look them up all at once
+    const { stripes, resources: { values: { records } } } = this.props;
+    if (stripes.hasPerm('ui-users.view')) {
+      // convert the list of values-plus-metadata to a de-duped list of
+      // metadata-userids
+      const userIds = [...new Set(records
```
mkuklis: @zburke it's nice to see the Set here :)

lib/ControlledVocab/ControlledVocab.js
```diff
@@ -136,6 +153,30 @@ class ControlledVocab extends React.Component {
...
+        .map(r => `id=="${r.metadata.updatedByUserId}"`))
+      ];
+
+      // only query if we have users to look for
+      // when loading system data; the list may be empty!
+      // quiet, lint, I know what I'm doing. yes, I know this will kick
+      // off another render. THAT'S EXACTLY WHAT I'M TRYING TO DO.
+      // so helpful, you are.
+      if (!isEqual(userIds, this.state.userIds)) {
+        this.setState({ userIds }); // eslint-disable-line react/no-did-update-set-state
```
mkuklis: I wonder do we need to rely on the local state here? Could we read the values from the `updaterIds` resource directly?

zburke: You're right, reading directly from the resource is simpler.

PR #480, opened 2019-04-17
STCOM-451: Create date range filter

MaksymDryha:
This is a component for filtering records by a date range.
It has almost the same interface as CheckboxFilter and MultiSelectionFilter have. Unlike other filter components, this one does not require a `dataOptions` prop to be provided since the dates are manually entered by users and cannot be predefined. 
Another difference is that a `makeFilterString` function should be provided. When a user enters a date range, this function receives entered start and end dates and returns a string which will be used in the request parameters. Probably it can be given a better name.

Here is an example of using the component (inspired by ui-search):
```javascript
export default class Filters extends React.Component {
  static propTypes = {
    activeFilters: PropTypes.objectOf(PropTypes.array),
    onChange: PropTypes.func.isRequired,
  };

  onChangeHandler = (filter) => {
    this.props.onChange(filter);
  }

  retrieveDatesFromDateRangeFilterString(filterValue) {
    let dateRange = {
      startDate: '',
      endDate: '',
    };

    if (filterValue) {
      const [startDate, endDate] = filterValue[0].split(':');
      dateRange = {
        startDate: startDate || '',
        endDate: endDate || '',
      };
    }

    return dateRange;
  }

  makeDateRangeFilterString(startDate, endDate) {
    /**
    returned string will be passed to onChangeHandler and then used
    as the filter's value in the address bar.
    Given startDate is 1997-03-01 and endDate is 2000-01-21,
    then the address bar will look like:
    http://localhost:8080/codexsearch?filters=dateRange.1997-03-01%3A2000-01-21
    */
    return `${startDate}:${endDate}`; 
  }

  render() {
    const {
      activeFilters,
    } = this.props;

    return (
      <Accordion
        label="Date range"
        id="dateRange"
        name="dateRange"
      >
       <DateRangeFilter
          name="dateRange"
          selectedValues={this.retrieveDatesFromDateRangeFilterString(activeFilters.dateRange)}
          onChange={this.onChangeHandler}
          makeFilterString={this.makeDateRangeFilterString}
          dateFormat="YYYY-MM-DD"
        />
      </Accordion>
    );
  }
}
```

### DIFF
--- a/lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
+++ b/lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
@@ -85,10 +85,9 @@ export default class DateRangeFilter extends Component {
     const dateRangeIsNotEmpty = startDate !== '' || endDate !== '';
     const validationData = dateRangeIsNotEmpty
       ? validateDateRange(props.selectedValues)
-      : {};
+      : defaultFilterState;
 
     this.state = {
-      ...defaultFilterState,
       ...validationData,
       prevSelectedDates: {
         startDate,
@@ -99,7 +98,7 @@ export default class DateRangeFilter extends Component {
 
   onApplyButtonClick = () => {
     const {
-      handleValidation: afterValidation,
+      applyFilterIfValid,
       props: { dateFormat },
     } = this;
 
@@ -122,11 +121,11 @@ export default class DateRangeFilter extends Component {
       this.clearFilter();
     } else if (!areBothDatesEmpty) {
       const dateValidationData = validateDateRange(dateRange, dateFormat);
-      this.setState(dateValidationData, afterValidation);
+      this.setState(dateValidationData, applyFilterIfValid);
     }
   }
 
-  handleValidation = () => {
+  applyFilterIfValid = () => {
     const { isDateRangeValid } = this.state;
 
     if (isDateRangeValid) {
--- a/lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
+++ b/lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
@@ -30,7 +30,7 @@ export default class DateRangeFilter extends Component {
     makeFilterString: PropTypes.func.isRequired,
     name: PropTypes.string.isRequired,
     onChange: PropTypes.func.isRequired,
-    selectedDates: PropTypes.shape({
+    selectedValues: PropTypes.shape({
       endDate: PropTypes.string.isRequired,
       startDate: PropTypes.string.isRequired,
     }).isRequired,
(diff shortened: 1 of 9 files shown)


lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
```diff
@@ -0,0 +1,285 @@
...
+    const {
+      dateFormat,
+      selectedValues: nextSelectedDates,
+    } = nextProps;
+
+    const { prevSelectedDates } = prevState;
+
+    let validationDataUpdates = {};
+
+    const areSelectedDatesSetFirstTime =
```
Godlevskyi: Please cover cases with just one date recieved

lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
```diff
@@ -0,0 +1,285 @@
...
+      endDate,
+    } = props.selectedValues;
+
+    this.startDateInput = React.createRef();
+    this.endDateInput = React.createRef();
+
+    const dateRangeIsNotEmpty = startDate !== '' || endDate !== '';
+    const validationData = dateRangeIsNotEmpty
+      ? validateDateRange(props.selectedValues)
+      : {};
```
Godlevskyi: move default state

lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
```diff
@@ -0,0 +1,285 @@
...
+
+    this.state = {
+      ...defaultFilterState,
```
Godlevskyi: remove defaultFilterState

lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
```diff
@@ -0,0 +1,285 @@
...
+      prevSelectedDates: {
+        startDate,
+        endDate,
+      },
+    };
+  }
+
+  onApplyButtonClick = () => {
+    const {
+      handleValidation: afterValidation,
```
Godlevskyi: Please rename to applyFiltersIfValid

lib/SearchAndSort/components/DateRangeFilter/date-validations.js
```diff
@@ -0,0 +1,42 @@
+import moment from 'moment';
+
+export const isDateValid = (date, dateFormat) => {
+  const momentDateWrapper = moment(date, dateFormat, true);
+
+  return momentDateWrapper.isValid();
+};
+
+export const validateDateRange = (dateRange, dateFormat) => {
```
zburke: This is really nicely implemented. It's clear and easy to follow, and makes it easy to tell exactly which conditions failed. Nice work!

MaksymDryha: Thank you @zburke!

PR #481, opened 2019-04-17
STSMACOM-185: Add notes accordion

Yurii-Danylenko:
https://issues.folio.org/browse/STSMACOM-185

## Purpose
Create basic accordion for notes with MCL table,  "Add" and New"  buttons, which will be used to manage work with notes inside the UI module where notes used. It's the beginning part of the whole component. Next step is to add embed modal with possibility to search and attach notes.

## Approach
- Create Accordion
- Create NotesList component
- Add translations

## TODO
- Add modal component for attaching notes

### DIFF
--- a/lib/Notes/NotesAccordion/NotesAccordion.css
+++ b/lib/Notes/NotesAccordion/NotesAccordion.css
@@ -1,0 +1,3 @@
+a.new-button {
+  margin-left: 10px;
+}
--- a/lib/Notes/NotesAccordion/NotesAccordion.js
+++ b/lib/Notes/NotesAccordion/NotesAccordion.js
@@ -3,9 +3,6 @@ import PropTypes from 'prop-types';
 import {
   FormattedMessage,
 } from 'react-intl';
-import {
-  get,
-} from 'lodash';
 
 import {
   Accordion,
@@ -17,6 +14,30 @@ import {
 import NotesList from './components';
 import styles from './NotesAccordion.css';
 
+const defaultColumnsConfig = [
+  {
+    name: 'date',
+    title: <FormattedMessage id="stripes-smart-components.date" />,
+    width: '30%',
+  },
+  {
+    name: 'updatedBy',
+    title: <FormattedMessage id="stripes-smart-components.updatedBy" />,
+    width: '30%',
+  },
+  {
+    name: 'title',
+    title: <FormattedMessage id="stripes-smart-components.title" />,
+    width: '40%',
+  },
+];
+
+export const columnShape = PropTypes.shape({
+  name: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
+  width: PropTypes.string,
+});
+
 export const noteShape = PropTypes.shape({
   id: PropTypes.string,
   lastSavedDate: PropTypes.instanceOf(Date),
@@ -26,10 +47,23 @@ export const noteShape = PropTypes.shape({
 
 export class NotesAccordion extends Component {
   static propTypes = {
+    columnsConfig: PropTypes.arrayOf(columnShape),
     notes: PropTypes.arrayOf(noteShape),
     onCreate: PropTypes.func.isRequired,
     onNoteClick: PropTypes.func.isRequired,
     open: PropTypes.bool,
+  }
+
+  static defaultProps = {
+    notes: [],
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.columnsConfig = Array.isArray(props.columnsConfig)
+      ? props.columnsConfig
+      : defaultColumnsConfig;
   }
 
   renderHeader = () => {
@@ -51,9 +85,7 @@ export class NotesAccordion extends Component {
 
     return (
       <Fragment>
-        <Button
-          data-test-notes-accordion-add-button
-        >
+        <Button data-test-notes-accordion-add-button>
           <FormattedMessage id="stripes-smart-components.add" />
         </Button>
         <Button
@@ -67,15 +99,11 @@ export class NotesAccordion extends Component {
     );
   }
 
-  getNotesQuantity() {
-    return get(this.props.notes, 'length');
-  }
-
   renderQuantityIndicator() {
     return (
       <Badge>
         <span data-test-notes-accordion-quantity-indicator>
-          {this.getNotesQuantity()}
+          {this.props.notes.length}
         </span>
       </Badge>
     );
@@ -97,6 +125,7 @@ export class NotesAccordion extends Component {
         displayWhenOpen={this.renderHeaderButtons()}
       >
         <NotesList
+          columnsConfig={this.columnsConfig}
           notes={notes}
           onNoteClick={onNoteClick}
         />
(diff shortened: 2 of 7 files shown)


lib/Notes/NotesAccordion/components/NotesList/NotesList.js
```diff
@@ -0,0 +1,83 @@
...
+  date: '30%',
+  updatedBy: '30%',
+  title: '40%',
+};
+
+const columnsMap = {
+  date: <FormattedMessage id="stripes-smart-components.date" />,
+  updatedBy: <FormattedMessage id="stripes-smart-components.updatedBy" />,
+  title: <FormattedMessage id="stripes-smart-components.title" />,
+};
```
JohnC-80: Depends on just how generic you'd want to make this component - consider setting up the `column` variables - `COLUMN_NAMES`, `COLUMN_WIDTHS` and `columnsMap` as `defaultProps` so it could accept whichever configuration other modules might need to use.

lib/Notes/NotesAccordion/NotesAccordion.js
```diff
@@ -0,0 +1,106 @@
...
+          onClick={onCreate}
+        >
+          <FormattedMessage id="stripes-smart-components.new" />
+        </Button>
+      </Fragment>
+    );
+  }
+
+  getNotesQuantity() {
+    return get(this.props.notes, 'length');
```
MaksymDryha: I think if we make `notes` prop default to `[]`, we'll be able to write just `this.props.notes.length` here and `lodash.get` won't be necessary. This way `getNotesQuantity` will return `0` in case when there is no notes, which is more sensible than `undefined`.
Also I guess `getNotesQuantity` will be unnecessary too. You will be able to just use `this.props.notes.length` inside of the `renderQuantityIndicator`:
```js
renderQuantityIndicator() {
    return (
      <Badge>
        <span data-test-notes-accordion-quantity-indicator>
          {this.props.notes.length}
        </span>
      </Badge>
    );
  }
```

lib/Notes/NotesAccordion/NotesAccordion.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  renderQuantityIndicator() {
+    return (
+      <Badge>
+        <span data-test-notes-accordion-quantity-indicator>
+          {this.getNotesQuantity()}
```
MaksymDryha: Considering my previous comment, you could probably `have this.props.notes.length` instead of `this.getNotesQuantity()` here

lib/Notes/NotesAccordion/NotesAccordion.js
```diff
@@ -0,0 +1,106 @@
...
+    return (
+      <Accordion
+        id="notes-accordion"
+        open={open}
+        label={this.renderHeader()}
+        displayWhenClosed={this.renderQuantityIndicator()}
+        displayWhenOpen={this.renderHeaderButtons()}
+      >
+        <NotesList
+          notes={notes}
```
MaksymDryha: From `NotesList`'s prop types I see that `notes` prop is required. However, in `NotesAccordion`'s prop types `notes` prop is not required, which means it can be `undefined`. Also [this string](https://github.com/folio-org/stripes-smart-components/pull/481/files#diff-8c2a5babb8d6fd2cc9f8396d17f3856fR71) means that `notes` prop can be `undefined`. Therefore `notes` should be passed to `NoteList` like this: `notes={notes || []}`. But I would better prefer to make the `note` prop default to `[]`.

lib/Notes/NotesAccordion/NotesAccordion.js
```diff
@@ -0,0 +1,106 @@
...
+  Badge,
+  Button,
+} from '@folio/stripes-components';
+
+import NotesList from './components';
+import styles from './NotesAccordion.css';
+
+export const noteShape = PropTypes.shape({
+  id: PropTypes.string,
+  lastSavedDate: PropTypes.instanceOf(Date),
```
MaksymDryha: I think this shape should be moved to a separate file, since it's not specific to NotesAccordion and can be used in different components

Yurii-Danylenko: Will consider to move the shape out during adding of the other Notes specific components

lib/Notes/NotesAccordion/NotesAccordion.js
```diff
@@ -0,0 +1,135 @@
...
+
+  static defaultProps = {
+    notes: [],
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.columnsConfig = Array.isArray(props.columnsConfig)
+      ? props.columnsConfig
```
Godlevskyi: Why not to use the defaultProps for it?

lib/Notes/NotesAccordion/components/NotesList/NotesList.js
```diff
@@ -0,0 +1,90 @@
...
+              day="numeric"
+            />
+          ),
+          updatedBy: lastSavedUserFullName,
+          title,
+        };
+      });
+  }
+
+  getColumnNames = memoize((config) => {
```
Godlevskyi: Do not see the purpose to use memoize here as the number of columns can not be too high

lib/Notes/NotesAccordion/components/NotesList/NotesList.js
```diff
@@ -0,0 +1,90 @@
...
+export default class NotesList extends React.Component {
+  static propTypes = {
+    columnsConfig: PropTypes.arrayOf(columnShape),
+    notes: PropTypes.arrayOf(noteShape).isRequired,
+    onNoteClick: PropTypes.func.isRequired,
+  }
+
+  getItems() {
+    return this.props.notes
+      .map(note => {
```
Godlevskyi: I belive it should be decomposed to noteItem functional component

lib/Notes/NotesAccordion/NotesAccordion.css
```diff
@@ -0,0 +1,3 @@
+a.new-button {
```
MaksymDryha: is `a` needed?

Yurii-Danylenko: Unfortunately, it's needed to override specific selector

lib/Notes/NotesAccordion/components/NotesList/NotesList.js
Yurii-Danylenko: Unfortunately, it returns only an object with data. So definitely, it's better to keep it in a function, but not in a component

lib/Notes/NotesAccordion/NotesAccordion.js
```diff
@@ -0,0 +1,129 @@
...
+  {
+    name: 'title',
+    title: <FormattedMessage id="stripes-smart-components.title" />,
+    width: '40%',
+  },
+];
+
+export const columnShape = PropTypes.shape({
+  name: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
```
JohnC-80: This should be `PropTypes.node`, instead of `PropTypes.string` right? due to the `<FormattedMessage>` use here.

PR #482, opened 2019-04-17
provide a11y props to Layer

zburke:
The new-entry layer was missing the `contentLabel` prop, necessary for
adequate a11y labeling.


MikeTaylor:
w00t

PR #483, opened 2019-04-17
refactor to simplify, avoid permissions issues

zburke:
The structure of the previous iteration was complicated and did not
check permissions before running queries, which caused trouble for users
without access to the `users` resource. This refactoring follows a
pattern similar to the recent refactoring of `<ControlledVocab>` in
#479.


lib/ViewMetaData/ViewMetaData.js
```diff
@@ -1,37 +1,43 @@
...
+          query: (queryParams, pathComponents, resourceValues) => {
+            if (resourceValues.updaterIds && resourceValues.updaterIds.length) {
+              return `(${resourceValues.updaterIds.join(' or ')})`;
+            }
+            return null;
+          },
+        }
+      }
     },
+    updaterIds: [],
```
doytch: Is this supposed to be an `{}` or is this some syntax I don't know about?

zburke: It's supposed to be an empty array, but in fact local resources are either objects or strings, so `{}` is correct.

PR #495, opened 2019-05-01
Remove .isRequired from 'records' propType

MikeTaylor:
The actual code already allows for the situation where no value is
passed in for `records`, i.e. the top level of the JSON returned from
the WSAPI is an array of records rather than an object containing a
member whose value is this array. Since this works correctly, a valid
invocation should not yield a console error.


MikeTaylor:
You know what? I'm gonna go ahead an presume that @zburke's review is going to be positive.

MikeTaylor:
That's actually a surprisingly accurate summary of what's happened here.

PR #498, opened 2019-05-03
STSMACOM-188, STCOM-505: Create NoteForm component

MaksymDryha:
## Purpose
Different applications may want to have the ability to create/edit notes and attach them to their entities. `<NoteForm />` is a reusable dumb component which apps can use for editing and creating a note. It provides the whole layout for create/edit note pages, including all needed panes, accordions, fields, buttons and the form itself.

## Screens

**Create note page**
image

**Edit note page**
image

Related stories: [STCOM-505](https://issues.folio.org/browse/STCOM-505), [STSMACOM-188](https://issues.folio.org/browse/STSMACOM-188)

## TODO
Discuss where `NavigationModal` component should live. Previously, it was used only by eholdings and was stored in ui-eholdings repository. Now, it's also used by NoteForm and is no longer eholdings-specific. For now, it was temporarily moved to stripes-smart-components, next to all notes-related components

PR #504, opened 2019-05-10
STSMACOM-195: Added note view component

Godlevskyi:
## Purpose
This PR is complimentary for https://github.com/folio-org/stripes-smart-components/pull/498

To add dumb component to use it on the note view page.


lib/Notes/NoteForm/NoteView.js
```diff
@@ -239,8 +242,9 @@ export default class NoteView extends Component {
                   <Col xs={12}>
                     <KeyValue
                       label={<FormattedMessage id="stripes-smart-components.details" />}
-                      value={get(noteData, 'content', '-')}
-                    />
+                    >
+                      <div dangerouslySetInnerHTML={noteContentMarkup} />
```
doytch: Do we have a way of ensuring that the markup is safe? White-listed tags etc to ensure there aren't `<script>` tags or similar?

Godlevskyi: Hi @doytch, this markup is saved with the '<Editpr/>' component and it sanitize it removing the <script> tags.
And on this page we receive this markup from our backend and not user input nor the 3rd party. We rely on our backend.

PR #516, opened 2019-05-21
UIINV-6 support row parse

NikitaSedyx:
**ticket**: https://issues.folio.org/browse/UINV-6

**purpose**: tenant addresses will work with `mod-configs` and this module supports only `value` field. In this case value will be a JSON of `name` and  `address` so we need a way to parse value to work with this component in right way

**approach**: new property to parse row

related to https://github.com/folio-org/ui-tenant-settings/pull/17

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -82,6 +82,7 @@ class ControlledVocab extends React.Component {
     }).isRequired,
     nameKey: PropTypes.string,
     objectLabel: PropTypes.node.isRequired,
+    parseRow: PropTypes.func,
     preCreateHook: PropTypes.func,
     preUpdateHook: PropTypes.func,
     readOnlyFields: PropTypes.arrayOf(PropTypes.string),
@@ -221,6 +222,14 @@ class ControlledVocab extends React.Component {
     return rows.filter(row => this.props.rowFilterFunction(row));
   }
 
+  parseRows(rows) {
+    if (!this.props.parseRow) {
+      return rows;
+    }
+
+    return rows.map(this.props.parseRow);
+  }
+
   hideConfirmDialog() {
     this.setState({
       showConfirmDialog: false,
@@ -357,7 +366,9 @@ class ControlledVocab extends React.Component {
       />
     );
 
-    const rows = this.filteredRows(this.props.resources.values.records || []);
+    const rows = this.parseRows(
+      this.filteredRows(this.props.resources.values.records || [])
+    );
 
     const hideList = this.props.listSuppressor && this.props.listSuppressor();
 


lib/ControlledVocab/ControlledVocab.js
```diff
@@ -357,7 +366,9 @@ class ControlledVocab extends React.Component {
       />
     );
 
-    const rows = this.filteredRows(this.props.resources.values.records || []);
+    const rows = this.parseRows(
+      this.filteredRows(this.props.resources.values.records || [])
```
aliaksei-chumakou: Don't you think that `resources` could be `null` at some point? Maybe it makes sense to put here `_.get(this.props.resources, 'values.records', [])`

NikitaSedyx: agreed

NikitaSedyx: just noticed that we have https://github.com/folio-org/stripes-smart-components/pull/516/files#diff-6a43094b2f04c9ea138bb80cc1d5352dR357

with it `_.get` doesn't make sense here

aliaksei-chumakou: That's OK, but in case it's not loaded, `values` might exists, and records could be undefined, right?

aliaksei-chumakou: Moreover, I think `this.props.resources` could be null at some point.

NikitaSedyx: it's required property, up to developer

zburke: `props.resources` isn't provided by the developer; it's provided by `stripes-connect` when it processes the manifest when this component has `connect` called on it. Using @aliaksei-chumakou's suggestion of `_.get(this.props.resources, 'values.records', [])` or `this.props.resources.values || {}).records || []` are both common patterns.

NikitaSedyx: @zburke but developer writes code and it's his responsibility to control all props are passed in right way: `resources` is required property so developer should do everything required from him to pass it to the component

suggestion won't work because of https://github.com/folio-org/stripes-smart-components/blob/657ea72111bf5bc3562fb055d6ed62e3fe222141/lib/ControlledVocab/ControlledVocab.js#L359 I commented previously

zburke: Your comment that a "developer writes code and its his responsibility to control all props" is technically correct but suggests a misunderstanding of how this component was intended to be used. It would be like describing a sword as perfect for slicing a block of cheese; it's not technically wrong, but we don't typically use swords in the kitchen! Sure, in principle, `resources` could be manually provided by a developer who passes it to an instance of this component. In practice, this prop is never manually passed; [it is provided by stripes-connect](https://github.com/folio-org/stripes-connect/blob/master/doc/api.md#using-the-connected-component) when the component is connected. 

All that said, you're correct that there is already a check in place above at line 359 so the check here is safe. Because that code was already in place it was not part of the diff output and therefore easy to miss. Sorry for dragging this out when your code was technically correct to start with. This code does not follow the common idiom of doing all the validation of `...resources.values.records` in a single line and that threw me off. Please consider this idiom in future PRs.

PR #520, opened 2019-05-22
UIOR-27 navigation for SearchAndSort

NikitaSedyx:
https://issues.folio.org/browse/UIOR-27

purpose: we need a place to put navigation buttons in the first column
navigation

approach: new render prop

related https://github.com/folio-org/ui-orders/pull/405

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -158,6 +158,7 @@ class SearchAndSort extends React.Component {
     path: PropTypes.string,
     queryFunction: PropTypes.func,
     renderFilters: PropTypes.func,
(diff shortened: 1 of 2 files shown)


doytch:
It seems like the complete set of changes that have necessitated this feature will be pretty extensive. On the ERM apps, we're working through UX suggestions like the ones you got for Orders and I'm planning on just refactoring to use `SearchAndSortQuery` and [the new routing guidelines](https://github.com/folio-org/stripes/blob/master/doc/dev-guide.md#guidelines-for-structuring-your-stripes-module) rather than further ballooning the API of SAS.

@JohnC-80 ?

NikitaSedyx:
@doytch seems we need this ASAP for Chalmers

PR #526, opened 2019-05-28
Fix bug with unclosing note form dropdown

MaksymDryha:
This small PR adds forgotten `onToggle` function call to dropdown item click handler

### DIFF
--- a/lib/Notes/NoteForm/NoteForm.js
+++ b/lib/Notes/NoteForm/NoteForm.js
@@ -143,14 +143,17 @@ export default class NoteForm extends Component {
     );
   }
 
-  getActionMenu = () => {
+  getActionMenu = ({ onToggle }) => {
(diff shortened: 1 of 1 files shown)


lib/Notes/ReferredRecord/ReferredRecord.css
```diff
@@ -5,4 +5,3 @@
   font-weight: bold;
   color: var(--color-text);
 }
-
```
Godlevskyi: What for this change? Please check Jenkins comment on this

MaksymDryha: This change is to fix that comment

MaksymDryha: @Godlevskyi, there were 2 empty lines

PR #528, opened 2019-05-29
Notes: Fixed referrer data and it's shape

Godlevskyi:
This PR fixes referrer data on the note-view page in case when we navigate directly to the note and not referrer record data is present

### DIFF
--- a/lib/Notes/NoteForm/NoteView.js
+++ b/lib/Notes/NoteForm/NoteView.js
@@ -44,7 +44,10 @@ export default class NoteView extends Component {
     onEdit: PropTypes.func.isRequired,
     onUnassign: PropTypes.func.isRequired,
     paneHeaderAppIcon: PropTypes.string.isRequired,
-    referredEntityData: referredEntityDataShape,
+    referredEntityData: PropTypes.oneOfType([
+      referredEntityDataShape,
+      PropTypes.bool
+    ]),
   };
 
   state = {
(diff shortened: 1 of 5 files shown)


lib/Notes/NoteForm/NoteView.js
```diff
@@ -234,10 +240,12 @@ export default class NoteView extends Component {
...
                 onToggle={this.handleSectionToggle}
               >
-                <ReferredRecord
-                  referredEntityData={referredEntityData}
-                  entityTypeTranslationKeys={entityTypeTranslationKeys}
-                />
+                { referredEntityData &&
+                  <ReferredRecord
+                    referredEntityData={referredEntityData}
+                    entityTypeTranslationKeys={entityTypeTranslationKeys}
```
MaksymDryha: please move it to render method for consistency with NoteForm component https://github.com/folio-org/stripes-smart-components/blob/master/lib/Notes/NoteForm/NoteForm.js#L132

lib/Notes/NoteForm/noteShapes.js
```diff
@@ -10,11 +10,14 @@ export const noteDataShape = PropTypes.shape({
...
-  name: PropTypes.string.isRequired,
-  type: PropTypes.string.isRequired,
-});
+export const referredEntityDataShape = PropTypes.oneOfType([
+  PropTypes.shape({
+    id: PropTypes.string.isRequired,
+    name: PropTypes.string.isRequired,
+    type: PropTypes.string.isRequired,
+  }),
+  PropTypes.bool
```
MaksymDryha: I'm wondering if `PropTypes.bool` is needed. If yes, then I think it's better to use `PropTypes.bool` in components' prop-types:
```js
referredEntityData: PropTypes.oneOfType([ referredEntityDataShape, PropTypes.bool ])
```

Godlevskyi: agree, good point

PR #532, opened 2019-05-31
STSMACOM-210: Disallow unassigning of the note if links number is equal to 1

Yurii-Danylenko:


lib/Notes/NotesAccordion/components/NotesAssigningModal/NotesAssigningModal.css
```diff
@@ -9,8 +9,8 @@
 }
 
 .assign-checkbox label::before {
-  margin-left: 10px;
-  margin-right: 10px;
+  margin-left: 5px;
```
Godlevskyi: Is it possible to use variables here?

PR #539, opened 2019-06-06
UIDATIMP-10: add TagsAccordion component

OleksandrYehorovEPAM:
Jira ticket: https://issues.folio.org/browse/UIDATIMP-10

The purpose of this PR is to add `TagsAccordion` component to provide an ability to use tags in an accordion (like in `ui-eholdings` module).
The PR also includes:
1) a bug fix for displaying a toast when creating a new tag;
1) code refactoring for the `Tags` and `TagsForm` components.
1) a "render props" support for `Tags` component to provide a way to use tags independently of a markup.

Note: there is a "refreshRemote"-hack in the `componentDidUpdate` method in `Tags` component. It is needed to update resources when the `link` prop changes and it should be removed when https://issues.folio.org/browse/STCON-81 is done.

### DIFF
--- a/lib/Tags/Tags.js
+++ b/lib/Tags/Tags.js
@@ -1,12 +1,14 @@
-import { get, uniq, sortBy, difference } from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+import { get, uniq, sortBy, difference, noop, isFunction } from 'lodash';
+
 import { Callout, Pane } from '@folio/stripes-components';
+import { stripesConnect } from '@folio/stripes-core';
 
 import TagsForm from './TagsForm';
 
-export default class Tags extends React.Component {
+class Tags extends React.Component {
   static manifest = Object.freeze({
     tags: {
       type: 'okapi',
@@ -25,6 +27,8 @@ export default class Tags extends React.Component {
   });
 
   static propTypes = {
+    children: PropTypes.func,
+    link: PropTypes.string.isRequired,
     mutator: PropTypes.shape({
       entities: PropTypes.shape({
         PUT: PropTypes.func.isRequired,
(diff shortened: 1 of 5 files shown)


MikeTaylor:
Hi, @OleksandrYehorovEPAM. Just to let you know I'm not technically part of the FOLIO project any more, so I shouldn't really be tagged in review requests. I don't _mind_ — just know that I won't always respond, so don't ever rely on me for reviews!

OleksandrYehorovEPAM:
> Hi, @OleksandrYehorovEPAM. Just to let you know I'm not technically part of the FOLIO project any more, so I shouldn't really be tagged in review requests. I don't _mind_ — just know that I won't always respond, so don't ever rely on me for reviews!

Okay, sure.

lib/Tags/Tags.js
```diff
@@ -42,100 +47,117 @@ export default class Tags extends React.Component {
...
-    const tags = this.getEntityTags();
-    const tagList = tags.filter(t => (t !== tag));
-    entity.tags = { tagList };
-    this.props.mutator.entities.PUT(entity);
-  }
-
   getEntity() {
-    const entities = (this.props.resources.entities || {}).records || [];
-    return entities[0] || {};
+    return get(this.props, ['resources', 'entities', 'records', 0], {});
```
MaksymDryha: I guess `resources` will always be present so so you can do
`get(this.props.resources, ['entities', 'records', 0], {})`

lib/Tags/Tags.js
```diff
@@ -42,100 +47,117 @@ export default class Tags extends React.Component {
...
   }
 
   getEntityTags() {
     const entity = this.getEntity();
+
     return get(entity, ['tags', 'tagList'], []);
   }
 
+  getTags() {
+    return get(this.props, ['resources', 'tags', 'records'], []);
```
MaksymDryha: same here

lib/Tags/Tags.js
OleksandrYehorovEPAM: Thank you, but I would prefer to leave it as is, just in case.

PR #540, opened 2019-06-06
STSMACOM-212: Disable search in Notes Assigning Modal by query and filters if previous one is in progress

Yurii-Danylenko:


lib/Notes/components/NotesAccordion/components/NotesAssigningModal/NotesAssigningModal.js
```diff
@@ -359,6 +368,15 @@ export default class NotesAssigningModal extends React.Component {
     });
   }
 
+  getNotesStatusOptions() {
+    return notesStatusOptions.map((option) => {
+      return {
```
MaksymDryha: You can get rid of this `return` as you are using an arrow function

PR #548, opened 2019-06-11
UIOR-298 support results title for SearchAndSort component

NikitaSedyx:
Purpose: https://issues.folio.org/browse/UIOR-298
in one module SearchAndSort is used twice, so we need a way to specify its title (we don't use SearchAndSortQuery approach for now) 

Approach: new property to use provided title instead of module display name

doc is updated


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -170,6 +170,7 @@ class SearchAndSort extends React.Component {
       connect: PropTypes.func,
       hasPerm: PropTypes.func.isRequired
     }).isRequired,
+    title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
aliaksei-chumakou: Could it be just `PropTypes.node` ?

NikitaSedyx: `Pane` accepts `PropTypes.oneOfType([PropTypes.string, PropTypes.element])`

PR #551, opened 2019-06-13
UIORG-81: Prevent changes of the vertical position of elements when a validation error message is displayed and prevent input field validation on cancel.

SergiyVSergiyenko:
### Purpose
Prevent changes of the vertical position of elements when a validation error message is displayed and prevent input field validation on cancel.[Story UIORG-81](https://issues.folio.org/browse/UIORG-81).

**Before**

Before

**After**

After

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-smart-components
 
+## 2.7.3 (IN PROGRESS)
+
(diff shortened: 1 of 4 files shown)


VictorSoroka1:
@rasmuswoelk, thank you for review. We do this since that class is used for the view version as well. So in our case, a top alignment will be only true for the edit version.

PR #558, opened 2019-06-24
STSMACOM-219 & STSMACOM-220: Bold note title

sduvvuri-ebsco:
Per https://issues.folio.org/browse/STSMACOM-219 and https://issues.folio.org/browse/STSMACOM-220, we are supposed to bold note title and assignment status.


sduvvuri-ebsco:
Thanks for your feedback @doytch - updated.

PR #559, opened 2019-06-24
STSMACOM-221: Update color of text in links

sduvvuri-ebsco:
Per https://issues.folio.org/browse/STSMACOM-221, we want to fix the style so the list items do not appear as links.

### DIFF
--- a/lib/Notes/components/AssignmentsList/AssignmentsList.css
+++ b/lib/Notes/components/AssignmentsList/AssignmentsList.css
@@ -1,5 +1,0 @@
-@import "@folio/stripes-components/lib/variables.css";
-
-.assignments-list__item {
-  color: var(--color-text);
-}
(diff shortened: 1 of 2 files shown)


lib/Notes/components/AssignmentsList/AssignmentsList.css
```diff
@@ -1,5 +1,5 @@
 @import "@folio/stripes-components/lib/variables.css";
 
 .assignments-list__item {
-  color: var(--primary);
+  color: var(--color-text);
```
doytch: Can we just remove this file and thus the corresponding import instead? I have a feeling the default styling would be fine then and less likely to cause confusion down the road.

sduvvuri-ebsco: Thanks for your feedback, updated.

PR #572, opened 2019-07-10
UIU-1070: Fix requests count and alert message after loan due date changed

YevheniiMaltsev:
## Purpose

Fix the next issues in scope of the [jira ticket](https://issues.folio.org/browse/UIU-1070):
- Change column name to Requests
- Fix requests amount display
- Change the alert message to "Item has been recalled" for the items that has recall requests

Some tests are added in the scope of [this pull request](https://github.com/folio-org/ui-users/pull/894)

## Screenshots

item_recall_alert
requests_amount_and_column_name


lib/ChangeDueDateDialog/ChangeDueDate.js
```diff
@@ -86,6 +86,7 @@ class ChangeDueDate extends React.Component {
     onCancel: PropTypes.func,
     onDueDateChanged: PropTypes.func,
     onDueDateChangeFailed: PropTypes.func,
+    requestCounts: PropTypes.object,
```
VictorSoroka1: Should not this one be required?

lib/ChangeDueDateDialog/ChangeDueDateDialog.js
```diff
@@ -99,6 +112,14 @@ class ChangeDueDateDialog extends React.Component {
     return null;
   }
 
+  componentDidMount() {
+    const loans = this.loans();
```
VictorSoroka1: Please move this logic to a method and call it there, so componentDidMount would be used for other stuff also.

lib/ChangeDueDateDialog/ChangeDueDateDialog.js
```diff
@@ -138,12 +163,53 @@ class ChangeDueDateDialog extends React.Component {
...
+        this.setState({ openRequests });
+      });
+  };
+
+  getOpenRequestCounts() {
+    const { openRequests } = this.state;
+    const requestCounts = openRequests.reduce((map, record) => {
+      map[record.itemId] = map[record.itemId] ? map[record.itemId] + 1 : 1;
+
+      return map;
```
VictorSoroka1: Let's call it `requests`.

lib/ChangeDueDateDialog/ChangeDueDateDialog.js
YevheniiMaltsev: @VictorSoroka1  thanks for you comments. They were handled and changes can be checked for now.

VictorSoroka1: Thanks @YevheniiMaltsev.

doytch:
@JohnC-80 asked me to comment a bit about some refactoring work I've recently done for situations like this.

A few of us had a bit of a discussion earlier this week about how best to handle these kinds of situations where you need to fetch data that's dependent on some data you fetch beforehand. @skomorokh calls them "client-side joins" which I like!

I've always done it the way you do here, with `GET` mutators triggered off of something, be it `componentDidMount` or a check in `componentDidUpdate` to see if the data (loans in this case) has changed and warrants a refetch of the dependent resource (`requests` in this case).

That said, I've never been super thrilled with my methodology because it starts breaking down when you continue adding resources. I recently refactored ui-agreements to switch to a functional method of building `path` and `params`, which lets stripes-connect figure out all the lifecycle/difference checking for me.

Here's what my [AgreementViewRoute](https://github.com/folio-org/ui-agreements/pull/233/files#diff-94d0ba3d926b932d0cfc7716ff44b689) looks like now. Note how I've shifted out the logic contained in the various `fetch___` functions called in `componentDidMount` and `componentDidUpdate` into new `params` functions for those resources in the `manifest` itself.

I see the following benefits:
- It's more obvious what a resource is dependent on, simplifying maintenance.
- It minimises overuse of component lifecycle methods and entry points into the code that actually _calls_ your `GET()`s.
- Provides a cleaner upgrade path to GraphQL.

VictorSoroka1:
Hi @doytch, I thought the idea that we can succeed from in that PR is to get rid of `resourceShouldRefresh` which is based on your recent changes to the stripes-connect.
https://github.com/folio-org/stripes-connect/pull/96

doytch:
Definitely @VictorSoroka1, and good catch. I noticed the blanket `true` when looking at the code but then forgot to mention it when I was writing up that comment 😆

PR #573, opened 2019-07-10
STSMACOM-218: Notes modal UI updates

Godlevskyi:
## Purpose

This PR approaches the following UI updates in notes assignment modal:
- Fixed borders
- Added caution message (Notes with one assignment cannot be unassigned)
- Added the # of assignments column on the results list
- Place Cancel and Save buttons in the order as outlined on the mockup

## Screenshot
Screen Shot 2019-07-10 at 18 29 28


lib/Notes/components/NotesAccordion/components/NotesAssigningModal/NotesAssigningModal.js
```diff
@@ -481,8 +486,15 @@ export default class NotesAssigningModal extends React.Component {
             )}
             defaultWidth="70%"
           >
+            {notes.totalCount > 0 && (
+              <p className={styles['caution-message']}>
+                <span>
```
Yurii-Danylenko: please get rid of `span`

PR #575, opened 2019-07-12
UIU-1070: Fix alert display for recalled items in change due date modal

YevheniiMaltsev:
## Purpose

Fix "Item has been recalled" message display in the change due date modal and data fetching on modal state handling.

It was agreed to move the data fetching process from the manifest object to the static method to provide data refreshing between modal state transitions. 

The previous version #572


VictorSoroka1:
@JohnC-80, the other approach is fine, but the issue in the scope of the issue required a bit more. The idea of the issue it to retrieve requests based on the loan ids. Given that we can end up with the same loan ids still, require us to refetch requests as those could have changed.

PR #590, opened 2019-08-05
STSMACOM-233: UX Changes to Expand/Collapse Search and Filter pane

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/STSMACOM-233

## Changes
- Updated show/hide search pane toggle
- Removed old toggle component

## Note
One of the requirements for the ticket is a tooltip that is rendered when a user hovers the toggle button. I have left this out for now because we do not currently have a component for this purpose but it's in the works. 

Instead of creating something custom for this feature, I would rather wait until we have a universal component for this purpose and then implement that instead.

## GIF
Aug-05-2019 14-18-45


index.js
```diff
@@ -26,7 +26,6 @@ export { default as SearchAndSort } from './lib/SearchAndSort';
 export { default as SearchAndSortQuery } from './lib/SearchAndSort/SearchAndSortQuery';
 export { default as SearchAndSortNoResultsMessage } from './lib/SearchAndSort/components/NoResultsMessage';
 export { default as SearchAndSortResetButton } from './lib/SearchAndSort/components/ResetButton';
-export { default as SearchAndSortSearchButton } from './lib/SearchAndSort/components/SearchButton';
```
doytch: Is this on purpose? This is a breaking change for modules that use `SearchAndSortQuery` like the ERM modules. [Here's our usage.](https://github.com/folio-org/ui-agreements/blob/master/src/components/views/Agreements.js#L23)

doytch: Other users are at least:
- ui-licenses
- ui-local-kb-admin
- ui-plugin-find-[user/license/agreement]

rasmuswoelk: Hey @doytch 

This work is still WIP and I was going to check up on that part.  But thanks for letting me know! I'll leave it in there so nothing breaks.

I assume that the changes for the UX in this PR will have to be reflected across all apps eventually.

It's easily achieved with the `<PaneHeaderIconButton>` from stripes-components. Note that the collapse button is in the filter pane now and the expand button will be located in the results pane.

doytch: Phew! 😅 

And gotcha...yeah, figured that we'd eventually have to update the other apps to reflect these changes. I think this is one of those places where having separated everything due to `SearchAndSortQuery` can hurt us at the moment and we'd benefit from a few more `SearchAndSortUIWidget` components to smooth over transitions.

rasmuswoelk: @doytch I re-added the search pane toggle component 👍

doytch:
Is there any way to export the two new buttons for apps using `SearchAndSortQuery`?

rasmuswoelk:
@doytch It's pretty much just the `<PaneHeaderIconButton>` from stripes-components that are being used here. But there is of course some translated aria-labels there too.. 

I can move the code into two components that can be exported 👍

PR #597, opened 2019-08-13
bump react-final-form version

zburke:
Bumping [RFF](https://github.com/final-form/react-final-form) to v6, consistent with what we use in other repositories.


mkuklis:
@zburke I wonder if this is the cause of the errors we see in ui-inventory?

mkuklis:
@zburke I just wanted to say thank you for finding this 🙏!  I tried it locally and it does fix bunch of BigTest tests in ui-inventory. I think the issues we currently see are related to pulling in two different versions of react-final-form.

zburke:
> thank you for finding this ...

Praise due to @ttamj for writing [duplicate-package-checker-webpack-plugin](/darrenscerri/duplicate-package-checker-webpack-plugin) in the stripes-core webpack config!

PR #598, opened 2019-08-15
UIORGS-79 Hide route for New record if no permission

aliaksei-chumakou:
Per https://issues.folio.org/browse/UIORGS-79 we need to hide route under Create permission, I think it's the only way to do it in SearchAndSort.
Any comments are welcome, thanks.

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -55,6 +55,23 @@ import buildUrl from './buildUrl';
 
 import css from './SearchAndSort.css';
 
+const NO_PERMISSION_NODE = (
+  <div
+    style={{
+      position: 'absolute',
+      right: '1rem',
+      bottom: '1rem',
+      width: '34%',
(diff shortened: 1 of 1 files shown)


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -820,25 +824,29 @@ class SearchAndSort extends React.Component {
...
-              parentMutator={parentMutator}
-              onSubmit={record => this.createRecord(record)}
-              onCancel={this.closeNewRecord}
-              {...detailProps}
-            />
-          </Layer>
-        )}
-      </IntlConsumer>
-    );
+    if (!disableRecordCreation && stripes.hasPerm(newRecordPerms)) {
```
VictorSoroka1: Why don't use `IfPermission`?
<IfPermission perm={newRecordPerms}>

https://github.com/folio-org/stripes-smart-components/pull/598/files#diff-13cdc54e1bc729fe97e046aac8dec7f4R709

PR #602, opened 2019-08-20
STSMACOM-241: Add tests for notes-smart-accordion

Yurii-Danylenko:
## Purpose
For now testing for Notes is implemented in each app using it. Adding new functionality causes adding the same tests in every application. In this PR, tests for internal Notes accordion stuff is extracted and placed near NotesSmartAccordion component.

## TODO 
Clean up tests in ui-eholdings, ui-users, ui-agreements apps. Leave only stuff for testing integration of the notes accordion into an app.

### DIFF
--- a/lib/Notes/NotesSmartAccordion/tests/notes-smart-accordion-test.js
+++ b/lib/Notes/NotesSmartAccordion/tests/notes-smart-accordion-test.js
@@ -1,7 +1,6 @@
 import { beforeEach, describe, it } from '@bigtest/mocha';
 import React from 'react';
 import { expect } from 'chai';
-import { faker } from '@bigtest/mirage';
 
 import NotesSmartAccordion from '../NotesSmartAccordion';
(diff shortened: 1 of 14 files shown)


lib/Notes/NotesSmartAccordion/tests/notes-smart-accordion-test.js
```diff
@@ -0,0 +1,256 @@
...
+      });
+    });
+
+    describe('and search box and filters are not empty', () => {
+      beforeEach(async () => {
+        await notesModal.enterSearchQuery('some note');
+        await wait(300);
+        await notesModal.noteTypeFilter.options(1).clickOption();
+        await notesModal.selectUnassignedFilter();
+        await notesModal.resetAll();
```
MaksymDryha: it seems like what happens in `beforeEach` is different from what is stated in `describe`

PR #607, opened 2019-08-23
Delete useless test

SergiyVSergiyenko:


lib/ControlledVocab/tests/ControlledVocab-test.js
```diff
@@ -139,10 +139,6 @@ describe('ControlledVocab', () => {
     await when(() => usersLoaded);
   });
 
-  it('should render all filter data options', () => {
```
doytch: 🤣

doytch: I kinda think the intention here is to confirm that the component rendered properly, though. You could make the argument that any future test will _also_ do that, but the whole incremental-test-cases philosophy could argue this case should exist.

VictorSoroka1: In my opinion, if that was the only test in that file, it would make sense. Since it is not, then there is no sense in such a test. Moreover, I found that no filter stuff is not rendered in those tests.

doytch: Yeah I don't mind one way or another. I did find the existence of the test hilarious at first and it took me a few beats to figure out why it was actually there. Was more following-up my initial guffaw with what I realised in case anyone else could learn something. :)

VictorSoroka1: Thanks 🙂

PR #610, opened 2019-08-28
ERM-390 Manage License term value pick list values (refdata values)

CalamityC:
set throwErrors to false for refdataValues in ControlledVocab

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -34,6 +34,7 @@ class ControlledVocab extends React.Component {
       type: 'okapi',
       fetch: false,
       clientGeneratePk: '!{clientGeneratePk}',
+      throwErrors: false,
       PUT: {
         path: '!{baseUrl}',
       },


zburke:
I understand what this change does (it suppresses the Okapi error/Javascript alert) related to constraint violations ... but I don't understand why that's a good thing, nor how it helps to resolve [ERM-390](https://issues.folio.org/browse/ERM-390). Can somebody enlighten me?

doytch:
In general, we show the `alert()` box to inform a user some error has occurred. We added the `throwErrors` flag to stripes-connect to be able to turn off that `alert`, which is useful in situations where the app is catching errors and already displaying them to the user in a useful place.

ControlledVocab handles displaying errors, so it turns `throwErrors: false` in the normal [`values`](https://github.com/CalamityC/stripes-smart-components/blob/ffb119a6302c6e841db84039333f36be337a0dfe/lib/ControlledVocab/ControlledVocab.js#L20) resource.

However, when the [`refdataValues`](https://github.com/CalamityC/stripes-smart-components/blob/ffb119a6302c6e841db84039333f36be337a0dfe/lib/ControlledVocab/ControlledVocab.js#L33) resource was added (I don't love that it exists but 🤷‍♂ ), it didn't turn off `throwErrors` like it should've. As a result, the errors were being correctly shown in CV, but the `alert` box was still being shown _anyway_.

PR #613, opened 2019-08-30
UIOR-345 common ui-plugin-find-* component based on `SearchAndSortQuery`

aliaksei-chumakou:
To reduce copy-paste code in ui-plugin-find-* repos were created common components, `PluginFindRecord` and `PluginFindRecordModal`. They are based on `ui-plugin-find-user` code and use `SearchAndSortQuery` pattern instead of importing list components directly from repos (ui-users, ui-inventory, etc.). 
Since we have a lot of such plugins (ui-plugin-find-contact, ui-plugin-find-interface, etc.) we've put that common code inside stripes-acq-components. However, it could be placed in stripes-smart-components repo, to be available for any module.
https://github.com/folio-org/stripes-acq-components/tree/master/lib/PluginFindRecord
https://issues.folio.org/browse/UIOR-345

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,4 +62,4 @@ export { default as NoteViewPage } from './lib/Notes/NoteViewPage';
 export { default as NoteEditPage } from './lib/Notes/NoteEditPage';
 export { default as NotesSmartAccordion } from './lib/Notes/NotesSmartAccordion';
 
-export * from './lib/PluginFindRecord';
+export { PluginFindRecord, PluginFindRecordModal } from './lib/PluginFindRecord';
--- a/index.js
+++ b/index.js
@@ -61,3 +61,5 @@ export { default as NoteCreatePage } from './lib/Notes/NoteCreatePage';
 export { default as NoteViewPage } from './lib/Notes/NoteViewPage';
 export { default as NoteEditPage } from './lib/Notes/NoteEditPage';
 export { default as NotesSmartAccordion } from './lib/Notes/NotesSmartAccordion';
+
+export * from './lib/PluginFindRecord';
--- a/lib/PluginFindRecord/PluginFindRecordModal.js
+++ b/lib/PluginFindRecord/PluginFindRecordModal.js
@@ -278,6 +278,7 @@ class PluginFindRecordModal extends React.Component {
                               marginBottom0
                               name="query"
                               onChange={getSearchHandlers().query}
+                              onClear={getSearchHandlers().reset}
                               value={searchValue.query}
                             />
                             <Button
--- a/lib/PluginFindRecord/PluginFindRecordModal.js
+++ b/lib/PluginFindRecord/PluginFindRecordModal.js
@@ -121,6 +121,14 @@ class PluginFindRecordModal extends React.Component {
     });
   }
 
+  onRowClick = (e, row) => {
+    const { isMultiSelect, onSelectRow } = this.props;
+
+    if (!isMultiSelect) {
+      onSelectRow(e, row);
+    }
+  }
+
   render() {
     const {
       closeModal,
@@ -134,7 +142,6 @@ class PluginFindRecordModal extends React.Component {
       modalLabel,
       onComponentWillUnmount,
       onNeedMoreData,
-      onSelectRow,
       queryGetter,
       querySetter,
       renderFilters,
@@ -344,9 +351,10 @@ class PluginFindRecordModal extends React.Component {
                         formatter={mixedResultsFormatter}
                         id="list-plugin-find-records"
                         isEmptyMessage={resultsStatusMessage}
+                        key={`checkedRecordsLength_${checkedRecordsLength}`}
                         onHeaderClick={onSort}
                         onNeedMoreData={onNeedMoreData}
-                        onRowClick={!isMultiSelect && onSelectRow}
+                        onRowClick={this.onRowClick}
                         sortDirection={sortOrder.startsWith('-') ? 'descending' : 'ascending'}
                         sortOrder={sortOrder.replace(/^-/, '').replace(/,.*/, '')}
                         totalCount={count}
--- a/lib/PluginFindRecord/PluginFindRecordModal.js
+++ b/lib/PluginFindRecord/PluginFindRecordModal.js
@@ -121,12 +121,6 @@ class PluginFindRecordModal extends React.Component {
     });
   }
 
-  saveMultiple = () => {
-    const selectedRecords = Object.values(pickBy(this.state.checkedMap));
-
-    this.props.onSaveMultiple(selectedRecords);
-  };
-
   render() {
     const {
       closeModal,
--- a/lib/PluginFindRecord/PluginFindRecordModal.js
+++ b/lib/PluginFindRecord/PluginFindRecordModal.js
@@ -67,10 +67,10 @@ class PluginFindRecordModal extends React.Component {
             <FormattedMessage id={hideOrShowMessageId}>
               {hideOrShowMessage => (
                 <FilterPaneToggle
-                  visible={filterPaneIsVisible}
                   aria-label={`${hideOrShowMessage} \n\n${appliedFiltersMessage}`}
-                  onClick={this.toggleFilterPane}
                   badge={!filterPaneIsVisible && filterCount ? filterCount : undefined}
+                  onClick={this.toggleFilterPane}
+                  visible={filterPaneIsVisible}
                 />
               )}
             </FormattedMessage>
@@ -157,9 +157,9 @@ class PluginFindRecordModal extends React.Component {
         <div data-test-find-records-no-results-message>
           <NoResultsMessage
             data-test-find-records-no-results-message
-            source={source}
-            searchTerm={query.query || ''}
             filterPaneIsVisible
+            searchTerm={query.query || ''}
+            source={source}
             toggleFilterPane={noop}
           />
         </div>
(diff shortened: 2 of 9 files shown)


lib/PluginFindRecord/PluginFindRecordModal.js
```diff
@@ -0,0 +1,396 @@
...
+      const isAllChecked = records.every(record => Boolean(checkedMap[record.id]));
+
+      return {
+        checkedMap,
+        isAllChecked,
+      };
+    });
+  }
+
+  saveMultiple = () => {
```
VictorSoroka1: This method is declared two times. Remove one of them.

doytch:
If we're gonna do this - and I'm apprehensive about it given what we just dragged ourselves out of wrt SearchAndSort - I'd really hope we implement the [`renderTrigger` (or similar) approach](https://github.com/folio-org/ui-plugin-find-agreement/blob/master/src/AgreementSearch.js#L53) for [this `Button`](https://github.com/folio-org/stripes-smart-components/pull/613/files#diff-d7488605f2a43854b0dd1fa4bd6854adR51) rather than hardcoding the render.

Not only does it make things more flexible for the consuming application (eHoldings uses that functionality of find-agreement), but we're 100% gonna need something like it for [supporting autofocus a11y in repeating fields](https://folio-project.slack.com/archives/C96DDV0Q7/p1566305149000600).

aliaksei-chumakou:
@doytch yes, `renderTrigger` approach looks nice, I think it deserves to be in the common component. But on this very moment we have just a lot of copy-paste around these plugins, so we should start at some point and then improve it as we want, IMHO.

index.js
```diff
@@ -61,3 +61,5 @@ export { default as NoteCreatePage } from './lib/Notes/NoteCreatePage';
 export { default as NoteViewPage } from './lib/Notes/NoteViewPage';
 export { default as NoteEditPage } from './lib/Notes/NoteEditPage';
 export { default as NotesSmartAccordion } from './lib/Notes/NotesSmartAccordion';
+
+export * from './lib/PluginFindRecord';
```
doytch: Absolute nit-picking, but could we export explicitly from here? It's useful, when checking what things are called and what gets exported to just look at the top-level `index.js` and not have to dive into subdirectories etc.

lib/PluginFindRecord/PluginFindRecordModal.js
```diff
@@ -0,0 +1,401 @@
...
+                              id="clickable-reset-all"
+                              onClick={resetAll}
+                            >
+                              <Icon icon="times-circle-solid">
+                                <FormattedMessage id="stripes-smart-components.resetAll" />
+                              </Icon>
+                            </Button>
+                          </div>
+                          {renderFilters
+                            ? renderFilters(activeFilters.state, getFilterHandlers())
```
doytch: It seems there isn't a way to get resources (such as `data` into `renderFilters`). Am I missing something? Eg, would I be able to [use a set of filters like this](https://github.com/folio-org/ui-plugin-find-agreement/blob/master/src/Filters.js) that uses the `data` prop filled out by the [Container](https://github.com/folio-org/ui-plugin-find-agreement/blob/master/src/Container.js)?

lib/PluginFindRecord/PluginFindRecordModal.js
```diff
@@ -0,0 +1,401 @@
...
+                      >
+                        <form onSubmit={onSubmitSearch}>
+                          <div className={css.searchGroupWrap}>
+                            <SearchField
+                              autoFocus
+                              className={css.searchField}
+                              data-test-plugin-search-input
+                              marginBottom0
+                              name="query"
+                              onChange={getSearchHandlers().query}
```
doytch: It'd probably be useful to have an `onClear={getSearchHandlers().reset}` here.

lib/PluginFindRecord/PluginFindRecordModal.js
```diff
@@ -0,0 +1,401 @@
...
+
+                        }}
+                        formatter={mixedResultsFormatter}
+                        onRowClick={!isMultiSelect && onSelectRow}
+                        onNeedMoreData={onNeedMoreData}
+                        onHeaderClick={onSort}
+                        sortOrder={sortOrder.replace(/^-/, '').replace(/,.*/, '')}
+                        sortDirection={sortOrder.startsWith('-') ? 'descending' : 'ascending'}
+                        isEmptyMessage={resultsStatusMessage}
+                        autosize
```
doytch: Can we alpha-sort all these props? Ideally throughout the PR but esp in these long prop lists.

lib/PluginFindRecord/PluginFindRecordModal.js
aliaksei-chumakou: Maybe I didn't catch the question, but you can pass whatever you want to this function in Container, it's a callback from there.

lib/PluginFindRecord/PluginFindRecordModal.js
aliaksei-chumakou: This code is based on ui-plugin-find-user, by @JohnC-80 . I'm not very familiar with it and I don't see why your proposal is useful, sorry. My intention is not to break things, so maybe you could put more details to this

lib/PluginFindRecord/PluginFindRecordModal.js
aliaksei-chumakou: I think such kind of things should be defined in a project style-guide, eslint, prettier , etc. In other cases it's per dev choice, and in Acquisitions we don't have such rule. However, I'll change it here, despite the fact that stripes-smart-components doesn't have this rule defined in eslint config.

lib/PluginFindRecord/PluginFindRecordModal.js
doytch: If you don't provide an `onClear` to `SearchField`, it doesn't show the clear icon. By providing one that clears the search, you support clearing just the search field as opposed to having to reset the entire search and filters.

aliaksei-chumakou:
> do we want to have tests for this?

Yes, but I think this might be a tech-debt work. Right now we short on time, we have a bunch of modules broken on search and filter, that shouldn't go into release lie that. So it's an attempt to fix them all by one shot. After 3.2 release we could add test and rewrite these common components to hooks.
https://issues.folio.org/browse/UIOR-345
https://issues.folio.org/browse/UIOR-349
https://issues.folio.org/browse/UIORGS-87

lib/PluginFindRecord/PluginFindRecordModal.js
doytch: 🤦‍♂ You're right, of course. I'm struggling today.

aliaksei-chumakou:
There are some weird tests failings on jenkings, related to `ControlledVocab` and `Notes smart accordion`, which aren't part of PR...

lib/PluginFindRecord/PluginFindRecordModal.js
```diff
@@ -0,0 +1,402 @@
...
+                      </Pane>
+                    }
+                    <Pane
+                      defaultWidth="fill"
+                      firstMenu={this.renderResultsFirstMenu(activeFilters)}
+                      padContent={false}
+                      paneSub={resultPaneSub}
+                      paneTitle={RESULTS_HEADER}
+                    >
+                      <MultiColumnList
```
JohnC-80: This might fall under 'future work' for this component, but I would take the Search field, and the contents of the  "results" pane, and wrap them both in render functions that can be overridden by props. As @doytch said, we've been down this road with SearchAndSort for prettymuch the same reasons (reduce the boilerplate.) If a find-record plugin needs to differentiate (ie, have a different interface for search fields - see something like the tabbed search UI for eHoldings or for ui-greements) it's only left with the option to fork this and adjust it locally. The use of this is pretty limited to 'find this one record' but it has to be almost as flexible as SearchAndSortQuery itself - with the healthy defaults to get us over through the UI boilerplate this is meant to help.

aliaksei-chumakou: John, mostly I agree with you at the high level. But can we do it after the release?
The main point of this PR - is to fix ui-plugin-find-* that are broken on search and filtering before 3.2 release. I've just leveraged code, that was copy-pasted across plugins, and the code, you've posted into ui-plugin-find-user. If I had more time I would love to provide some solid and better solution, but as we are short on time and we have a couple ui-plugins to fix - I don't think it's a good call to do it before release. I think the main point not to break existing functionality in the release.

zburke: You're both right; we need more flexibility in the long term _and_ we have a bug to squash in the short term. I think your "future work" comment is spot-on, @JohnC-80. 

As long as the current implementation doesn't cut off avenues for future deployment, i.e. we can extend the API later with render props but leave the current API in place, I think we should commit this as-is. Agreed?

JohnC-80: Yep. Let's just ear-mark the idea of override patterns in the component. No need to implement that immediately.

craigmcnally:
@aliaksei-chumakou @zburke @JohnC-80 @doytch What's left here to be done before we can get approval?  It's important that we fix UIOR-345 in the Q3.2 release.

JohnC-80:
After some discussion about this with the stripes-architecture, we feel that it's best to not pressurize this into centralized code as part of the public API for all modules. It steps close to the similar mistakes that we made with `<SearchAndSort>` despite its smaller scope. The best route at this point is for this to exist within `ui-find-instance` to resolve the immediate issues there. Afer that, we can organically bubble it up and generalize it for the central resources. We're on a path to gradually encapsulate parts of the boilerplate here into bigger molecules that could alleviate the copy/paste issue as far as the UI is concerned - but we've learned, it's best that this happen in smaller pieces first than starting with large abstractions.

craigmcnally:
OK, thanks guys

aliaksei-chumakou:
@craigmcnally I think @zburke has provided some input in slack, but I'm still not sure what's required to fix the bug.

aliaksei-chumakou:
https://github.com/folio-org/ui-plugin-find-instance/pull/39

PR #616, opened 2019-09-06
STSMACOM-227: Add handling server errors during item creation.

OleksandrAntonenko1:
# Story
https://issues.folio.org/browse/STSMACOM-227
# Screenshot
Screen Shot 2019-09-06 at 12 39 53 PM


lib/EditableList/processBadResponse.js
```diff
@@ -0,0 +1,69 @@
...
+      id={`stripes-smart-components.error.${errorCode}`}
+      values={values}
+    />
+  );
+};
+
+const handleError = (errorMessages, message, values) => {
+  if (values && values.fieldLabel) {
+    if (errorMessages.fieldErrors[values.fieldLabel]) {
+      errorMessages.fieldErrors[values.fieldLabel].push(message);
```
zburke: This function works just fine as is, but, would it be better to return an updated copy of `errorMessages` rather than modifying it in place? Mutating arguments feels icky.

OleksandrAntonenko1: Done.

PR #617, opened 2019-09-06
Adjust notes modal layout. Remove unneccessary logic in EditableList

JohnC-80:
Additionally attempts [UINOTES-55](https://issues.folio.org/browse/UINOTES-55?jql=text%20~%20%22Notes%20sorting%22) by adding a check for the existence of the parameter arrays before using their `.length` property.

## Before:
image

## After:
image


lib/Notes/NotesSmartAccordion/tests/notes-smart-accordion-test.js
```diff
@@ -99,7 +99,7 @@ describe('Notes smart accordion', () => {
         await notesModal.notes(0).clickCheckbox();
       });
 
-      it('should not allow to unassign', () => {
+      it.only('should not allow to unassign', () => {
```
doytch: Oops!

JohnC-80: Yep. * long-winded-sigh....

JohnC-80: I guess if es-lint doesn't catch it, sonar cloud will at least make us look...

PR #620, opened 2019-09-13
Implement tooltip on expand/collapse pane buttons

rasmuswoelk:
Implemented the newly added `<Tooltip>`-component on the expand/collapse filter pane buttons.

This change was originally a part of [STSMACOM-233](https://issues.folio.org/browse/STSMACOM-233) but we didn't have a tooltip component at that time, so I left out that part of the ticket.

Tested and working with RTL.

# Screenshots
image

image


rasmuswoelk:
@aliaksei-chumakou Thanks for your review. I will add a few more devs for review as well.

The failing tests will be fixed soon 👍

PR #643, opened 2019-10-30
STCOM-801: Add property 'editable' to have opportunity to make EditableListForm not editable

annamelnyk:
### Purpose
 Add possibility to make EditableListForm  not editable

### Approach
In ControlledVocad component added prop 'editable' (true by default). If set 'editable' to false, EditableListForm appears without 'New' button and 'Actions' column

### Issues
This need for implementation https://issues.folio.org/browse/UIIN-801

###Screenshot
image


lib/ControlledVocab/tests/ControlledVocab-test.js
```diff
@@ -6,118 +6,156 @@ import {
...
+import React from 'react';
 import ControlledVocabInteractor from './interactor';
 import mountComponent from './mountComponent';
 
 import { setupApplication } from '../../../tests/helpers';
 
+import ControlledVocab from '../ControlledVocab';
 
-describe('ControlledVocab', () => {
+describe.only('ControlledVocab', () => {
```
maximdidenkoepam: Please remove only

lib/EditableList/EditableListForm.js
```diff
@@ -261,29 +265,44 @@ class EditableListForm extends React.Component {
...
+    }
+
+    return visibleFields;
   }
 
   getReadOnlyColumns() {
     const actionsArray = ['actions'];
-    if (this.props.readOnlyFields) {
+
+    if (this.props.readOnlyFields && this.props.editable) {
```
OleksandrAntonenko1: Please, use destructuring here.

lib/ControlledVocab/tests/ControlledVocab-test.js
```diff
@@ -6,118 +6,156 @@ import {
...
       });
-    });
-  });
 
-  describe('clicking New button', () => {
-    beforeEach(async () => {
-      await cv.newButton().click();
+      describe('when focusing on the input field', () => {
+        beforeEach(async () => {
+          await cv.fillInputField('asdf');
```
OleksandrAntonenko1: Please use proper values for inputs. Of course, it doesn't matter in the end but I don't think that input should be filled with 'asdf'.

lib/EditableList/EditableListForm.js
```diff
@@ -43,6 +43,10 @@ const propTypes = {
    * Label for the 'Add' button
    */
   createButtonLabel: PropTypes.node,
+  /**
+   *  boolean that indicates if the component is editable (has 'New' button and column 'actions').
+   */
+  editable: PropTypes.bool,
```
JohnC-80: This should probably be added as `true` to `defaultProps` to preserve functionality in instances of `<EditableList>` without `<ControlledVocab>`.

lib/EditableList/EditableListForm.js
```diff
@@ -261,29 +266,49 @@ class EditableListForm extends React.Component {
...
+
+    return visibleFields;
   }
 
   getReadOnlyColumns() {
     const actionsArray = ['actions'];
-    if (this.props.readOnlyFields) {
-      return this.props.readOnlyFields.concat(actionsArray);
+    const {
+      getReadOnlyColumns,
```
NikitaSedyx: @annamelnyk   is it new property? can't find it

PR #645, opened 2019-10-31
Fix locallyChangedSearchTerm. Refs UIIN-758

mkuklis:
I noticed this issue when switching between different search segments in ui-inventory and ui-orders. After performing search on the first segment and switching to another segment the search field would get repopulated with the value form the previous segment.

When a query resource is defined (it's configured via manifest) on multiple components (in the same ui package) its values get cached and are accessible between the components.

This PR doesn't populate locallyChangedSearchTerm if the query value from the query resource doesn't match the current query param in the url.

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -239,20 +239,7 @@ class SearchAndSort extends React.Component {
   }
 
   componentDidMount() {
-    const {
-      location: {
-        search,
-      },
-      browseOnly,
-    } = this.props;
-    const resQuery = this.queryParam('query');
-    const { query } = queryString.parse(search);
-
-    // update locallyChangedSearchTerm only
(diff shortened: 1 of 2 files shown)


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -239,10 +239,20 @@ class SearchAndSort extends React.Component {
...
+      browseOnly,
+    } = this.props;
+    const resQuery = this.queryParam('query');
+    const { query } = queryString.parse(search);
+
+    // update locallyChangedSearchTerm only
+    // when query from resourceQuery matches url query
+    const term = (browseOnly && resQuery === query) ? query : '';
+
+    this.setState({ locallyChangedSearchTerm: term });
```
VictorSoroka1: The code looks good @mkuklis. I am just wondering about moving the logic itself to the method and just make a call in the component did mount. What do you think about that?

PR #647, opened 2019-11-01
STCOM-802: fix mistake

annamelnyk:
Fix mistake with property, which I made in PR: STCOM-801


aliaksei-chumakou:
@annamelnyk Thanks for the quick fix!

PR #648, opened 2019-11-04
STSMACOM-259 show unknown-count message as needed

zburke:
When a search returns more than 10k results, Okapi does not return an 
actual result-count but instead returns the value `999,999,999`. Here, 
we use that value as a  trigger to report an "unknown result count" message.

Fixes [STSMACOM-259](https://issues.folio.org/browse/STSMACOM-259)


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -987,9 +987,19 @@ class SearchAndSort extends React.Component {
...
 
-    const paneSub = !source.loaded()
-      ? <FormattedMessage id="stripes-smart-components.searchCriteria" />
-      : <FormattedMessage id={messageKey} values={{ count }} />;
+    // when there are > 10k results to a search, the `totalRecords`
+    // value comes back as `999999999`, so we use that value as a
+    // trigger to display an "unknown total count" value.
+    let paneSub;
+    if (source.loaded) {
+      if (count === 999999999) {
```
aliaksei-chumakou: Maybe it makes sense to have a constant for this magic number

zburke: OMG yes, my CS-101 prof. would have failed an assignment if I turned in code like that! 🤦‍♂

doytch:
Is there a way we could have this logic somewhere that `SearchAndSortQuery` benefits from it as well? Eg, `StripesConnectedSource`?

zburke:
I like the idea of making this part of the resource API but that causes `StripesConnectedResource` to need to know about `FormattedMessage`, which I maybe would rather not because it's nicely focused around its resource at present. An alternative would be to return the translation key to use, but that still feels a little kludgey (`translationKeyForResultCount`, maybe)? Dunno 😕 

What do you think?

doytch:
Not necessarily; what if StripesConnectedSource looks something like:
```
  totalCount() {
    let res;
    if (!this.recordsObj.hasLoaded) {
      res = null;
    } else {
      res = other.totalRecords !== 999999999 ? other.totalRecords : undefined;
    }
    this.logger.log('source', 'totalCount:', res);
    return res;
  }
```
And then SearchAndSort looks like:
```
    const count = source.totalCount(); // count === undefined
    ....
    let paneSub;
     if (source.loaded) {
       if (count === undefined) {
         paneSub = <FormattedMessage id="stripes-smart-components.searchResultsCountUnknown" />;
       } else {
         paneSub = <FormattedMessage id={messageKey} values={{ count }} />;
       }
     } else {
       paneSub = <FormattedMessage id="stripes-smart-components.searchCriteria" />;
     }
```

doytch:
I generally don't like suggesting solutions that ascribe meaning to falsey values like `undefined`, but I think it actually fits here. The total count _is_ undefined.

lib/SearchAndSort/ConnectedSource/StripesConnectedSource.js
```diff
@@ -23,8 +25,19 @@ export default class StripesConnectedSource {
...
+   * that the count is, in fact, undefined vs returning null to indicate
+   * that the count has not been calculated.
+   */
   totalCount() {
-    const res = this.recordsObj.hasLoaded ? this.recordsObj.other.totalRecords : null;
+    let res;
+    if (!this.recordsObj.hasLoaded) {
+      res = null;
+    } else {
+      res = this.recordsObj.other.totalRecords !== 999999999 ? this.recordsObj.other.totalRecords : undefined;
```
aliaksei-chumakou: maybe to use `UNKNOWN_RECORDS_COUNT` since it's declared above

zburke: Srsly, what is my problem with this constant?!?

NikitaSedyx:
@zburke just curious, why do we need to show `> 10k` only when `total records > 999999999`? we will show number for `999999995`, but `>10k` for `999999999+`, a bit strange

mb this message should be displayed for 10k+?

NikitaSedyx:
@zburke @aliaksei-chumakou explained me what's wrong with BE, previous question doesn't make sense now :)
+1 to use constant as it's already declared :)

lib/SearchAndSort/ConnectedSource/StripesConnectedSource.js
aliaksei-chumakou: I think we all have the same problems from time to time, that's why we need a code review process :rescue_worker_helmet:

PR #654, opened 2019-11-08
SearchAndSort props for view/create links STSMACOM-251

skomorokh:


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -16,6 +17,7 @@ import {
   noop,
   isEmpty,
   defer,
+  omit,
```
MikeTaylor: Note that `omit` is, idiotically, slated for removal from Lodash. I suppose we'll just wait till that happens, then bulk-port all our uses of it to whatever `lodash-omit-replacement` library emerges at that point.

(To be clear, this is not really a comment on your perfectly cromulent code, just a whinge about the state of the world.)

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -725,10 +743,11 @@ class SearchAndSort extends React.Component {
               <Button
                 id={`clickable-new${objectName}`}
                 aria-label={ariaLabel}
-                href={this.craftLayerUrl('create')}
                 buttonStyle="primary"
                 marginBottom0
-                onClick={this.addNewRecord}
+                href={createRecordPath ? undefined : this.craftLayerUrl('create')}
```
MikeTaylor: Why do we have `href` as well as `to`?

If we're going to have it at all, I think you might helpfully reorder the props here so that `href` is next to `to`.

lib/SearchAndSort/SearchAndSort.js
MikeTaylor: Sorry it took so long to get to this, by the way!

lib/SearchAndSort/SearchAndSort.js
skomorokh: We have `href` because that's what it was and I wanted to avoid changing existing behaviour. We have `to` because that's needed for intra-application routes. I did helpfully group the ones that are conditional based on createRecordPath so it's certainly in the vicinity but I agree that it would be better to swap the order of those two. If I happen to wind up in that area again I might but otherwise I think it's not worth it.

MikeTaylor: Agreed: best not change `href` if it was already there, on the basis of Chesterton's Fence.

I say go ahead and merge.

zburke:
Rats; I was optimistic the `yarn.lock` file would put us on the road to passing tests. >:(

PR #660, opened 2019-11-14
UIU-1316: add 422 error handling into onSave method

annamelnyk:
### Description
When 422 error received, error message is not displaying
### Approach
In onSave method add 'handleError' method, where is checking, which error is received from server,
and showing it's message
### Issue
https://issues.folio.org/browse/UIU-1316
https://issues.folio.org/browse/UICIRC-380


lib/EntryManager/tests/entry-manager-test.js
```diff
@@ -15,131 +16,223 @@ import {
...
+    <ConnectedComponent
+      nameKey="test"
+      paneTitle={paneTitle}
+      entryList={entryList}
+      entryLabel={entryLabel}
+      enableDetailsActionMenu
+      parentMutator={
+        {
+          entries: {
+            POST: () => new Promise((resolve, reject) => {
```
VictorSoroka1:
```suggestion
            POST: () => Promise.reject(response)
```

lib/EntryManager/tests/entry-manager-test.js
```diff
@@ -15,131 +16,223 @@ import {
...
+        <ConnectedComponent
+          nameKey="test"
+          paneTitle={paneTitle}
+          entryList={entryList}
+          entryLabel={entryLabel}
+          enableDetailsActionMenu
+          parentMutator={
+            {
+              entries: {
+                POST: () => new Promise((resolve) => {
```
VictorSoroka1:
```suggestion
            POST: () => Promise.resolve()
```

lib/EntryManager/tests/entry-manager-test.js
```diff
@@ -15,131 +16,223 @@ import {
...
+  });
+
+  describe('Test 400 error', () => {
+    const resp400 = new Response();
+    resp400.status = 400;
+    resp400.statusText = '400 Bad Request';
+
+    setupApplication({ scenarios: ['controlled-vocab-default-error'] });
+    beforeEach(async () => mountErr(resp400));
+
```
VictorSoroka1: Remove empty line.

lib/EntryManager/tests/entry-manager-test.js
```diff
@@ -15,131 +16,223 @@ import {
...
+  });
 
-          describe('item deletion prohibit modal', () => {
-            beforeEach(async () => {
-              await entryManagerInteractor.detailsSection.actionsButton.click();
-              await entryManagerInteractor.detailsSection.actionsDropdown.whenDeleteLoaded();
-              await entryManagerInteractor.detailsSection.actionsDropdown.delete.click();
-            });
+  describe('Test 422 error', () => {
+    const resp422 = new Response();
```
VictorSoroka1: I am wondering if it is possible to do like this to simplify things

```
     new Response(422, { 'Content-Type': 'application/json' }, JSON.stringify({
        errors: [{
          message: 'Cannot create entity; name is not unique',
          code: 'name.duplicate'
        }]
      }));
```

lib/EntryManager/tests/entry-manager-test.js
```diff
@@ -75,6 +74,11 @@ describe('entry manager', () => {
           it('callout should be shown', () => {
             expect(entryManagerInteractor.calloutElement.isPresent).to.be.true;
           });
+
+          it('callout should display error message', () => {
+            // eslint-disable-next-line max-len
```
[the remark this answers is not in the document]
annamelnyk: Unfortunately, this does not work this way, I just removed lint comment

lib/EntryManager/tests/entry-manager-test.js
annamelnyk: Yes, you are right, because of json method, this notation is not available

lib/EntryManager/EntryWrapper.js
```diff
@@ -107,9 +107,24 @@ export default class EntryWrapper extends React.Component {
...
+
+    if (error.status === 422) {
+      const response = await error.json();
+      errorMessage = response.errors[0].message;
+    } else {
+      const { message, statusText } = error;
+      errorMessage = message || statusText;
+    }
+
+    this.showSubmitErrorCallout(errorMessage);
```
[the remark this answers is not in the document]
annamelnyk: This topic was already discussed with the team and Zak Burke. There no single pattern for implementation of translation of server-side error messages, and better to avoid one-off implementations. Also translation is out of scope this issue.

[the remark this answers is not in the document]
VictorSoroka1: Even if it is not a part of the issue, I think this good point to raise and discuss. Not necessarily to handle it in the scope of this issue though. Since these changes takes place as part of the shared component this why I raise it.
cc @zburke

lib/EntryManager/tests/entry-manager-test.js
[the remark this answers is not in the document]
VictorSoroka1: The following has worked for me.
```
        it('callout should display error message', () => {
          const { errorMessage } = entryManagerInteractor.calloutElement;

          expect(errorMessage).to.equal('Cannot create entity; name is not unique');
        });

```

lib/EntryManager/EntryWrapper.js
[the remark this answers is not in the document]
zburke: Agreed. I am glad you brought it up, and maybe if we keep beating the "It is wrong to forward localizable values from the server straight to the UI" drum, eventually we'll get some traction from a PO or a non-English institution. 

Here's the note I sent to @annamelnyk when she asked about this: 
> I would just forward the server side message directly to the UI. We do need to handle translation of server side error messages, but since we don’t have pattern for that implementation so we can be sure all modules function in the same way, I would like to avoid a one-off implementation. Does that make sense?
> 
> I wish we had a solution in place to manage translation of server-side error messages, but all the early-implementors have put a low priority on translation-related tasks. We have the same problem with permissions: their labels and descriptions are always displayed in English in ui-users. We have tickets for these problems ([UXPROD-370](https://issues.folio.org/browse/UXPROD-370), [UXPROD-371](https://issues.folio.org/browse/UXPROD-371)) but until a customer makes them a priority, the work will never get assigned to us developers even though it feels like an obvious bug  :shrug:

lib/EntryManager/tests/entry-manager-test.js
```diff
@@ -15,131 +16,217 @@ import {
...
+    });
+  });
+
+  describe('Test 400 error', () => {
+    const resp400 = new Response();
+    resp400.status = 400;
+    resp400.statusText = '400 Bad Request';
+
+    setupApplication({ scenarios: ['controlled-vocab-default-error'] });
+    beforeEach(async () => mountErr(resp400));
```
VictorSoroka1: Can `mountErr` be moved to [here](https://github.com/folio-org/stripes-smart-components/pull/660/files#diff-3bb23c7ac7ff4457983b8a73992d2850R215)? I just do not see the reason for two beforeEach blocks when it is possible to deal with one.

The same applies to [here](https://github.com/folio-org/stripes-smart-components/pull/660/files#diff-3bb23c7ac7ff4457983b8a73992d2850R181)?

lib/EntryManager/tests/entry-manager-test.js
```diff
@@ -15,131 +16,217 @@ import {
...
+          await entryManagerInteractor.entryForm.entryFormSave();
+        });
+
+        it('callout should be shown', () => {
+          expect(entryManagerInteractor.calloutElement.isPresent).to.be.true;
+        });
+
+        it('callout should display error message', () => {
+          const { errorMessage } = entryManagerInteractor.calloutElement;
+          expect(errorMessage).to.equal('400 Bad Request');
```
VictorSoroka1: Please add an empty line after variable declarations. Check similar cases.

PR #664, opened 2019-11-19
initialize before testing

zburke:
The record-count tests fail sometimes, for some people, in some
environments, on occassion. Luckily, I'm already mostly bald, so I
haven't had to worry about tearing my hair out about this, because
otherwise I would totally be tearing my hair out about this.

My speculation is that the `beforeEach` blocks are really hooks that
Mocha pulls out and executes first, ya know _before_ it executes
anything else, which is leading to errors like
```
Error: Uncaught TypeError: Cannot read property 'manifest' of null (tests/index.js:355930)
```
which don't clearly relate to anything. @johnc-80 figured out that is
coming from `stripes-connect` and @mkuklis pointed out that it must be
calling `connect` on `null`, which I finally sussed out to be this, I
hope. I _really really_ hope.

Refs [FOLIO-2360](https://issues.folio.org/browse/FOLIO-2360)


zburke:
> This branch is out-of-date with the base branch

Are you sure, GitHub, because nothing has merged to this repo since yesterday. Why is this appearing now?

zburke:
@VictorSoroka1 found a solution that _actually works_ in #665.

PR #677, opened 2019-12-02
Remove explicit fake-xml-http-request in advance for the stripes-core

VictorSoroka1:
Having multiple versions of the `fake-xml-http-request` can prevent patching.


doytch:
Do you need to remove it in `resolutions` as well?

VictorSoroka1:
Yes, unless there is alternative fix.

PR #682, opened 2019-12-09
UIU-1173: Update en.json in translations

iduranm:
I modified the variable cv.cannotDeleteTermMessage so that it will adapt to the required message of the UIU-1173 issue:

JIRA:
Current message: This Fee/fine owner cannot be deleted, as it is in use by one or more records.
New message: This Fee/fine owner cannot be deleted, as it is associated with one or more manual fee/fine charges.

The other possibility is to create a personalized message through the UI-USERS module so as not to modify this variable cannotDeleteTermMessage

Thank you.

### DIFF
--- a/translations/stripes-smart-components/en.json
+++ b/translations/stripes-smart-components/en.json
@@ -35,7 +35,7 @@
   "feesfines.policy.lost.duplicate": "The Lost item fee policy name entered already exists. Please enter a different name.",
 
   "cv.cannotDeleteTermHeader": "Cannot delete {type}",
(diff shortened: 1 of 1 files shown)


zburke:
As @doytch noted, this value is used in a generic `<ControlledVocab>` component so we cannot change it to use specific messaging. An alternative would be to provide an additional prop to `<ControlledVocab>` that names a local string, e.g. `ui-users.feefine.cannotDeleteTermMessage`. Then `<CV>` could do something like 
```
defaultProps = {
  cannotDeleteTermMessage = "stripes-smart-components.cv.cannotDeleteTermMessage",
}
```
to accept the custom value via props, or use the existing default.

PR #684, opened 2019-12-12
Support improved result list paging and sorting

ryandberger:
- Supports [STCON-57](https://issues.folio.org/browse/STCON-57) by adding integration point for `resultOffset`. 
- Fulfills [STSMACOM-269](https://issues.folio.org/browse/STSMACOM-269) by resetting `resultCount` and `resultOffset` if applicable.
- Supports upcoming MCL "load More" button (PR [here](https://github.com/folio-org/stripes-components/pull/1117)) in `stripes-components` by passing `pagingType`.


aliaksei-chumakou:
reviewed, :+1:

lib/SearchAndSort/ConnectedSource/StripesConnectedSource.js
```diff
@@ -74,6 +75,11 @@ export default class StripesConnectedSource {
     // ... and stripes-connect notices this change and does the rest by magic
   }
 
+  // fetch by offset which is passed in as index. _askAmount is ignored.
```
zburke: If this is a brand new function, why are we writing it with an argument we ignore?

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -364,12 +367,13 @@ class SearchAndSort extends React.Component {
...
   onFilterChangeHandler = (filter) => {
     const {
-      parentMutator: { resultCount },
+      parentMutator: { resultCount, resultOffset },
       initialResultCount,
       onFilterChange,
     } = this.props;
 
     resultCount.replace(initialResultCount);
+    if (resultOffset) resultOffset.replace(0);
```
zburke: This isn't wrong, but personally, I strongly prefer that all conditionals be written with `{}` across multiple lines. It's so easy to indent something under this line and think it's part of the condition when it is not. Defensive programming and all that. 

TBH, given all the stooooooooooooooopid AirBNB eslint rules we put up with, I'm astonished this practice isn't among them.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -533,6 +549,9 @@ class SearchAndSort extends React.Component {
     const sortOrder = orders.slice(0, maxSortKeys).join(',');
 
     this.log('action', `sorted by ${sortOrder}`);
+    // Reset result count when sorting so that only 1 page is initially retrieved
```
zburke: Are we confident this is the right behavior? For me, the Principle of Least Surprise would be that if I had clicked "Show more" twice and so had 300 records, clicking "Sort" would sort those 300 records. I understand I might get 300 _different_ records if I hadn't "Show more"-ed all the way through the result set, but I just think of sort rearranging the order of things, not changing my limit and offset parameters. 

Maybe this is an unlikely use-case and folks are most likely to sort before clicking "Show more". Dunno; I'm a dev, not a UX guy, so my instincts here are not necessarily useful.

lib/SearchAndSort/ConnectedSource/StripesConnectedSource.js
ryandberger: Good point. I can remove it.

lib/SearchAndSort/SearchAndSort.js
ryandberger: Yes, I confirmed with Vince that the result count should reset. The sorting just affects the parameters that are passed into the API calls. So, if we don't reset the count, we potentially need to fire off `n` number of API calls at once depending on how many times the user clicked "load more". Resetting the count is more performant and considered more predictable according to Vince.

lib/SearchAndSort/SearchAndSort.js
ryandberger: Fair enough. I'll fix them.

lib/SearchAndSort/SearchAndSort.js
VictorSoroka1: So sorting will be always limited by the one page size, right?

ryandberger: Yes, but this only affects modules that use `searchAndSort`.

ryandberger:
> We need to revisit the unit test changes here and in #665. There should be tests both for 99,999 records (which should show "99,999 records found") _and_ for 999,999,999 records (which should show "More than 10k records found") but it looks like that is no longer the case. This isn't meant to derail this PR; it's more of a note to self that I think we're no longer testing that threshold correctly.

I added the other test case back in so both are covered now.

PR #694, opened 2020-01-09
Corrected resultOffset to be defined as optional in propTypes

ryandberger:
This prevents warnings from displaying in the console for modules that have not opted in to the new "load more" workflow for result lists.


ryandberger:
> Great that `resultOffset` can be optional!
> 
> But when it _is_ supplied, shouldn't the `replace` function be mandatory? stripes-connect should provide that function in all mutators.

Good point, I'll fix that!

PR #698, opened 2020-01-28
STSMACOM-296 bump ControlledVocab limit to 2000

zburke:
Increase the `<ControlledVocab>` fetch limit from 500 to 2000 because
people want to have lots of things.

Refs [STSMACOM-296](https://issues.folio.org/browse/STSMACOM-296), [UIIN-819](https://issues.folio.org/browse/UIIN-819)


MikeTaylor:
This really does seem like an awful lot. I don't know about you, but when we first introduced `<ControlledVocab>`, I imagined it being used on maybe ten to twenty values. It's hard to imagine an approach intended for small sets to work well on 1500. 

Of course, that doesn't mean we shouldn't let the customer have this small fix that they need in the short term. But I do think we're going to need to think about introducing something a bit more SearchAndSort-like for long controlled vocabs like this.

zburke:
Duly noted, @MikeTaylor.

PR #707, opened 2020-02-10
UIU-1295: Changing due date of multiple loans sequentially

VictorSoroka1:
POC for renewing loans sequentially according to the discussion in [CIRC-545](https://issues.folio.org/browse/CIRC-545). Note, that `batch` (now called `silent`) option is new in the system and now is part of another [PR](https://github.com/folio-org/stripes-connect/pull/126).


VictorSoroka1:
Sounds good @mkuklis, it definitely would be helpful to have some like this.

PR #710, opened 2020-02-14
Add prop to Editablelist for disabling '+ Add new' button

Vladyslav-Velytskyi:


lib/EditableList/EditableListForm.js
```diff
@@ -476,7 +482,7 @@ class EditableListForm extends React.Component {
               <Row end="xs">
                 <Col xs>
                   <Button
-                    disabled={isEditing}
+                    disabled={isEditing || canCreate}
```
zburke: I would expect the "Add new" button to be _disabled_ when `canCreate` is false. Am I thinking of this backwards? 

To maintain existing functionality, I would expect this to be a new prop with a default value of `true` that you would have to pass in as `false` in order to suppress the button.

JohnC-80:
Thank you for this @Vladyslav-Velytskyi - Despite being able to accomplish this through `actionProps`, it is a more straightforward way of specifically disabling the "Add" button. The two props will be able to coexist nicely, I think.

lib/EditableList/EditableListForm.js
Vladyslav-Velytskyi: Thanks, I'm agree with you, you are right, I'm already fixed it.

PR #716, opened 2020-02-21
Modify ChangeDueDate to handle 'declared lost' items

Baroquem:
See https://issues.folio.org/browse/UIU-1207

### DIFF
--- a/lib/ChangeDueDateDialog/ChangeDueDate.js
+++ b/lib/ChangeDueDateDialog/ChangeDueDate.js
@@ -176,9 +176,9 @@ class ChangeDueDate extends React.Component {
 
         if (error.url && error.url.indexOf(putEndpoint)) {
           const failedLoanId = error.url.split(putEndpoint)[1];
-          const errorMessage = this.itemIsDeclaredLost(failedLoanId) ?
-            <FormattedMessage id="stripes-smart-components.cddd.itemDeclaredLost" /> :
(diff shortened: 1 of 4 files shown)


lib/ChangeDueDateDialog/ChangeDueDate.js
```diff
@@ -173,9 +175,12 @@ class ChangeDueDate extends React.Component {
         const putEndpoint = '/circulation/loans/';
 
         if (error.url && error.url.indexOf(putEndpoint)) {
-          const failedLoan = error.url.split(putEndpoint)[1];
+          const failedLoanId = error.url.split(putEndpoint)[1];
+          const errorMessage = this.itemIsDeclaredLost(failedLoanId) ?
+            <FormattedMessage id="stripes-smart-components.cddd.itemDeclaredLost" /> :
+            <FormattedMessage id="stripes-smart-components.cddd.changeFailed" />;
```
annamelnyk:
```
 const errorMessage = this.itemIsDeclaredLost(failedLoanId) 
    ? <FormattedMessage id="stripes-smart-components.cddd.itemDeclaredLost" />
    : <FormattedMessage id="stripes-smart-components.cddd.changeFailed" />;
```
Please, use more readable way of transposition in a multi-line ternary

Baroquem: Thanks, I will do it that way.

PR #725, opened 2020-02-28
STSMACOM-275: Ability to change Custom Fields accordion label

Vladyslav-Velytskyi:
## Purpose
Add API connection and functionality to change Custom fields accordion label for different modules 

## Link
https://issues.folio.org/browse/STSMACOM-275
https://issues.folio.org/browse/STSMACOM-276


lib/CustomFields/utils/useSectionTitleFetch.js
```diff
@@ -0,0 +1,44 @@
...
+        setSectionTitle({ ...configs[0] });
+      } else {
+        setSectionTitleHasError(true);
+      }
+    };
+
+    if (!sectionTitleLoading) {
+      fetchCustomFields();
+    }
+  });
```
MaksymDryha: please, add all of the necessary dependencies to the effect. now it runs after every component update

lib/CustomFields/utils/useSectionTitleFetch.js
```diff
@@ -0,0 +1,44 @@
...
+      if (response.ok) {
+        const { configs } = await response.json();
+
+        setSectionTitle({ ...configs[0] });
+      } else {
+        setSectionTitleHasError(true);
+      }
+    };
+
+    if (!sectionTitleLoading) {
```
MaksymDryha: you won't have to make this check if you declare all the necessary dependencies of the effect

lib/CustomFields/components/CustomFieldsForm/CustomFieldsForm.js
```diff
@@ -297,22 +314,32 @@ const CustomFieldsForm = ({
...
-          }
-        />
-      )}
-    </FormattedMessage>
+    <TextField
+      autoFocus
+      value={title}
+      inputClass={styles.accordionNameInput}
+      className={styles.inputWrapper}
+      onChange={e => onChangeTitle(e.target.value)}
```
MaksymDryha: I would write  `onChange={onChangeTitle}`

lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
```diff
@@ -72,23 +84,42 @@ const EditCustomFieldsSettings = ({
...
+  };
+
   const fetchUsageStatistics = useCallback(fieldId => {
     return makeOkapiRequest(`custom-fields/${fieldId}/stats`)({
       method: 'GET',
     });
   }, [makeOkapiRequest]);
 
-  const saveCustomFields = async data => {
+  const saveCustomFields = async (data, accTitle) => {
```
MaksymDryha: please, name the second argument "accordionTitle"

lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
```diff
@@ -36,18 +42,23 @@ const EditCustomFieldsSettings = ({
...
     customFields,
     customFieldsLoaded,
-    hasError,
+    customFieldsFetchFailed,
   } = useCustomFieldsFetch(okapi, backendModuleId, entityType);
-
+  const {
+    sectionTitle,
+    sectionTitleFetchFailed,
+  } = useSectionTitleFetch(okapi, backendModuleName.toUpperCase());
```
MaksymDryha: We also need to check if section title is loaded here: https://github.com/folio-org/stripes-smart-components/pull/725/files#diff-47bbc820a330f2283c6d6ead7954b1e0R127
Currently, you display the default section title until the response is 
The same applies to `ViewCustomFieldsSettings` component

lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
```diff
@@ -72,23 +84,42 @@ const EditCustomFieldsSettings = ({
...
-      setSubmitFinished(true);
-    }
+
+    Promise.all([updateCustomFields(data), updateTitle(accTitle)])
+      .then(redirectToView)
+      .catch(() => {
+        showErrorNotification();
+        setSubmitting(false);
+        setSubmitFinished(true);
+      });
```
MaksymDryha: Fetch rejects promises only when the request fails because of network-related issues.
If the response's status code is something like 404, `catch`'s callback won't be executed and the error won't be handled. 
You need to check if `response.ok` is `true` to make sure that the response was fulfilled successfully.

PR #730, opened 2020-03-04
UIU-1492 Custom fields: Create UI permissions for Settings > Users > Custom Fields CRUD

BogdanDenis:
## Purpose

- Create UI permissions for Custom Fields


lib/CustomFields/pages/ViewCustomFieldsSettings/ViewCustomFieldsSettings.js
```diff
@@ -3,6 +3,10 @@ import { PropTypes } from 'prop-types';
 import { connect } from 'react-redux';
 import { FormattedMessage } from 'react-intl';
 
+import {
+  stripesConnect,
```
Godlevskyi: Please remove the stripesConnect as it is not needed

zburke:
This PR added a _required_ prop, `permissions` to `ViewCustomFieldSettings`. Technically, this is a breaking change and should have been accompanied by a major version update to this repository, and hence to `@folio/stripes` as well. Alternatively, and perhaps preferably, you could have chosen to preserve the existing behavior by adding a default props object. As things currently stand, code based on the previous implementation now throws an NPE. 

For example, this is preventing any PR from merging in ui-users because the unit tests related to (the old implementation) now fail.

PR #735, opened 2020-03-05
STSMACOM-314-fix: fix bugs

Vladyslav-Velytskyi:


zburke:
@Vladyslav-Velytskyi, thank you for fixing this! In future PRs, please provide some indication of what happened in the PR title and/or body. Even if it's just clean-up work for a previous PR, that's helpful. Right, the only way to figure out what happened is read through the commit history.

PR #740, opened 2020-03-16
STSMACOM-299: move custom fields to final form

MaksymDryha:
In this PR, I made the CustomFieldsForm component utilize stripes-final-form


lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
```diff
@@ -143,17 +199,26 @@ const EditCustomFieldsSettings = ({
...
-                initialValues={getInitialValues()}
-                redirectToView={redirectToView}
-                onSubmit={saveCustomFields}
-                submitting={submitting}
-                submitSucceed={submitFinished}
-                fetchUsageStatistics={fetchUsageStatistics}
-                sectionTitle={sectionTitle.value || defaultSectionTitle}
-                permissions={permissions}
-              />
+              <>
```
BogdanDenis: I don't think that we need extra `<>` here because there's only one component returned

MaksymDryha: @BogdanDenis, good catch, thank you!

package.json
```diff
@@ -56,18 +56,21 @@
     "@folio/stripes-components": "~6.1.0",
     "@folio/stripes-connect": "~5.5.0",
     "@folio/stripes-core": "~4.1.0",
+    "@folio/stripes-final-form": "^2.1.100041",
```
Godlevskyi: Such version exist only on the folioci. 
Please change it to "@folio/stripes-final-form": "^2.1.1",

MaksymDryha: Thanks! I checked that the latest released version is 2.1.0

PR #749, opened 2020-03-20
UIU-1527: Settings > Users > Custom fields : Checkbox settings labels should display above 

BogdanDenis:
## Purpose

- Change how Custom Fields checkboxes are displayed to be consistent with other applications

## Issue
[UIU-1527](https://issues.folio.org/browse/UIU-1527)

## Screenshots

**HiddenField**, **RequiredField**
image


lib/CustomFields/components/FieldAccordion/edit-fields/shared-fields/HiddenField.js
```diff
@@ -17,10 +17,11 @@ const HiddenField = ({ fieldNamePrefix }) => (
   <Col xs={3}>
     <Field
       type="checkbox"
-      component={Checkbox}
       name={`${fieldNamePrefix}.hidden`}
       label={<FormattedMessage id="stripes-smart-components.customFields.hidden" />}
-    />
+    >
+      {props => <Checkbox {...props} vertical />}
+    </Field>
```
Vladyslav-Velytskyi: Why you tossing child component in this way? It is correctly works with `vertical` prop in Field component
```
<Field
  type="checkbox"
  component={Checkbox}
  name={`${fieldNamePrefix}.hidden`}
  label={<FormattedMessage id="stripes-smart-components.customFields.hidden" />}
  vertical
/>
```
Some modules could has warning on rest props, this way can omit warnings in future

PR #755, opened 2020-03-31
UIREQ-415: Add prop hasNewButton to SearchAndSort component

annamelnyk:
# Approach
Add `hasNewButton` prop to `SearchAndSort` component. 

- By default `New` button of the pane appears and  `hasNewButton` sets to `true`.
- If  `hasNewButton` sets to `false` - `New` button of the pane does not appear.

# Link
https://issues.folio.org/browse/UIREQ-415

# Screenshots
**hasNewButton = true (by default)**

image

**hasNewButton = false**

image

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -105,6 +105,7 @@ class SearchAndSort extends React.Component {
     finishedResourceName: PropTypes.string,
     getHelperComponent: PropTypes.func,
     getHelperResourcePath: PropTypes.func,
+    hasNewButton: PropTypes.bool,
     history: PropTypes.shape({ // provided by withRouter
       push: PropTypes.func.isRequired,
     }).isRequired,
@@ -218,7 +219,8 @@ class SearchAndSort extends React.Component {
     getHelperComponent: noop,
     massageNewRecord: noop,
     renderNavigation: noop,
-    selectedIndex: ''
+    selectedIndex: '',
+    hasNewButton: true,
   };
 
   constructor(props) {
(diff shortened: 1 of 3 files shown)


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -1122,7 +1125,7 @@ class SearchAndSort extends React.Component {
           appIcon={<AppIcon app={moduleName} />}
           paneTitle={title || (module && module.displayName)}
           paneSub={paneSub}
-          lastMenu={this.renderNewRecordBtn()}
+          lastMenu={hasNewButton ? this.renderNewRecordBtn() : null}
```
VictorSoroka1: As I see there is already the [prop](https://github.com/folio-org/stripes-smart-components/blob/02eb030c260491b1e365e64165349e5ac99f05b8/lib/SearchAndSort/SearchAndSort.js#L789) for this. Please check whether it does what you need.

annamelnyk: Victor, `disableRecordCreation` prop does not allow to create new record at all, whereas I need to have opportunity to create new record with the `New` button, which is located inside `Actions` dropdown, but not display `New` button by it's own, which is hardcoded so I have creted extra prop.

annamelnyk: It prevents creation [here](https://github.com/folio-org/stripes-smart-components/blob/02eb030c260491b1e365e64165349e5ac99f05b8/lib/SearchAndSort/SearchAndSort.js#L903)

zburke: IOW, this is basically a short-term fix to allow "un-migrated" apps to continue displaying the `New` button, and to allow "migrated" apps (those that have added the "new" action to their action menu) to hide it. Correct?

annamelnyk: Yes, Zak, this is correct. For hiding `New` button, `hasNewButton` prop should be setted to `false`

PR #759, opened 2020-04-07
Helper-app rendering no-ops if the helper is undefined

MikeTaylor:
The previous version contained what looks like a straight-up bug where
the rendering no-opped only if the helper _name_ was undefined. This
meant you could invoke a URL with, say, `helper=noSuchThing`, and
`<SearchAndSort>` would fail to find that helper but still try to
render the component that implements it, even though it was undefined.

Thanks to @EthanFreestone for helping me spot this.

--

NB. I nominated @VictorSoroka1 to review this because the original test is in code he added, and I want to be sure I've not misunderstood his intent.

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -750,7 +750,7 @@ class SearchAndSort extends React.Component {
     const helper = this.queryParam('helper');
     const HelperAppComponent = this.getHelperComponent(helper);
 
-    if (!helper) {
+    if (!HelperAppComponent) {
       return null;
     }
 


VictorSoroka1:
@mike, my commit was regarding code clean up. I found that @mkuklis added the possibility to render the helper application.

MikeTaylor:
Thanks, @VictorSoroka1. `git blame` does seem to indicate that this conditional was yours, though: https://github.com/folio-org/stripes-smart-components/blame/8e46d6985e38a3652e9484df289cf3868916f71e/lib/SearchAndSort/SearchAndSort.js#L753-L755

VictorSoroka1:
Previously it was in the way `helper && (` @MikeTaylor.

MikeTaylor:
Oh, good spot! Ok, you are innocent :-)

MikeTaylor:
Yes, this logic was originally the work of @mkuklis in https://github.com/folio-org/stripes-smart-components/commit/29119d02#diff-13cdc54e1bc729fe97e046aac8dec7f4R471-R480

Michal, are you happy with this change, or have I overlooked a case that you had in mind?

PR #760, opened 2020-04-07
Remove deprecated intlShape, towards react-intl upgrade STRIPES-656

skomorokh:


skomorokh:
Because it was removed upstream. react-intl went to TypeScript and no longer exports intlShape as they don't maintain proptypes.

https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md#breaking-api-changes

As mentioned in the JIRA referenced in the commit message, apologies, I really should get in the habit of adding links to those in PRs.

MikeTaylor:
Dear me, wouldn't it be lovely if just _one_ of the React development groups would stop making backwards-incompatible changes every five minutes? What a flatly idiotic thing for them to have done. I suppose we have no real choice here but to make our code less type-safe in accordance with their wishes.

MikeTaylor:
Anyway, I can't find the "accept" button for the review, but consider this an endorsement of your PR.

PR #763, opened 2020-04-13
UIDATIMP-461: Support custom getEntity and getEntityTags methods in Tags component :: DONE

taras-tkachenko:
## Root cause:
- `stripes-smart-components` `TagsForm` component does not support custom wrapped entities.

## Approach:
- Change `stripes-smart-components` `TagsForm` component functionality to accept custom `entity wrappers` and custom `entity paths`;
- Implement  those custom wrappers for all of `data-import` `ViewProfile` sections.

## Example:
```
{tagsEnabled && (
  <div data-test-tags-accordion>
    <TagsAccordion
      link={tagsEntityLink}
      entityWrapper={entity => ({
        id: entity.id,
        profile: entity,
        addedRelations: [],
        deletedRelations: [],
      })}
    />
  </div>
)}
```


lib/Tags/Tags.js
```diff
@@ -103,7 +109,13 @@ class Tags extends React.Component {
   }
 
   getEntity() {
-    return get(this.props, ['resources', 'entities', 'records', 0], {});
+    const entity = get(this.props, this.props.entityPath, {});
+    console.log('Entity Wrapper: ', this.props.entityWrapper, this.props);
```
VictorSoroka1, line 113: Not sure that log is needed.

lib/Tags/Tags.js
```diff
@@ -103,7 +109,13 @@ class Tags extends React.Component {
...
+    if (this.props.entityWrapper === noop) {
+      return entity;
+    }
+    console.log('Entity after: ', entity);
```
VictorSoroka1, line 117: Not sure that log is needed.

taras-tkachenko: Removed

lib/Tags/Tags.js
taras-tkachenko: Removed

lib/Tags/Tags.js
```diff
@@ -28,6 +28,8 @@ class Tags extends React.Component {
 
   static propTypes = {
     children: PropTypes.func,
+    entityPath: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
```
doytch, line 31: Just wondering - I don't have any stake in this - but might it be a bit safer to just have a `getEntity` prop instead, with a default value of `getEntity: props => props.resources?.entities?.records?.[0],`.

It wouldn't be a big complexity change for users of this override, but it:
- allows just a bit more functionality than a hardcoded path
- is a more common API (function) than an array of path tokens
- lets us minimize our usage of lodash/get

Hell, I'm not sure exactly how you're using `entityWrapper`, but it might let you get away with not needing that prop at all....

taras-tkachenko: It makes sense.
Will think about this.

VictorSoroka1: For the sake of completeness the default prop than should be 
```
getEntity: props => props.resources?.entities?.records?.[0] || {},
```

taras-tkachenko:
Hi @zburke , @doytch 
I agree it is the best solution but how can it be done in *class* component?
`this` is not enabled in static props like `defaultProps`, so I cannot define the following:
```
getEntity: props => props.resources?.entities?.records?.[0] || {},
```

zburke:
```
static defaultProps = {
  getEntity: props => props.resources?.entities?.records?.[0] || {},
};
```
Then you can scrap `this.getEntity` entirely and replace all the calls to it with 
```
this.props.getEntity(this.props)
```

taras-tkachenko:
@zburke @doytch @VictorSoroka1 
I have done the changes you'll like to :)
Could you please review them now?

CHANGELOG.md
```diff
@@ -2,9 +2,10 @@
 
 ## [3.2.0](IN PROGRESS)
 
-* Custom fields: create page accordions for create/edit/view record. UIU-1279
+* Custom fields: create page accordions for create/edit/view record. UIU-1279.
 * Custom fields: apply checkbox vertical alignment. Refs UIU-1527.
 * Add optional prop `hasNewButton` to `SearchAndSort` component. Refs UIREQ-415.
+* Support custom **getEntity** and **getEntityTags** methods in **Tags** component (UIDATIMP-461)
```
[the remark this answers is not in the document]
taras-tkachenko: OK

lib/Tags/Tags.js
```diff
@@ -28,28 +28,30 @@ class Tags extends React.Component {
...
     refreshRemote: PropTypes.func.isRequired,
     resources: PropTypes.shape({
-      entities: PropTypes.shape({
-        records: PropTypes.arrayOf(PropTypes.object),
-      }),
-      tags: PropTypes.shape({
-        records: PropTypes.arrayOf(PropTypes.object),
-      }),
+      entities: PropTypes.shape({ records: PropTypes.arrayOf(PropTypes.object) }),
+      tags: PropTypes.shape({ records: PropTypes.arrayOf(PropTypes.object) }),
```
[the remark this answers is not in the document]
taras-tkachenko: I had this done because of eslint errors.
However, I can return it back to its state if needed.

PR #764, opened 2020-04-13
STSMACOM-329: Use filter delimiter instead of comma

mkuklis:
https://issues.folio.org/browse/STSMACOM-329

Use filter delimiter (currently set to `\t`) from stripes-components (https://github.com/folio-org/stripes-components/pull/1238) instead of comma in order to correctly split filters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Custom fields: create page accordions for create/edit/view record. UIU-1279
 * Custom fields: apply checkbox vertical alignment. Refs UIU-1527.
 * Add optional prop `hasNewButton` to `SearchAndSort` component. Refs UIREQ-415.
+* Use filter delimiter instead of comma for filter splitting. Fixes STSMACOM-329.
 
 ## [3.1.0](https://github.com/folio-org/stripes-smart-components/tree/v3.1.0) (2020-03-16)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v3.0.0...v3.1.0)
--- a/lib/SearchAndSort/SearchAndSortQuery.js
+++ b/lib/SearchAndSort/SearchAndSortQuery.js
@@ -9,11 +9,12 @@ import cloneDeep from 'lodash/cloneDeep';
 import queryString from 'query-string';
 import { withRouter } from 'react-router';
 
+import { FILTER_DELIMITER } from '@folio/stripes-components';
+
 import {
   mapNsKeys,
   getNsKey,
 } from './nsQueryFunctions';
-
 import buildUrl from './buildUrl';
 
 const locationQuerySetter = ({ location, history, nsValues }) => {
@@ -34,8 +35,8 @@ const buildFilterString = (activeFilters) => {
     .map((filterName) => {
       return activeFilters[filterName].map((filterValue) => {
         return `${filterName}.${filterValue}`;
-      }).join(',');
-    }).join(',');
+      }).join(FILTER_DELIMITER);
+    }).join(FILTER_DELIMITER);
 
   return newFiltersString;
 };
@@ -47,7 +48,7 @@ const buildFilterParams = (activeFilters) => {
 
 const filterStringToObject = (str) => {
   const filterObject = {};
-  const filterArray = str.split(',');
+  const filterArray = str.split(FILTER_DELIMITER);
   filterArray.forEach(f => {
     const [filterName, ...rest] = f.split('.');
     const filterValue = rest.join('.'); // support both filterName.value and filterName.value.with.dot
@@ -597,7 +598,7 @@ class SearchAndSortQuery extends React.Component {
     }
 
     return filtersString
-      .split(',')
+      .split(FILTER_DELIMITER)
       .reduce((resultFilters, currentFilter) => {
         const [filterName, filterValue] = currentFilter.split('.');
 


zburke:
Can/should you add updates to `/lib/SearchAndSort/parseFilters.js` in this PR? Agreed, @MikeTaylor?

MikeTaylor:
Without knowing the background to this, it look like a very bad idea. _Lots_ of software knows that entries in the filter string are comma-separated: they include `<FilterGroups>` and its constituent parts, `makeQueryFunction`, `getSASparams` and `generateQueryParams`, and possibly many more. How are we going to keep all these things working?

mkuklis:
@MikeTaylor thanks for your feedback.

We need to address the issue described in: 

https://issues.folio.org/browse/STSMACOM-329 (please see the comments from @zburke there).

I started poking around a bit more and you are right there are many areas which may be affected by this change. 

https://github.com/search?q=org%3Afolio-org+generateQueryParams&type=Code

I'm not familiar with `getSASparams` and `generateQueryParams`.

Do you know if `generateQueryParams` actually uses anything from `SearchAndSort` or `FilterGroups` from stripes-components? They seem to be independent from this change.

I think `makeQueryFunction` should be also fine.

I'm of course open for other ideas.

zburke:
With due respect, I think the Very Bad Idea was for us to allow lots of software to implicitly agree on using the comma rather than a constant, but I see your point. Can we handle this more conservatively with a different approach?

1. `export const FILTER_DELIMITER = ',';`
1. search high and low for joins and splits that are hard coded to `,` and change them to `FILTER_DELIMITER`
1. In a workspace, change `FILTER_DELIMITER` to some other character, run all our unit and integration tests and see what breaks
1. ???
1. Profit!
1. There is no rule six.
1. Stop mixing memes.

MikeTaylor:
`makeQueryFunction` was the original query-generation function that I made for ui-users and ui-instances (back before it was ui-inventory!); `getSASparams` was created in the ERM work to do a similar job for generating a query targeted the ERM team's non-RMB backend modules which use a different query language from CQL; and `generateQueryParams` is (I think) @doytch's simplified and consequently more flexible variant of that.

To make things worse, as you have discovered, there are several copies of at least some of these functions, and trying to change them all would be both tedious and error prone (not to mention the difficulty of synchronizing the changes that would need to happen across multiple modules).

I don't disagree with @zburke that "the Very Bad Idea was for us to allow lots of software to implicitly agree on using the comma rather than a constant". But I don't think that contradicts my sense that it would be _another_ Very Bad Idea to try to change it now.

My conclusion is that this ought not to be attempted. That leaves us with two options:
1. We can just say that commas in tag values are not supported. An app that wants to use a tag value like "Access Only, Do Not Loan" needs to be reworked to use IDs instead of plain text.
2. We can use an encoding approach for embedded commas that works backwards-compatibly for filter values that do _not_ have commas (i.e. 99% of them).

We can think a bit more about how #2 would work, but I am imagining something like mapping a literal `,` in a value to something distinctive like `,,` (which I think can never legitimately occur) in a filter string; then mapping it back from `,,` to literal `,` when unpacking the filter string.

Since `,` is the _only_ special character in the context, that ought to suffice.

zburke:
TBH, I'm not crazy about the `,,` approach given the potential for sequential commas to look like empty list items which then get omitted and thus transformed back to single commas, and then we're back to square one. 

Searching for `','` in the platform-core repos, there are ~60 instances across ~40 files. Admittedly, this is a naive approach, but it's not completely brain-dead. And while that's a lot of changes to make, it's not an overwhelming count. 

What if we investigate those instances, (and probably search for `'.'` too), and convert them to `FILTER_SEPARATOR` and `UNIT_SEPARATOR` where necessary and then issue `stripes` `v4.0.0` where we change `FILTER_SEPARATOR` to a non-comma character such as `\t`?

doytch:
Yeah, Mike's summarisation of `getSASParams` and `generateQueryParams` is correct.  Here's [where I depend on a comma](https://github.com/folio-org/stripes-erm-components/blob/9f43564c3c81f8c56dac66d22a63319e86b036fa/lib/generateQueryParams.js#L21), for example. This does remind me of what I was doing that actually got me onto the current implementation of gQP. 

I was implementing the filters we recently added to Licenses that let you basically construct complex queries about license terms. I wanted to be able to use commas in my filter values to represent logical `AND`. At first, I thought that to do this I'd need to go and pass in overrides to `SearchAndSortQuery`'s `filtersToString`, `filtersToParams` and `filterParamsMapping` props that would let me opt out of the comma-separator. (Though I literally wrote all the code, I figured out a simpler way that didn't require it in the end. 🤦)

If this was an issue for a module or two, I'd probably advocate that approach. It does seem like we might run into this over and over again, though. And I definitely see the world where we just run into a whack-a-mole scenario where we switch it to another character, but then want to use _that_ character in a filter down the line somewhere.

So I kind of see a few separate things here:
- What should the default separator be? There has to be one.
- Given that the world is cruel and unjust, how do we ameliorate the future situation when someone needs to use that `FILTER_DELIMITER` in their filter? Can we pass in a `filterDelimiter` prop? Anyone doing things that custom should probably already have opted out of `FilterGroups` so it won't hassle them on that front so we probably don't even need to worry about it there. Or should we then just make them implement filtersToString/filtersToParams/filterParamsMapping and avoid cluttering up the API. @JohnC-80 would likely have an opinion :) 
- Do we make these changes now and damn anyone relying on the comma? Should there first be a notification to warn people who _are_ depending on it? I _believe_ I'm the only writer of custom url-params-to-fetch-params code and this doesn't seem like a difficult thing for me to support.

mkuklis:
Thanks @doytch. So it does sound like ERM modules won't be affected by this change at all. Is my understanding correct? 

@zburke and I were thinking about choosing `\t` for a default separator. It's not perfect but it's very unlikely that it will be used in the ways the comma is currently used in: 

https://issues.folio.org/browse/STSMACOM-329. 

I'm happy to choose a different separator if you all have better candidate.

My understanding here is that this change together with the change in: 

https://github.com/folio-org/stripes-components/pull/1238

would _just work_ and we wouldn't have to notify anybody. The only difference people would see it the change in the urls.

I really like what @zburke proposed here:

https://github.com/folio-org/stripes-smart-components/pull/764#issuecomment-613948750

doytch:
> Thanks @doytch. So it does sound like ERM modules won't be affected by this change at all. Is my understanding correct?

No. I'd have to import and use the new delimiter in [`generateQueryParams`](https://github.com/folio-org/stripes-erm-components/blob/9f43564c3c81f8c56dac66d22a63319e86b036fa/lib/generateQueryParams.js#L21) and [`getSASParams`](https://github.com/folio-org/stripes-erm-components/blob/bc31ac13e7756d4878b7b4784ef3b70121de1c4b/lib/getSASParams.js#L27). Because I'd have to do that, I should be requiring a given version of Stripes for that delimiter export in peer dependencies. This is either a minor or major version update for Stripes (depending on your perspective on whether the comma was part of the API), but either way, it'd require me to push a new major version of stripes-erm-components.

mkuklis:
Thank you again for explaining this @doytch and @MikeTaylor.

I think perhaps the best way to approach it would be to move forward with:

https://github.com/folio-org/stripes-components/pull/1238

but without replacing the comma and with just exposing `FILTER_DELIMITER`. Then create a new minor release of stripes-components.

Then give people some time to adopt the new `FILTER_DELIMITER`. 
After that come up with the new separator.

Please let me know if that makes sense to all of you.

MikeTaylor:
Well, @zburke, I am not hung up on `,,` as the representation of a literal comma within a filter value. But I do think that some variant of this approach is required.

Here's why: there are ultimately only to ways for us to allow commas within values: either (A) provide a special way to encode them; or (B) change to a different separator.

And to me B is unappealing because it's incredibly hard to know when we've captured _all_ the places where a comma is used in the code — especially as individual apps can and do have their own code that works on filter-strings, and we can't even know what apps exist. And even we could find them all, sychronizing the switchover of the core modules and all the apps would be nightmarish.

(Also, whatever we decide to use instead of `,` as the separator will surely crop up sooner or later as required in a filter value, so we'll end up having to implement a quoting mechanism anyway.)

If we do end up going with some variant of option B, then @mkuklis's incremental approach is definitely the way to do it with relatively minimal disruption. But it seems all wrong to me. Isn't it much simpler just to somehow quote the separator when it occurs in a value?

MikeTaylor:
Since @mkuklis has now fixed UIU-1596 (what used to be STSMACOM-329) with an eight-character fix (using the patron-group ID instead of its human-readable name in the filter string, see https://github.com/folio-org/ui-users/pull/1258/ ) I think the best thing to do with the present PR is to quietly close it and pretend it never happened.

What we _might_ usefully do is add a note to the filtering documentation advocating the use of IDs instead of labels.

mkuklis:
@MikeTaylor I think I agree with you for the time being. I do like the idea of having one common delimiter everybody agrees on (which can be exported from one place). My primary motivation with this was to fix https://github.com/folio-org/ui-users/pull/1258 which should now work so I'm going to close this.

Thank you everyone for spending your time on this with me. It was a very valuable lesson.

MikeTaylor:
Well, we _do_ have one common delimiter everybody agrees on!  :-) / 2

PR #781, opened 2020-04-23
UIU-1565: Create/Edit/View User record | Custom Field | Single select dropdown

BogdanDenis:
## Purpose
- Add support for Single select dropdown Custom Field type in User record

## Issue

[UIU-1565](https://issues.folio.org/browse/UIU-1565)

## TODOS
Coverage for Custom Fields will be done in scope of [STSMACOM-331](https://issues.folio.org/browse/STSMACOM-331)

## Screenshots
EditCustomFieldsRecord
image

image

### DIFF
--- a/lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
+++ b/lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
@@ -126,7 +126,7 @@ const EditCustomFieldsRecord = ({
     } else {
       return createCustomFieldRenderFunction(customField)();
     }
-  }
+  };
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
```diff
@@ -79,6 +81,17 @@ const EditCustomFieldsRecord = ({
   const formatedCustomFields = chunk(visibleCustomFields, columnCount);
   const columnWidth = rowShapes / columnCount;
 
+  const collectDataOptionsForSelect = (customField) => {
+    if (customField.type !== fieldTypes.SINGLE_SELECT_DROPDOWN) {
```
MaksymDryha, line 85: I'd better move this check to line 107:
```js
dataOptions={customField.type !== fieldTypes.SINGLE_SELECT_DROPDOWN
  ? collectDataOptionsForSelect(customField)
  : null
}
```

mkuklis: I agree. We also don't need to introduce `dataOptions` for other types. If it would be better to split up `renderCustomField` into two different methods. One for `SINGLE_SELECT_DROPDOWN` and another one for the reset of them.

PR #783, opened 2020-04-23
Added PersistedPaneset component

doytch:
# PersistedPaneset

Paneset with LocalStorage-backed persistence

## Description

`stripes-components` provides a `Paneset` component that supports drag-and-drop resizing of its child `Pane`s. This component saves those layout sizes to the client browser's `localStorage`.

## Example

```
<PersistedPaneset
  appId="@folio/agreements"
  id="agreements-paneset"
>
  // NB: The Pane's must all have IDs as well.
  //     This is a requirement for Pane's resizing functionality.
  <Pane id="left-pane">Foo</Pane>
  <Pane id="middle-pane">Baz</Pane>
  <Pane id="right-pane">Inga</Pane>
</PersistedPane>
```

### Properties

Name | Type | Required | Description
--- | --- | --- | ----
appId | string | Yes | The app ID that will be used for namespacing `localStorage`. Eg, `@folio/users`
id | string | Yes | The ID of the Paneset itself. This will be used for namespacing `localStorage`.


lib/PersistedPaneset/PersistedPaneset.js
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    <Paneset
+      id={panesetId}
+      initialLayouts={storedPanesetLayout}
+      onResize={({ layoutCache }) => {
+        // We don't want to persist any layout that contains a NaN so filter those out first.
+        const persistedLayoutCache = layoutCache.filter(layout => {
+          const widths = Object.values(layout);
+          const hasNaN = widths.some(width => width.includes('NaN'));
```
doytch, line 29: This is here because I noticed that occasionally - especially during initial mounting and page load - some layouts resolve a Pane to a width of `"NaN%"`. This generally occurs when rendering a Paneset with a single Pane in it, such as when filters are hidden and no record is selected.

lib/PersistedPaneset/PersistedPaneset.js
```diff
@@ -0,0 +1,42 @@
...
+  children: PropTypes.node,
+  id: PropTypes.string.isRequired,
+};
+
+const PersistedPaneset = ({
+  appId,
+  children,
+  id: panesetId,
+}) => {
+  const persistKey = `${appId}/${panesetId}/layout`;
```
skomorokh, line 18: John brought up a good point that Jakub made earlier: we'll want to be sure to namespace localstorage by user also to account for the common use case that multiple users are using the same service point.

lib/PersistedPaneset/PersistedPaneset.js
```diff
@@ -0,0 +1,42 @@
...
+  const [storedPanesetLayout] = useLocalStorage(persistKey);
+
+  return (
+    <Paneset
```
VictorSoroka1, line 22: I that original `Paneset` can accept more props than specified here like `isRoot`, `nested` and whatever props in the future. For the sake of completeness, what do you think about adding the possibility to pass other props?

doytch:
New commit contains:
- An extra `storageId` prop for segmenting via whatever the parent wants. I don't want to subscribe to the entire `stripes` object in this for just the user id. If we add a `useCurrentUser` context hook I could add support for that.
- Passing along `...rest` props to the `Paneset`.

PR #787, opened 2020-04-28
UIREQ-454: Prevent ProxyManager from triggering additional form submissions

Baroquem:
This PR addresses a problem with the ProxyManager component. ProxyManager includes a form component that is submitted when a user chooses a proxy relationship in the component's modal. This is invoked from within another app (primarily ui-requests, I think). The problem arises because the ProxyManager form is nested inside the ui-requests RequestForm form. When the ProxyManager form is submitted, the submit handler bubbles up and triggers a submission of RequestForm as well.

This fixes the problem by preventing the ProxyManager form submission from propagating to the outer form.


lib/ProxyManager/ProxyForm.js
```diff
@@ -82,7 +93,7 @@ class ProxyForm extends React.Component {
     ));
 
     return (
-      <form id="proxy-form" onSubmit={handleSubmit}>
+      <form id="proxy-form" onSubmit={e => this.handleSubmitLocal(e)}>
```
SergiyVSergiyenko, line 96:
```suggestion
      <form id="proxy-form" onSubmit={this.handleSubmitLocal}>
```
It seems to work too.

PR #791, opened 2020-04-29
STSMACOM-292: implement "Radio button set" custom field type

MaksymDryha:
This PR adds the functionality of the creation of "Radio button set" custom field.

## Screenshot
image

## To-do
* Test coverage for the whole custom fields feature will be done in the scope of [STSMACOM-331](https://issues.folio.org/browse/STSMACOM-331)
* Logic of handling radio button and select options will be significantly simplified after the backend provides a unique id for each option


lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
```diff
@@ -160,6 +180,18 @@ const EditCustomFieldsSettings = ({
...
     return customFieldsFormData.customFields.map(({ values: { hidden, ...customFieldData } }) => {
+      const fieldHasOptions =
+        customFieldData.type === fieldTypes.RADIO_BUTTON_SET ||
+        customFieldData.type === fieldTypes.SELECT;
+
+      if (fieldHasOptions) {
+        if (customFieldData.selectField.defaults === NO_DEFAULT_OPTIONS_VALUE) {
+          customFieldData.selectField.defaults = [];
+        } else {
+          customFieldData.selectField.defaults = [customFieldData.selectField.defaults];
```
BogdanDenis, line 191: Is `customFieldData.selectField.defaults` an array? Do we need to do destructuring here?

MaksymDryha: Since we use radion buttons to chose the default option, final-form stores the data as a plain value: `{ defaults: 'option' }`. Backend expects us to send an array of default options:  `{ defaults: ['option'] }`, that's why I'm doing these transformations

PR #802, opened 2020-05-07
UIU-1514: add radio button set view and edit functionality

MaksymDryha:
Requires https://github.com/folio-org/ui-users/pull/1295
## Purpose
To be able to edit and see values of custom fields of type "Radio button set" on entity records

### Populating custom fields with their default values
 Some custom fields can have default values (like select and radio button set). It's not possible to pass these default values to a form, since `<Form initialValues={...} />` doesn't live in `stripes-smart-components`. That's why, in order to populate fields with default values, I imperatively change their values after there are rendered.
### Supporting final-form and redux-form
To imperatively change the value of a field it's necessary to know if its a part of `final-form` or `redux-form`. For this purpose, a couple of props were added:
### redux-form props
* `isReduxForm` - used to tell custom fields if they are being rendered inside `redux-form`. If `isReduxForm` is `true`, then custom fields will dispatch `redux-form`'s `change` action to populate fields with default values
* `formName` - when dispatching the `change` action, we need to know the name of the form whose value will be changed

### final-form props
If `isReduxForm` is `false`, then the following props should be passed:
* `changeFormValue` - a function used to change `final-form` field value. We can simply pass `form.change` function passed with `Form` render props
* `finalFormCustomFieldsValues` - custom fields values stored in final-form state. Can be retrieved using `Form` render props:  {form.getState().values.customFields}. 

## Example of using with final-form
```js
<Form ...>
  {({form: { change, getState } }) => (
    <EditCustomFieldsRecord
      changeFormValue={change}
      finalFormCustomFieldsValues={getState().values.customFields}
      accordionId="customFields"
      onToggle={this.toggleSection}
      expanded
      backendModuleName="mod-kb-ebsco-java"
      entityType="package"
      fieldComponent={Field}
    />
  )}
</Form>

```

## Example of using with redux-form
```js
<EditCustomFieldsRecord
  isReduxForm
  formName="userForm"
  accordionId="customFields"
  onToggle={this.handleSectionToggle}
  expanded={sections.customFields}
  backendModuleName="users"
  entityType="user"
  fieldComponent={Field}
/>
```

## Screenshot
image

### DIFF
--- a/lib/CustomFields/constants.js
+++ b/lib/CustomFields/constants.js
@@ -14,7 +14,7 @@ export const fieldTypes = {
   CHECKBOX: 'SINGLE_CHECKBOX',
   SELECT: 'SINGLE_SELECT_DROPDOWN',
   MULTISELECT: 'MULTI_SELECT_DROPDOWN',
-  RADIO_BUTTON_SET: 'RADIO_BUTTON',
+  RADIO_BUTTON_GROUP: 'RADIO_BUTTON',
 };
 
 export const fieldTypesLabels = {
@@ -23,7 +23,7 @@ export const fieldTypesLabels = {
   [fieldTypes.CHECKBOX]: <FormattedMessage id="stripes-smart-components.customFields.fieldTypes.CHECKBOX" />,
   [fieldTypes.SELECT]: <FormattedMessage id="stripes-smart-components.customFields.fieldTypes.SELECT" />,
   [fieldTypes.MULTISELECT]: <FormattedMessage id="stripes-smart-components.customFields.fieldTypes.MULTISELECT" />,
-  [fieldTypes.RADIO_BUTTON_SET]: <FormattedMessage id="stripes-smart-components.customFields.fieldTypes.RADIO_BUTTON" />,
+  [fieldTypes.RADIO_BUTTON_GROUP]: <FormattedMessage id="stripes-smart-components.customFields.fieldTypes.RADIO_BUTTON" />,
 };
 export const NO_DEFAULT_OPTIONS_VALUE = 'no-default';
 
@@ -66,11 +66,11 @@ export const defaultFieldConfigs = {
       }
     },
   },
-  [fieldTypes.RADIO_BUTTON_SET]: {
+  [fieldTypes.RADIO_BUTTON_GROUP]: {
     name: '',
     visible: true,
     helpText: '',
-    type: fieldTypes.RADIO_BUTTON_SET,
+    type: fieldTypes.RADIO_BUTTON_GROUP,
     selectField: {
       multiSelect: false,
       defaults: NO_DEFAULT_OPTIONS_VALUE,
@@ -100,7 +100,7 @@ export const fieldComponents = {
   [fieldTypes.TEXTAREA]: TextArea,
   [fieldTypes.CHECKBOX]: Checkbox,
   [fieldTypes.SELECT]: Select,
-  [fieldTypes.RADIO_BUTTON_SET]: RadioButtonGroup,
+  [fieldTypes.RADIO_BUTTON_GROUP]: RadioButtonGroup,
 };
 
 export const rowShapes = 12;
--- a/lib/CustomFields/constants.js
+++ b/lib/CustomFields/constants.js
@@ -8,11 +8,6 @@ import {
   RadioButtonGroup,
 } from '@folio/stripes-components';
 
-export const formTypes = {
-  FINAL_FORM: 'finalForm',
-  REDUX_FORM: 'reduxForm'
-};
-
 export const fieldTypes = {
   TEXTFIELD: 'TEXTBOX_SHORT',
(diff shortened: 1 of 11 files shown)


lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
```diff
@@ -57,8 +77,14 @@ const EditCustomFieldsRecord = ({
   backendModuleId,
   backendModuleName,
   entityType,
-  fieldComponent,
   columnCount,
+  dispatch,
```
BogdanDenis, line 81: Where does this component get access to ‘dispatch’ from?
I think it’ll be better if we’re not giving the component access to ‘dispatch’ directly but bind actions in ‘mapDispatchToProps’

lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
```diff
@@ -100,29 +182,41 @@ const EditCustomFieldsRecord = ({
...
       });
+    } else if (customField.type === fieldTypes.RADIO_BUTTON_SET) {
+      return createCustomFieldRenderFunction(customField)({
+        children: getRadioButtonSetChildren(customField),
+        label: (
+          <Label required>
+            {renderCustomFieldLabel(customField)}
+          </Label>
+        ),
+        required: true,
```
BogdanDenis, line 218: Are all radio set fields required by default?

MaksymDryha:
> Looks great! Could you please update changelog and documentation with this description, because it could be difficult for some developers to find how working with redux form or final-form

The docs definitely have to be updated, good point, thank you!

lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
MaksymDryha: @BogdanDenis , makes sense!

lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
```diff
@@ -88,8 +113,65 @@ const EditCustomFieldsRecord = ({
...
+      });
+    };
+
+    if (customFieldsLoaded && sectionTitleLoaded) {
+      initializeFields();
+    }
+  }, [
+    customFieldsLoaded,
+    sectionTitleLoaded,
+    changeFinalFormField,
```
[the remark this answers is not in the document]
MaksymDryha: @BogdanDenis, I would prefer to follow the rule that encourages us to not lie about effect dependencies

lib/CustomFields/pages/EditRecordCustomFields/EditRecordCustomFields.js
```diff
@@ -0,0 +1,265 @@
...
+          const fieldIsEmpty = !hasIn(finalFormCustomFieldsValues, cf.refId);
+
+          if (fieldHasOptions && fieldIsEmpty) {
+            const valueToSet = cf.selectField.defaults[0];
+
+            changeFinalFormField(`customFields.${cf.refId}`, valueToSet);
+          }
+        }
+      });
+    };
```
Vladyslav-Velytskyi, line 142:
```suggestion
          const fieldIsEmpty = !hasIn(reduxFormCustomFieldsValues, cf.refId);

          if (fieldHasOptions && fieldIsEmpty) {
            const valueToSet = cf.selectField.defaults[0];

            if (isReduxForm) {
              changeReduxFormField(`customFields.${cf.refId}`, valueToSet);
            } else {
              changeFinalFormField(`customFields.${cf.refId}`, valueToSet);
            }
          }
      });
    };
```

little bit hard coded, you duplicate yourself

MaksymDryha: @Vladyslav-Velytskyi , we have to know if what type of form is used before calling 
`const fieldIsEmpty = !hasIn(reduxFormCustomFieldsValues, cf.refId);
`
or
`
const fieldIsEmpty = !hasIn(finalFormCustomFieldsValues, cf.refId);
`

Vladyslav-Velytskyi: oh, yes, you're right, sorry, i missed that first arguments are different

lib/CustomFields/constants.js
```diff
@@ -99,6 +100,7 @@ export const fieldComponents = {
   [fieldTypes.TEXTAREA]: TextArea,
   [fieldTypes.CHECKBOX]: Checkbox,
   [fieldTypes.SELECT]: Select,
+  [fieldTypes.RADIO_BUTTON_SET]: RadioButtonGroup,
```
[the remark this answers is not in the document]
MaksymDryha: Makes sense! I added a commit with this change

PR #813, opened 2020-05-15
STSMACOM-331: Increase custom field coverage to 80%

BogdanDenis:
## Purpose
Currently we have a tech debt to cover Custom Fields functionality with tests and increase coverage of code related to this feature. This PR covers writing tests for Custom Fields functionality


zburke:
> The problem is that I haven't found a way to mock `store.discovery.modules` object during testing, so during testing requests weren't being sent. 

Did you figure this out what you needed here? It isn't really clear to me what you are asking. 

The `discovery.modules` slice of the store is copied onto `props.stripes` so you can access it there. Alternatively, if you need to modify the way `stripes-core` is configured for testing, that happens in the [setup-application](https://github.com/folio-org/stripes-core/blob/master/test/bigtest/helpers/setup-application.js) and [stripes-config](https://github.com/folio-org/stripes-core/blob/master/test/bigtest/helpers/stripes-config.js) test-helpers.

zburke:
Also, I'm not thrilled about turning off the password-validation test; on master, it passes locally and in CI.

lib/CustomFields/utils/useCustomFieldsFetch.js
```diff
@@ -29,9 +29,7 @@ const useCustomFieldsFetch = (okapi, backendModuleId, entityType) => {
       }
     };
 
-    if (backendModuleId) {
```
MaksymDryha, line 32: This check shouldn't be removed. It's possible that the backend doesn't manage to send the response to the `/_/proxy/tenants/diku/modules` request before `useCustomFieldsFetch` is called. This mainly applies to situations when a user visits a page, that fetches custom fields, using a direct link (`/settings/users/custom-fields`, `/users/preview/some-id`, etc). In such cases, this change will lead to making a request for custom fields even though we don't know the `backendModuleId` yet.

lib/CustomFields/components/CustomFieldsForm/tests/CustomFieldsForm-test.js
```diff
@@ -0,0 +1,226 @@
...
+    expect(customFieldsForm.sectionTitleValue).to.equal('Custom fields');
+  });
+
+  describe('when clicking on an accordion', () => {
+    let initialOpenState;
+
+    beforeEach(async () => {
+      await customFieldsForm.whenLoaded();
+
+      customFieldsForm.customFieldAccordions(0).isOpen;
```
MaksymDryha, line 86: is this needed?

BogdanDenis: nope, that's what's left of a `console.log`:)

lib/CustomFields/utils/useCustomFieldsFetch.js
BogdanDenis: Reverted this change in a new commit

BogdanDenis:
@zburke thanks for pointing me in the right direction, I've resolved those problems and updated PR

PR #820, opened 2020-05-23
correctly prefix keys in IntlHarness

zburke:
The `<IntlHarness>` provided the context correctly but didn't prefix the
translation keys with the module-id, making it functionally useless.


tests/IntlHarness.js
```diff
@@ -10,10 +10,26 @@ import PropTypes from 'prop-types';
...
+    res[`${prefix}.${key}`] = obj[key];
+  }
+  return res;
+}
+
 const IntlHarness = ({ children }) => (
   <IntlProvider
     locale="en-US"
-    messages={messages}
+    messages={prefixKeys(messages, 'stripes-smart-components')}
```
mkuklis, line 32: @zburke I wonder if we could use `https://lodash.com/docs/4.17.15#mapKeys` instead of implementing it by hand but it all looks good.

zburke: Thanks for the tip; I didn't know about `mapKeys`.

PR #823, opened 2020-05-27
<Settings> supports new optional `additionalRoutes` property

MikeTaylor:
If provided, this is an array of `<Route>` objects that are included
in the `<Settings>` component's routing for its sub-pane, but which
are not displayed in the visible list of settings sections.

This is useful for routing paths below the top level of a settings
page. For example, if you want your Foo module's Bar settings page to
be able to see a specific Bar in detail, you might pass `pages`
including:

	{
	  route: 'bar',
	  id: 'Maintain all Bars',
	  component: ListOfBar,
	},

and also `additionalRoutes` including:

      <Route
        path={`${match.path}/bar/:id`}
        render={p => <SingleBar {...p} label="Maintain specific Bar" />}
      />


VictorSoroka1:
This is useful stuff. I recently searched for such an option and opted out to define the route within my app rather than through Settings component.

MikeTaylor:
> This is useful stuff. I recently searched for such an option and because of lack of such I opted to define the route within my app rather than through Settings component.

I wasn't able to make that work — hence the injection of the route into the Settings component itself. But this does also seem a more _honest_ approach that putting the route elsewhere.

PR #836, opened 2020-06-16
STSMACOM-355: Edit/View Custom Fields UI updates

Vladyslav-Velytskyi:
## Description
Providing UI updates to Edit/View Custom fields Pages according to [STSMACOM-355](https://issues.folio.org/browse/STSMACOM-355)

## Screenshots
image
image


lib/CustomFields/pages/ViewCustomFieldsSettings/ViewCustomFieldsSettings.js
```diff
@@ -67,7 +64,7 @@ const ViewCustomFieldsSettings = ({
 
   const { calloutRef } = useLoadingErrorCallout(customFieldsFetchFailed || sectionTitleFetchFailed);
 
-  const customFieldsExist = customFieldsLoaded && !isEmpty(customFields);
+  const isNewCustomField = isEmpty(customFields) && !!sectionTitle.value;
```
MaksymDryha, line 67: what about renaming it to something like `noDataSaved`? it guess I would make more sense since the name `isNewCustomField` doesn't imply that the variable value is somehow dependent on the `sectionTitle`

Vladyslav-Velytskyi: Nice suggestion, thanks!

PR #842, opened 2020-06-18
STSMACOM-370: Settings > Users > Custom Fields: Only 10 custom fields display

BogdanDenis:
## Purpose
- Apply limit to `/custom-fields` request to allow loading more than 10 Custom Fields

### DIFF
--- a/lib/CustomFields/utils/useCustomFieldsFetch.js
+++ b/lib/CustomFields/utils/useCustomFieldsFetch.js
@@ -14,7 +14,7 @@ const useCustomFieldsFetch = (okapi, backendModuleId, entityType) => {
 
   useEffect(() => {
     const fetchCustomFields = async () => {
-      const response = await makeOkapiRequest('custom-fields')({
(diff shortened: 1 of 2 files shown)


lib/CustomFields/utils/useCustomFieldsFetch.js
```diff
@@ -14,7 +14,7 @@ const useCustomFieldsFetch = (okapi, backendModuleId, entityType) => {
 
   useEffect(() => {
     const fetchCustomFields = async () => {
-      const response = await makeOkapiRequest('custom-fields')({
+      const response = await makeOkapiRequest('custom-fields?limit=2147483647')({
```
zburke, line 17: No magic numbers please!

1. use a constant, e.g. `MAX_RECORDS`
1. use a comment explaining where that value comes from, e.g. 
`// Okapi's max record-set size` or 
`// max value as requested in STSMACOM-370`

PR #851, opened 2020-06-23
STSMACOM-378: Add ability to customize path used after search is performed

mkuklis:
This PR adds a new prop to `SearchAndSort` called `basePath` which can be used to control which path should be used after search is performed.

https://issues.folio.org/browse/STSMACOM-378

This should help with addressing issue described in https://issues.folio.org/browse/UIIN-1074

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Fix cannot select a radio button when multiple Custom Field radio button sets are created. Refs STSMACOM-373.
 * Display custom fields accordion with a spinner while custom fields data is being loaded.
 * Add `label` prop to `<NotesSmartAccordion>` and `createFormTitle` for `<NoteForm>`. Part of UIREQ-457.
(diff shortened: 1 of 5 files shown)


SergiyVSergiyenko:
As I understand it, now every module that uses the SearchAndSort component (ui-inventory, ui-requests, etc) can pass a string to it as the value of the basePath property to explicitly determine the path in URL after performing the search. Right?

mkuklis:
Yes @SergiyVSergiyenko that's correct. The one thing I'm not sure about here is the name. Perhaps we should call it `searchPath`. Or keep it basePath if perhaps it will be useful for other things in the future?

SergiyVSergiyenko:
Yes, I agree with you, @mkuklis . `searchPath` sounds like we're building a search URL (in our case, the other way around). `rootPath` ​​does not fit either (this may not be necessarily the path to `root`). From my point of view, it's the right name. Thank you! Great solution!

PR #858, opened 2020-06-26
Add ClipCopy component

zburke:
Display a clickable-clipboard to copy text to the clipboard. 

I tried hard to get tests working for this component, honest, but never met with consistent success. There are some notes on `tests/ClipCopy-test.js` but the short version is that things just weren't reliable. Talking with @JohnC-80, we wonder if the problem is context not being available immediately after mount, but who knows? Not me, that's for sure. 

Refs [UIREQ-478](https://issues.folio.org/browse/UIREQ-478)


lib/ClipCopy/readme.md
```diff
@@ -0,0 +1,21 @@
+# ClipCopy
+
+Show a clipboard icon that copies text to the clipboard and displays a success toast
+
+## Usage
+
+```
+import React from 'react';
+import ClipCopy from '@folio/stripes/smart-components';
```
VictorSoroka1, line 9:
```suggestion
import { ClipCopy } from '@folio/stripes/smart-components';
```

lib/ClipCopy/tests/ClipCopy-test.js
```diff
@@ -0,0 +1,50 @@
...
+const ClipCopyMock = props => (
+  <CalloutHarness>
+    <ClipCopy {...props} />
+  </CalloutHarness>
+);
+
+//
+// this should work, but it doesn't, and I give up.
+// I can run
+//    expect(cc.hasClipboard).to.be.true;
```
VictorSoroka1, line 23: I am wondering whether the trick for waiting for the `hasClipboard` existence would help similar to how we use occasionally timer for stuff to mount. 

The idea is to add the following in the interactor and then wait for it at the end of `beforeEach`. It may though run into the same issue.
```
  whenLoaded() {
    return this.when(() => this. hasClipboard);
  }
```

zburke: Alas:
```
  beforeEach(async () => {
    await setupApplication();
    await mount(<ClipCopyMock text="some text" />);
    await cc.whenLoaded();
  });
```
still results in 
```
FAILED TESTS:
  ClipCopy
    ✖ "before each" hook for "shows a clipboard icon"
      Chrome 84.0.4147 (Mac OS X 10.14.6)
    Error: convergent assertion returned `false`
        at loop (tests/index.js:41569:17)
```

PR #860, opened 2020-06-30
UIREQ-467: Add changeable content to assigned accordion to Notes

Vladyslav-Velytskyi:
## Description

Add changeable content to NotesView and NotesForm components to show it in Assigned accordion. Part of [UIREQ-467](https://issues.folio.org/browse/UIREQ-467)

## Screenshot

image

### DIFF
--- a/lib/Notes/NoteCreatePage/NoteCreatePage.js
+++ b/lib/Notes/NoteCreatePage/NoteCreatePage.js
@@ -29,6 +29,7 @@ class NoteCreatePage extends Component {
     paneHeaderAppIcon: PropTypes.string.isRequired,
     paneTitle: PropTypes.node,
     referredEntityData: referredEntityDataShape.isRequired,
+    renderReferredRecord: PropTypes.func,
     resources: PropTypes.shape({
       noteTypesData: PropTypes.shape({
         records: PropTypes.arrayOf(noteTypesCollectionShape).isRequired,
@@ -121,6 +122,7 @@ class NoteCreatePage extends Component {
 
   render() {
     const {
+      renderReferredRecord,
       referredEntityData,
       entityTypeTranslationKeys,
       paneHeaderAppIcon,
@@ -147,6 +149,7 @@ class NoteCreatePage extends Component {
           {pageTitle => (
             <TitleManager record={pageTitle}>
               <NoteForm
+                renderReferredRecord={renderReferredRecord}
                 noteTypes={this.getNoteTypesSelectOptions()}
                 referredEntityData={referredEntityData}
                 entityTypeTranslationKeys={entityTypeTranslationKeys}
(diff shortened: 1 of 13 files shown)


lib/Notes/components/ReferredRecord/tests/referredRecord-test.js
```diff
@@ -0,0 +1,68 @@
...
+  const referredRecord = new ReferredRecordInteractor();
+  const referredEntityData = {
+    name: 'Test Name',
+    type: 'Type',
+    id: '1',
+  };
+  const entityTypeTranslationKeys = { [referredEntityData.type]: 'Test' };
+
+  describe('rendering referredRecord component', () => {
+    setupApplication();
```
BogdanDenis, line 27: Do we need separate `setupApplication` calls for each `describe` here?

Vladyslav-Velytskyi: No, we don't. Thanks for noticing.

lib/Notes/NoteCreatePage/NoteCreatePage.js
```diff
@@ -18,6 +18,7 @@ import { referredEntityDataShape } from '../components/NoteForm/noteShapes';
 @stripesConnect
 class NoteCreatePage extends Component {
   static propTypes = {
+    assignedContent: PropTypes.node,
```
MaksymDryha, line 21: let's make it a function named `renderReferredRecordData`. `assignedContent` implies that we are passing something that will entirely replace the content of the "Assigned" accordion

lib/Notes/components/ReferredRecord/tests/interactor.js
```diff
@@ -0,0 +1,12 @@
+import {
+  interactor,
+  text,
+} from '@bigtest/interactor';
+
+export default interactor(class PreferredRecordInteractor {
```
MaksymDryha, line 6: `class ReferredRecordInteractor`

lib/Notes/NoteViewPage/components/NoteView/NoteView.js
```diff
@@ -105,12 +106,14 @@ export default class NoteView extends Component {
...
   renderReferredRecord() {
     const {
+      assignedContent,
       referredEntityData,
       entityTypeTranslationKeys,
     } = this.props;
 
     return (
       <ReferredRecord
+        assignedContent={assignedContent}
```
MaksymDryha, line 116: I don't think we should render `assignedContent` inside `ReferredRecord`. Referred record renders default data of the referred entity, `assignedContent` (`renderReferredRecordData` is a better name, I guees) allows us to choose what data should be rendered and to format this data properly. `ReferredRecord` and `renderReferredRecordData` are equivalent, so it would be better to have something like `props.renderReferredRecordData() || <ReferredRecord />`

PR #862, opened 2020-07-01
STSMACOM-382: Prevent PUT request for accordion title during drag'n'drop custom field

Vladyslav-Velytskyi:
## Description
After drop custom field on new place EditCustomFieldSettings sends request to set Accordion title. It must be prevent because drag'n'drop doesn't change accordion title.

## Issue
[STSMACOM-382](https://issues.folio.org/browse/STSMACOM-382)

## Screenshot
OGG8vDud1B

### DIFF
--- a/lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
+++ b/lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
@@ -173,13 +173,16 @@ const EditCustomFieldsSettings = ({
     }));
   };
 
-  const saveCustomFields = async (data, redirectToView = true) => {
+  const saveCustomFields = async (data, redirectToView = true, titleShouldUpdate = true) => {
     setSubmitting(true);
 
-    Promise.all([
-      updateCustomFields(formatCustomFieldsForBackend(data)),
-      updateTitle(data.sectionTitle),
-    ]).then(responses => {
+    const requests = [updateCustomFields(formatCustomFieldsForBackend(data))];
+
+    if (titleShouldUpdate) {
+      requests.push(updateTitle(data.sectionTitle));
+    }
+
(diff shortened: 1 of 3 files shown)


lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
```diff
@@ -173,25 +173,36 @@ const EditCustomFieldsSettings = ({
...
 
-    Promise.all([
-      updateCustomFields(formatCustomFieldsForBackend(data)),
-      updateTitle(data.sectionTitle),
-    ]).then(responses => {
-      if (responses.every(({ ok }) => !!ok)) {
-        if (redirectToView) {
-          history.push(viewRoute);
-        }
+    if (titleShouldUpdate) {
```
BogdanDenis, line 179: For me it feels like there are too many conditions and it's hard to understand what's going on here. Plus, there's a lot of duplicated code inside both `if` and `else`
I'd suggest to first build an array of requests, and then use `Promise.all`.

```suggestion
    const requests = [updateCustomFields(formatCustomFieldsForBackend(data)];
    
    if (titleShouldUpdate) {
      requests.push(updateTitle(data.sectionTitle);
    }
    
    Promise.all(requests).then(responses => {
      if (responses.every(({ ok }) => !!ok)) {
        if (titleShouldUpdate && redirectToView) {
          history.push(viewRoute);
        }
        
        setCustomFields(formatCustomFieldsForSaving(data));
      } else {
        showErrorNotification();
        setSubmitting(false);
      }
    });
```

Vladyslav-Velytskyi: I did not even think of such a decision. You're right, your suggestion is much better. Thanks.

PR #863, opened 2020-07-02
STSMACOM-385: Add `entityTagsPath` to `Tags`

mariia-aloshyna:
## Description
Provide `entityTagsPath` prop to `<Tags>` to allow implementors to set custom entity tags path. (See the usage of it [here](https://github.com/folio-org/ui-data-import/blob/UIDATIMP-499/src/settings/JobProfiles/ViewJobProfile.js#L421))

## Approach
Add `entityTagsPath` prop to `<Tags>`
Set tags to an entity by the provided path.
Set `entityTagsPath` = "tags" by default

## Refs
[STSMACOM-385](https://issues.folio.org/browse/STSMACOM-385)
[UIDATIMP-499](https://issues.folio.org/browse/UIDATIMP-499)

### DIFF
--- a/lib/Tags/Tags.js
+++ b/lib/Tags/Tags.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
-import { get, uniq, sortBy, difference, noop, isFunction } from 'lodash';
+import { get, set, uniq, sortBy, difference, noop, isFunction } from 'lodash';
 
 import { Callout, Pane } from '@folio/stripes-components';
 import { stripesConnect } from '@folio/stripes-core';
@@ -28,6 +28,7 @@ class Tags extends React.Component {
 
   static propTypes = {
     children: PropTypes.func,
+    entityTagsPath: PropTypes.string,
     getEntity: PropTypes.func,
     getEntityTags: PropTypes.func,
     link: PropTypes.string.isRequired,
@@ -44,6 +45,7 @@ class Tags extends React.Component {
   };
 
   static defaultProps = {
+    entityTagsPath: 'tags',
     onToggle: noop,
     getEntity: props => get(props, ['resources', 'entities', 'records', 0], {}),
     getEntityTags: props => {
@@ -74,13 +76,13 @@ class Tags extends React.Component {
     const {
       getEntity,
       getEntityTags,
+      entityTagsPath,
     } = this.props;
     const entity = getEntity(this.props);
     const tags = getEntityTags(this.props);
     const tagList = tags.filter(t => t !== tag);
 
-    entity.tags = { tagList };
-
+    set(entity, entityTagsPath, { tagList });
     this.props.mutator.entities.PUT(entity);
   };
 
@@ -89,11 +91,13 @@ class Tags extends React.Component {
     const {
       getEntity,
       getEntityTags,
+      entityTagsPath,
     } = this.props;
     const entity = getEntity(this.props);
     const tagList = getEntityTags(this.props);
+    const tagsToSave = { tagList: sortBy(uniq([...tags, ...tagList])) };
 
-    entity.tags = { tagList: sortBy(uniq([...tags, ...tagList])) };
+    set(entity, entityTagsPath, tagsToSave);
     this.props.mutator.entities.PUT(entity);
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 5.0.0 (IN PROGRESS)
 
 * Increment `react-router` to `^5.2`. Refs STRIPES-674.
+* Add `entityTagsPath` to `Tags` to set custom entity tags path. Part of UIDATIMP-499.
 
 ## [4.2.0] (IN PROGRESS)
 
--- a/lib/Tags/TagsAccordion.js
+++ b/lib/Tags/TagsAccordion.js
@@ -12,8 +12,8 @@ import {
 
 import Tags from './Tags';
 
-const TagsAccordion = React.memo(({ link, getEntity, getEntityTags }) => (
-  <Tags link={link} getEntity={getEntity} getEntityTags={getEntityTags}>
+const TagsAccordion = React.memo(({ link, getEntity, getEntityTags, entityTagsPath }) => (
+  <Tags link={link} getEntity={getEntity} getEntityTags={getEntityTags} entityTagsPath={entityTagsPath}>
     {({
       entityTags,
       tagsForm,
@@ -40,6 +40,7 @@ const TagsAccordion = React.memo(({ link, getEntity, getEntityTags }) => (
 ));
 
 TagsAccordion.propTypes = {
+  entityTagsPath: PropTypes.string,
   getEntity: PropTypes.func,
   getEntityTags: PropTypes.func,
   link: PropTypes.string.isRequired,


mariia-aloshyna:
Hi @zburke Actually data-import stores them at `tags` property as well, but it has a different nesting. See attached screenshots, please.
tags_incorrect
tags_correct

zburke:
If I understand you correctly, you are saying other entities have a path like `entity.tags` but data-import stores them at `entity.profile.tags`. Forgive me, I don't follow how this is different than what I said.

mariia-aloshyna:
You are right. I will try to explain why data-import has a different structure. The thing is that profiles in data-import settings (job profile, match profile, and so on) have dependencies on each other, and deleting or adding these relationships is described in the properties `addedRelations` and `deletedRelations` (it contains an array of profiles' ids). While the profile information itself contained in the `profile` property (and `tags` is a part of profile information).
I hope it became clearer.

lib/Tags/Tags.js
```diff
@@ -74,13 +76,13 @@ class Tags extends React.Component {
     const {
       getEntity,
       getEntityTags,
+      entityTagsPath,
     } = this.props;
-    const entity = getEntity(this.props);
+    const entity = cloneDeep(getEntity(this.props));
```
zburke, line 81: Why is `cloneDeep` necessary now if it wasn't before?

mariia-aloshyna: Well, I used it because `set` modifies an entity, so I decided to work with copy, but I see that the previous approach also mutated the entity, so probably it's not really necessary here, I'll return as it was, thanks.

zburke:
So we are in agreement that data-import stores information on `entity.profile.*` that _literally every other service in the system_ stores on `entity.*` directly, including `tags` and (it appears from your screenshot) `metadata`. If that is the case, I'm inclined to view this as a bug in the backend data-import model rather than a bug in ui-data-import or stripes-smart-components and I think it should be fixed at the source. 

Look, I get it that the work is already done -- this right here is a working patch! -- but I think it's arguably the wrong work. An alternative perspective is that accepting `props.entityTagsPath` allows `<Tags>` to function as an adapter (in the design patterns sense) and that this increased flexibility is worth the increased complexity it introduces. (I'm still a little scarred from the beastly [array of 130+ props](/folio-org/stripes-smart-components/blob/cfb60d144e43a3798ee8eae8042f74cf5eaff269/lib/SearchAndSort/SearchAndSort.js#L76-L213) 😱  we now accept in `<SearchAndSort>` all in the name of flexibility.)

Let me ask a few other folks for review here to get their take on the situation. I will add a comment to the Jira ticket as well.

zburke:
@JohnC-80, @mkuklis, @doytch: my initial take on this PR is that flexibility it adds isn't worth the complexity it adds. Truthfully, the complexity within the actual component is minimal -- this PR is very clean -- but the non-standard data model really bugs me. What's your take?

zburke:
I agree, @mkuklis, the work in this PR itself is great: it's simple and clear and it preserves existing behavior while adding flexibility; I have no argument with the work itself. My problem is that I think it is the wrong work. To me, the smart component represents a tags interface and it is therefore the entity's responsibility to adapt to that interface, not the job of the interface to adapt to any different structure that is thrown at it. If we need an adapter here, it belongs in ui-data-import, not in stripes-smart-components. The Jira ticket even alludes to this: it's a ui-data-import ticket, not a stripes-smart-components ticket. 

I'm not going to stand in the way of merging this, but I'm not going to approve it. Personally, I think the system would be improved overall if entities that implemented common behaviors did so in common ways, i.e. that this problem could best be solved by changing the entity model to place the `tags` property where every other entity in the system places it. Failing that, I think this problem could also be handled by placing an adapter on the entity side of the relationship, much as you would write a driver for the ODBC interface rather than expecting the ODBC gateway itself to handle your custom DB. 

Finally, if folks insist on doing this work in stripes-smart-components, please clone UIDATAIMP-499 into the STSMACOM project, or create a linked ticket for it there so we can track the work as it pertains to the next stripes-smart-components release. [As noted in Slack](https://folio-project.slack.com/archives/C210UCHQ9/p1592420942329700): 

> PSA: When you have a ticket for an app bug (e.g. UIFOO-666) that you resolve by fixing something in a different library (e.g. stripes-components), please clone the original ticket into the relevant repository (e.g. STCOM) so we can assign the fix-version to a ticket on the project that contains the change. This makes it much easier to keep Jira and GitHub in sync when it comes to managing releases and we need to figure out “what version was this bug fixed in?“. :pray: 

Thank you. 

PS @TarasSpashchenko, can you please explain how nesting the tags property at `entity.profile.tags` instead of `entity.tags` will reduce the number of HTTP requests? That's a surprising finding and I would like to understand it.

TarasSpashchenko:
@zburke 
I think I did not explain my thought in detail. Because data structures are quite complex and tags can be assigned not only to “profile” but potentially to any other entities like “action”, “matching”, or “mapping” that is why `entity.profile.tags` is used instead of `entity.tags`. Otherwise, it would be a mess in `entity.tags`.

doytch:
I do see a rationale to keeping the base SSC version of `Tags` simple and handling just the 99% use case. I think sometimes we do need to realize that Stripes isn't intended to solve everybody's problem - it's intended to solve the most common problems.

In ERM, we have complicated tags because they're stored as objects rather than strings (because backend) so we needed a workaround as well. We wrote [our own version here](https://github.com/folio-org/stripes-erm-components/blob/master/lib/Tags/Tags.js) that we use instead. It continues to use `TagsForm` from smart-components, so just the container component is different. 

Whether that's preferable or not, I'm not gonna pretend to know with 100% certainty. On one hand, we're repeating ourselves a bit in ERM. OTOH, _this isn't the first time we've augmented `Tags` to add more "flexibility,"_ and I'd bet a stimulus cheque on the fact that we'll probably have to add more if we merge this one. There is a risk of SAS-ifying this thing if we continue to piecemeal changes like this.

I don't have a very strong feeling here, but given that I'm starting to see a bit of a slippery slope here and trend towards a growing API, I'm not a huge fan of this. I'd recommend the following:

- See if it's possible to turn `Tags` into more of a modular component. Options could include mimicking something like SASQ where `Tags` would take a `queryGetter`-like prop to get the tags resource, and pass default implementations of onAdd/onRemove as render props. 
- More simply, breaking out the work done by onAdd/onRemove/etc into importable functions. That would enable easier implementation of...
- ...custom `Tags` implementations like is done in ERM, while minimizing the amount of repetitive code.

I've noticed that my gut on this kind of stuff seems to be sometimes at odds with @skomorokh so it might be helpful if he's got a minute to chip in.

mariia-aloshyna:
@doytch If I got you right one of your suggestions is to pass custom onAdd/onRemove functions to <Tags> props, is that right? If yes then doesn't it add more "flexibility"?

doytch:
> @doytch If I got you right one of your suggestions is to pass custom onAdd/onRemove functions to props, is that right? If yes then doesn't it add more "flexibility"?

So I think I'd go one of two ways I'd go with this:
1. Leave `Tags` as-is and only supporting the overwhelmingly most common use case. That way, if you need to do something custom, you write a custom version of it like we did in ERM that renders the common `TagsForm`.
2. Introduce a new version of `Tags` (a la `SearchAndSortQuery`), say, `TagsContainer` that is highly modular. It'd let you pass in overrides to the various operations it does the way [SASQ lets you override its `filtersToString`, `queryStateReducer`, etc](https://github.com/folio-org/stripes-smart-components/blob/master/lib/SearchAndSort/SearchAndSortQuery.md#props). Once that's in place, we'd convert `Tags` to render `TagsContainer` and pass in some sane defaults to handle the 99% use case. And if you need something custom, you'd render `TagsContainer` directly by passing in your overrides.

Essentially, I'd like to avoid the slow buildup of sludge that started to happen with SearchAndSort. We got into that situation by adding another prop every time we needed something custom. It wasn't a big deal the first or second time, but it eventually became a beast.

VictorSoroka1:
IMO the current PR is quite simple and backward compatible with existing code (`defaultProps`) and does not bloat the component codebase now. As @TarasSpashchenko stated this is a reasonable case to do it this way and it is possible to do the adjustment here. SearchAndSort is another level of complexity than here (175 lines of code to just describe props and here we are discussing only about the fourth one).

mariia-aloshyna:
I agree that in an ideal world all of the components manage to support 100% of use cases without having to provide additional props. That's what we all strive for. But I don't think there is any significant complexity that this new prop adds. What's the issue with this prop, except that in a perfect world, we wouldn't have it? I'd prefer to not introduce this prop too, as well as many other props that many components provide to cover specific use cases. Having the container that you are suggesting to implement would introduce another layer of abstraction that devs would have to get familiar with and this adds some complexity too. On the other hand, copying this code in our module in order to change a couple of lines also doesn't seem to me to be a better idea.

zburke:
Now that I got that rant out of my system, I'm coming around, if too slowly, to the perspective that the "interface" to this component is its props and if we already provide props like `getEntity` and `getEntityTags` we might as well provide `entityTagsPath` (or perhaps `setEntityTags`) for parity. The default props implementations cover the 99% use-case as @doytch said, while allowing customization via those props for entities that need it.

mariia-aloshyna:
So do you think guys it can be merged or you insist on creating a custom component in ui-data-import (in this case, it will be necessary to monitor changes in the central component and apply them in the custom)?

doytch:
I still don't like the rationale that leads us in the community to contenting ourselves with patchwork fixes, but I'm not gonna stop this from being merged.

On the bright side, it may give @aditya-matukumalli and @EthanFreestone a chance to move off of the custom `Tags` implementations if it gives them trouble in the future.

PR #865, opened 2020-07-05
STSMACOM-380: Settings > Edit Custom Fields > change Save button label to Save & close

Vladyslav-Velytskyi:
## Description
Change `Save` button label to `Save & close`

## Issue
[STSMACOM-380](https://issues.folio.org/browse/STSMACOM-380)

## Screenshot
image

### DIFF
--- a/lib/CustomFields/components/CustomFieldsForm/CustomFieldsForm.js
+++ b/lib/CustomFields/components/CustomFieldsForm/CustomFieldsForm.js
@@ -156,7 +156,7 @@ const CustomFieldsForm = ({
         disabled={buttonsDisabled}
         data-test-custom-fields-save-button
       >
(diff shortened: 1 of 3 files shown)


lib/CustomFields/components/CustomFieldsForm/CustomFieldsForm.js
```diff
@@ -156,7 +156,7 @@ const CustomFieldsForm = ({
         disabled={buttonsDisabled}
         data-test-custom-fields-save-button
       >
-        <FormattedMessage id="stripes-smart-components.customFields.save" />
+        <FormattedMessage id="stripes-smart-components.customFields.saveAndClose" />
```
zburke, line 159: In fact we have `stripes-core.button.saveAndClose` for common labels like this, but the translation there is `Save and close` rather than `Save & close`. Your call.

PR #866, opened 2020-07-07
STSMACOM-386: Incorrect Last updated Source for Notes that were never updated

BogdanDenis:
## Purpose
If a Note was created and never updated then show username of the user who has created the note


lib/Notes/NoteViewPage/tests/interactor.js
```diff
@@ -0,0 +1,13 @@
+import {
+  interactor,
+} from '@bigtest/interactor';
```
Vladyslav-Velytskyi, line 3:
```suggestion
import { interactor } from '@bigtest/interactor';
```

BogdanDenis:
> You have only one uncovered condition, so please cover it, it will be enough to raise up to 100%

100% now 🎉

PR #867, opened 2020-07-07
Pull record name from nameKey

doytch:
When using the normal Edit button, the `nameKey` prop is used to render the delete confirmation modal. However, when using the actions dropdown, it just always uses `.name`.


doytch:
I spent more time rebasing the branch and rerunning Jenkins than I did writing the code 🤣

PR #869, opened 2020-07-09
STSMACOM-361: ability to delete custom fields options

MaksymDryha:
The purpose of this PR is to add the functionality of deletion custom fields options that haven't been save to records

## Delete buttons
image
## Confirmation message
image


lib/CustomFields/utils/useOptionsStatsFetch.js
```diff
@@ -0,0 +1,87 @@
...
+  useCallback,
+} from 'react';
+
+import { fieldTypesWithOptions } from '../constants';
+import makeRequest from './makeRequest';
+
+const isOptionUsed = option => option.count;
+const isEntityTypeMathing = entityType => option => entityType === option.entityType;
+
+
```
Vladyslav-Velytskyi, line 13: Please, delete this empty string

lib/CustomFields/utils/useOptionsStatsFetch.js
```diff
@@ -0,0 +1,87 @@
...
+  }, [customFields, getOptionUsageStatistics, entityType]);
+
+  return {
+    optionsStats,
+    optionsStatsLoaded,
+    optionsStatsFetchFailed,
+  };
+};
+
+
```
Vladyslav-Velytskyi, line 86: And this one empty string need to be removed

lib/CustomFields/pages/EditCustomFieldsSettings/tests/DeleteCustomFieldsOptions-test.js
```diff
@@ -0,0 +1,260 @@
...
+          });
+        });
+      });
+    });
+  });
+
+  describe('when there is a custom field with used options', () => {
+    beforeEach(async function () {
+      this.server.get('/custom-fields', () => ({
+        customFields: [{
```
Vladyslav-Velytskyi, line 206: I would recommend moving this object to a separate value. You use almost the same object earlier, so it would be better to minimize this point

lib/CustomFields/components/CustomFieldsForm/CustomFieldsForm.js
```diff
@@ -323,10 +361,13 @@ const CustomFieldsForm = ({
                       fieldsToDelete={formattedFieldsToDeleteData}
                       submitting={submitting}
                       handleConfirm={onConfirmDelete}
+                      optionsToDelete={optionsToDelete}
                       handleCancel={() => {
                         onCancelDelete();
                         insertDeletedFields(fields);
+                        insertDeletedOptions();
                       }}
+                      opt
```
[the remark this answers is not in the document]
MaksymDryha: @Vladyslav-Velytskyi , I have no idea on what `opt` is and why I added it here. Thanks

lib/CustomFields/components/CustomFieldsForm/tests/CustomFieldsForm-test.js
```diff
@@ -57,6 +57,10 @@ describe('CustomFieldsForm', () => {
         onFormReset={onFormReset}
         saveCustomFields={saveCustomFields}
         onSubmit={handleSubmit}
+        onOptionDelete={() => () => { }}
```
Vladyslav-Velytskyi, line 60:
```suggestion
        onOptionDelete={() => () => {}}

```

lib/CustomFields/pages/EditCustomFieldsSettings/tests/DeleteCustomFieldsOptions-test.js
```diff
@@ -0,0 +1,248 @@
+import React from 'react';
+import { describe, beforeEach, it } from '@bigtest/mocha';
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { mount, setupApplication } from '../../../../../tests/helpers';
```
Vladyslav-Velytskyi, line 6: Please, split each destructured value to separate lines, if them more then one

lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
```diff
@@ -91,14 +101,43 @@ const EditCustomFieldsSettings = ({
     }
   };
 
+  const handleOptionDelete = fieldId => optionData => {
+    const customField = customFields.find(cf => cf.id === fieldId);
+    const isOptionSaved = customField.selectField.options.values.find(option => optionData.id === option.id);
+
+    if (isOptionSaved) {
```
BogdanDenis, line 108: I'd format this block in this way:
```suggestion
    if (!isOptionSaved) {
      return;
    }
    
    setOptionsToDelete(current => {
      const newOptionsToDelete = { ...current };
      const fieldHasOptionsToDelete = fieldId in optionsToDelete;
      const option = {
        fieldName: customField.name,
        optionData,
      };
      newOptionsToDelete[fieldId] = fieldHasOptionsToDelete
        ? [...newOptionsToDelete[fieldId], option]
        : [option];
        
      return newOptionsToDelete;
    });
```

MaksymDryha: I don't like using `return` to just stop function execution because it's not semantically correct.
I'd prefer having human readable code, which focuses on what is happening in the app more:
```js
if(condition) {
  openModa();
  enableButton();
}
```
than having more technically detailed code 
```js
if(!condition) {
  return; // essentially, the same as stopFunctionExecution(). not very behavior driven-ish
}

openModa();
enableButton();
```

But in this particular case, it doesn't have much impact, so it's not my strong preference.

Vladyslav-Velytskyi: In my opinion, if a function is called as handleOptionDelete, which implies that it is performing some action and returning nothing, then I see no need for an empty return.

PR #874, opened 2020-07-15
STSMACOM-384: Remove the 'no data' message displayed when saving a record

SergiyVSergiyenko:
https://issues.folio.org/browse/STSMACOM-384

### Purpose
Remove the 'no data' message displayed when saving a record.

**Before**

before

**Approach**

Added display of `Loader` while the records are not loaded yet.

approach_2

### Additional challenge
Problems with passing tests in the `CustomFields` component have also been fixed. Achieved by mounting the component directly when using `setupApplication` instead of `mount` later on.

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -4,7 +4,7 @@ import { FormattedDate, FormattedMessage } from 'react-intl';
 import { Link } from 'react-router-dom';
 import { isEqual, uniqueId, pickBy } from 'lodash';
 
-import { Button, Callout, Col, ConfirmationModal, Modal, Pane, Paneset, Row } from '@folio/stripes-components';
+import { Button, Callout, Col, ConfirmationModal, Modal, Pane, Paneset, Row, Loading } from '@folio/stripes-components';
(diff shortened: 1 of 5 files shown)


lib/ControlledVocab/ControlledVocab.js
```diff
@@ -456,10 +456,12 @@ class ControlledVocab extends React.Component {
               onCreate={this.actuators.onCreate}
               onDelete={this.showConfirmDialog}
               isEmptyMessage={
-                <FormattedMessage
-                  id="stripes-smart-components.cv.noExistingTerms"
-                  values={{ terms: this.props.label }}
-                />
+                '?condition?'
```
zburke, line 459: Maybe 
```
const values = props.resources.values;
const condition = values.hasLoaded && values.records?.length === 0;
```

SergiyVSergiyenko: Yeah, exactly! I could not guess to check every property of the `values`! Thanks, @zburke, for the hint.
The only thing is the `hasLoaded` property is probably not suitable for us, because it remains `false` if no data has been received. But `isPending` is great for us.

PR #883, opened 2020-07-22
UIOR-591 Set record limit for libraries

AndreiShumski2:
https://issues.folio.org/browse/UIOR-591
Set record limit for libraries in `<LocationLookup>`

### DIFF
--- a/lib/LocationLookup/LocationModal.js
+++ b/lib/LocationLookup/LocationModal.js
@@ -145,7 +145,11 @@ export default class LocationModal extends React.Component {
 
     if (!campusId) return;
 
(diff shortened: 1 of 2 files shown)


lib/LocationLookup/LocationModal.js
```diff
@@ -145,7 +145,11 @@ export default class LocationModal extends React.Component {
 
     if (!campusId) return;
 
-    const params = { query: `(campusId=="${campusId}")` };
+    const params = {
+      query: `(campusId=="${campusId}")`,
+      limit: '1000',
```
aliaksei-chumakou, line 150: it will be a bug again if number of libraries is more than 1K in particular campus...

PR #895, opened 2020-07-31
STSMACOM-408: Extend header of <ControlledVocab> form

mariia-aloshyna:
## Description
`<ControlledVocab>` component renders a `<Pane>` header and an `<EditableList>` header with the same `label` prop. [UIDATIMP-592](https://issues.folio.org/browse/UIDATIMP-592) requires another label to be rendered at the `<EditableList>` header.
So this PR is about to add another prop to the `<ControlledVocab>` component to be able to render the `<EditableList>` component with another header.

## Task
[STSMACOM-408](https://issues.folio.org/browse/STSMACOM-408)

## Screenshot
Field protection


mariia-aloshyna:
> Given `ControlledVocab` creates the `Pane` where it displays `paneTitle={label}` (line 413), won't this block:
> 
> ```
> <Pane defaultWidth="fill" fluidContentWidth paneTitle={this.props.label}>
>   ...
>   <EditableList
>     label={listFormLabel || label}}
> ```
> 
> cause `label` to display twice for any component that calls `ControlledVocab` without `listFormLabel`? I'm not sure that's the desired behavior, though maybe it's no big deal. Dunno. What do you think?

Hi @zburke - Actually it was always displayed twice because before my changes the `label` prop was passed to the `<EditableList>` via rest parameters [here](https://github.com/folio-org/stripes-smart-components/pull/895/files#diff-6a43094b2f04c9ea138bb80cc1d5352dR418) and then used [here](https://github.com/folio-org/stripes-smart-components/blob/master/lib/EditableList/EditableListForm.js#L478)

PR #900, opened 2020-08-13
STSMACOM-411: Show correct number of characters in NoteList details

Vladyslav-Velytskyi:
## Description
Create function for getting HTML substring without tags

## Issue
[STSMACOM-411](https://issues.folio.org/browse/STSMACOM-411)

## Screenshot
A8IUeB04vC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Refactor from `bigtest/mirage` to `miragejs`.
 * Fix Custom Fields Edit page has multiple elements with the same ID. Fixes STSMACOM-410.
 * Add `listFormLabel` prop to `ControlledVocab` to set header of `EditableList`. Refs STSMACOM-408.
+* Show correct number of characters in NoteList details. Refs STSMACOM-411.
 
 ## [4.2.0] (IN PROGRESS)
 
--- a/lib/Notes/utils/getHTMLSubstring.js
+++ b/lib/Notes/utils/getHTMLSubstring.js
@@ -1,0 +1,54 @@
+const ELEMENT_NODE_TYPE = 1;
+
+/**
+ * Returns a substring of the required length with HTML tags
+ *
+ * @param {string} htmlString string with HTML elements
+ * @param {number} maxLength required max length of string
+ */
+export default function getHTMLSubstring(htmlString, maxLength) {
+  const element = document.createElement('div');
+  const fragment = document.createDocumentFragment();
+  let stringLength = 0;
+
+  element.innerHTML = htmlString;
+
+  function recurse(currentNode) {
+    const nodeValue = currentNode.nodeValue;
+    const clonedNode = currentNode.nodeType === ELEMENT_NODE_TYPE
+      ? currentNode.cloneNode(false)
+      : document.createDocumentFragment();
+
+    const currentStringLength = nodeValue
+      ? stringLength + nodeValue.length
+      : stringLength;
+
+    if (nodeValue && currentStringLength <= maxLength) {
+      stringLength += nodeValue.length;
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 * Refactor from `bigtest/mirage` to `miragejs`.
 * Fix Custom Fields Edit page has multiple elements with the same ID. Fixes STSMACOM-410.
 * Add `listFormLabel` prop to `ControlledVocab` to set header of `EditableList`. Refs STSMACOM-408.
+* Show correct numver of characters in NoteList details. Refs STSMACOM-411.
```
mariia-aloshyna, line 29:
```suggestion
* Show correct number of characters in NoteList details. Refs STSMACOM-411.
```

lib/Notes/utils/getHTMLSubstring.js
```diff
@@ -0,0 +1,52 @@
...
+export default function getHTMLSubstring(htmlString, maxLength) {
+  const element = document.createElement('div');
+  const fragment = document.createDocumentFragment();
+  let stringLength = 0;
+
+  element.innerHTML = htmlString;
+
+  function recurse(currentNode) {
+    const nodeValue = currentNode.nodeValue;
+    const clonedNode = currentNode.nodeType === 1
```
BogdanDenis, line 16: Please use a constant that describes this magic number

Vladyslav-Velytskyi: Done

lib/Notes/utils/tests/getHTMLSubstring-test.js
```diff
@@ -0,0 +1,56 @@
...
+  it,
+} from '@bigtest/mocha';
+import { expect } from 'chai';
+
+import { getHTMLSubstring } from '..';
+import { DETAILS_CUTOFF_LENGTH } from '../../constants';
+
+const getString = (length, character) => new Array(length).fill(character).join('');
+
+const createDOMElement = domObject => {
```
BogdanDenis, line 12: I think we shouldn't write any logic inside tests, because this may also be buggy and could cause unstable tests.
If I understood correctly what you're trying to do here - is generate a html string. If that's so then I'd prefer to do it by using string templates
For example:
```
const inputHTMLString = `
  <ul>
    <li>${getString(100, 'a')}</li>
    <li>${getString(100, 'b')}</li>
    <li>${getString(50, 'c')}${getString(50, 'd')}</li>
  </ul>
`
```
This accomplishes the same thing, but you don't need to understand `getStringFromElement` function to get what the test does and there wouldn't be a situation where you have to debug test code

Vladyslav-Velytskyi: Yes, but if we write as you indicated, then there is a space between each line, it is also considered as a character, so mocked lines need to be written in one line, which is why we will have a lint error(string-length). We can more simplest string, but it wouldn't cover all conditions. Additional functionality has been added in case we have to transfer dates of different sizes, configurations and can be easily scaled in the future.

MaksymDryha: @Vladyslav-Velytskyi, @BogdanDenis, I agree with Denis.  Too much code to grasp before moving on to the tests themselves.
Vlad, if you are concerned about huge html string, I'd suggest to move them to separate files. Not an elegant solution, but I think this is a reasonable price for having straightforward tests.

Vladyslav-Velytskyi: Done
@BogdanDenis , @MaksymDryha please review it

PR #903, opened 2020-08-20
STSMACOM-411: Fix 'Show more' button

Vladyslav-Velytskyi:


BogdanDenis (approved):

lib/Notes/utils/tests/helpers.js
```diff
@@ -0,0 +1,3 @@
+export default function getString(length, character) {
```
BogdanDenis, line 1: This should be a named export

BogdanDenis (changes requested):

lib/Notes/utils/tests/helpers.js
Vladyslav-Velytskyi: Fixed

PR #907, opened 2020-08-27
UIU-1495: Handle Aged to lost items in ChangeDueDateDialog

Baroquem:
https://issues.folio.org/browse/UIU-1495

This adds support for Aged to lost items in <ChangeDueDateDialog>. It also addresses a bug I encountered, wherein the "x items selected" text at the top of the dialog box would not change when items were selected or deselected.

Screen Shot 2020-08-26 at 5 16 18 PM

### DIFF
--- a/lib/ChangeDueDateDialog/ChangeDueDate.js
+++ b/lib/ChangeDueDateDialog/ChangeDueDate.js
@@ -249,7 +249,7 @@ class ChangeDueDate extends React.Component {
           { loans.length > 1 ?
             <SafeHTMLMessage
               id="stripes-smart-components.cddd.itemsSelected"
-              values={{ count: loans.length }}
+              values={{ count: Object.values(loanSelection).filter(Boolean).length }}
             /> : null
           }
           {warnings.map((warning, i) => (
--- a/lib/ChangeDueDateDialog/ChangeDueDate.js
+++ b/lib/ChangeDueDateDialog/ChangeDueDate.js
@@ -10,8 +10,7 @@ import SafeHTMLMessage from '@folio/react-intl-safe-html';
 import DueDatePicker from './DueDatePicker';
 import LoanList from './LoanList';
 import css from './ChangeDueDateDialog.css';
-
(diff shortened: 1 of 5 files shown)


lib/ChangeDueDateDialog/ChangeDueDate.js
```diff
@@ -11,7 +11,7 @@ import DueDatePicker from './DueDatePicker';
 import LoanList from './LoanList';
 import css from './ChangeDueDateDialog.css';
 
-const DueDateUnchangeableItemStatuses = ['Declared lost', 'Claimed returned'];
+const DueDateUnchangeableItemStatuses = ['Declared lost', 'Claimed returned', 'Aged to lost'];
```
SergiyVSergiyenko, line 14: Since we add items to the array for each new status, wouldn't it be better to move it into a constants file, as we already do in other components?

lib/ChangeDueDateDialog/ChangeDueDate.js
```diff
@@ -249,7 +250,7 @@ class ChangeDueDate extends React.Component {
           { loans.length > 1 ?
             <SafeHTMLMessage
               id="stripes-smart-components.cddd.itemsSelected"
-              values={{ count: loans.length }}
+              values={{ count: Object.values(loanSelection).filter(Boolean).length }}
```
SergiyVSergiyenko, line 253: Does it make sense to move the calculation of length into a variable?

lib/ChangeDueDateDialog/ChangeDueDate.js
Baroquem: Sure ... I didn't bother here because the scope is limited to a single class and there are no other similar constants, but it could be good for future expansion.

lib/ChangeDueDateDialog/ChangeDueDate.js
Baroquem: Sounds good!

SergiyVSergiyenko (approved):

PR #913, opened 2020-09-08
[STSMACOM-389] Refactor PasswordValidationField to use final-form instead of redux-form

yelyzavetakhokhlova:
[[STSMACOM-389]](https://issues.folio.org/browse/STSMACOM-389) Refactor PasswordValidationField to use final-form instead of redux-form

In an ideal world, we would issue a breaking change and `import Field`, `FieldArray `etc from `final-form` but in order to do this in a non-breaking way, we can refactor them to accept props like `fieldClass`, `fieldArrayClass`, etc. in order to allow the calling components to override the (default: redux-form) class that is instantiated.

## Test coverage
image


yelyzavetakhokhlova:
Hey, @zburke, I'm having trouble with tests. It seems like re-mounting component with new props doesn't work and it always fallbacks to the first mount. Maybe you know what may be the problem?

zburke:
Looks like tests are passing now! I thought maybe the changes in #917 would be required, but that still hasn't merged so I don't have a good explanation. 🤷

zburke (approved):

yelyzavetakhokhlova:
> Looks like tests are passing now! I thought maybe the changes in #917 would be required, but that still hasn't merged so I don't have a good explanation. 🤷
Yeah, they were passing before, but because I can't remount component, I can't fully cover the code. :(

PR #917, opened 2020-09-14
Fixing breaking tests in Stripes-Smart-Components

K-Felk:
Tests for EditCustomFields component were failing because they rely on
context that now has to be loaded via initialState variable.  Moved the
relevant context data to initialState.

refs: https://github.com/folio-org/ui-users/pull/1465


K-Felk:
okay, this is a new one...tests are running fine locally but breaking here.  WTF?

K-Felk:
In my previous fix, I didn't remove the old okapi endpoint for module information...here, I did take it out.  I wonder if this test environment is still using it?  I added it back and pushed that up.

K-Felk:
Ayup, that's it.  Remote testing is still using the okapi endpoint for module/interface data.  Now we know.

zburke (approved):
Smart work. Thank you!

mkuklis (approved):

mkuklis:
Thank you for fixing this @K-Felk!

PR #919, opened 2020-09-15
STSMACOM-431 Change due date modal has information about canceled requests

K-Felk:
The modal was not checking the status of requests, and since the back end 
includes data on cancelled requests, the modal was counting them as open.  
I wrote some checking into the modal so that it does not count requests 
which have been canceled.

refs: https://issues.folio.org/browse/STSMACOM-431


Baroquem (approved):

lib/ChangeDueDateDialog/ChangeDueDateDialog.js
```diff
@@ -298,6 +301,7 @@ class ChangeDueDateDialog extends React.Component {
     const openRequests = get(this.props, 'resources.openRequests.records', []);
 
     const requestCounts = openRequests.reduce((requests, record) => {
+      if (record.status === RequestCancelledStatus) { return requests; }
```
SergiyVSergiyenko, line 304:
```suggestion
      if (record.status === RequestCancelledStatus) return requests;
```
I like how you avoid counting canceled requests. Just a tiny bit of style advice. We usually don't use curly braces in a project when we return a single-word expression.

SergiyVSergiyenko (approved):

PR #924, opened 2020-09-22
STSMACOM-435: Custom Fields | Automate Deque aXe accessibility compliance checker

BogdanDenis:
## Purpose
This PR serves the purpose of integrating an accessibility compliance checker `axe-core` into existing tests to automate accessibility testing.

## Approach
`axe-core` can be run in `beforeEach` blocks when a component has been rendered and it returns a results object. Later in `it` blocks we can assert that axe hasn't found and violations and the component doesn't have any accessibility issues.

In `helpers.js` axe has been configured to ignore certain rules. This is because those rules do not make sense when axe is run against a single component and not a whole html document. Let's take a look at those rules to see what they are:
- `meta-viewport` - Zooming and scaling must not be disabled [Deque page](https://dequeuniversity.com/rules/axe/3.5/meta-viewport)
- `landmark-one-main` - Page must have one main landmark [Deque page](https://dequeuniversity.com/rules/axe/3.5/landmark-one-main)
- `page-has-heading-one` - Page must contain a level-one heading [Deque page](https://dequeuniversity.com/rules/axe/3.5/page-has-heading-one)
- `bypass` - Page must have means to bypass repeated blocks [Deque page](https://dequeuniversity.com/rules/axe/3.5/bypass)

As you can see, these rules have to be checked when a whole page is rendered, instead of a single component, as it is in our case.

Other rules may be disabled in each test file, if need be.

**Note:** Please make sure that the component is rendered before calling `axe.run` in `beforeEach`. Otherwise the results will be incorrect.

## Issue
[STSMACOM-435](https://issues.folio.org/browse/STSMACOM-435)


Vladyslav-Velytskyi (approved):
LGTM!!

zburke (approved):
Nice work!

yelyzavetakhokhlova (approved):

PR #925, opened 2020-09-24
UIU-1789: Execute <ChangeDueDate> operations in sequence.

Baroquem:
https://issues.folio.org/browse/UIU-1789

The change due date operation is very slow. For one or two items, this isn't too much of a problem, but it is when there are dozens or hundreds of loans being updated. The UI currently tries to send multiple change requests in parallel using `Promise.all()`, but the backend can't keep up, returning timeout errors for up to half of the requests at a time; and the whole operation can take several minutes to complete.

The new code tries to circumvent this problem by sending the requests one at a time, in series. This _normally_ avoids the timeout errors (but not always, possibly due to [this bug](https://issues.folio.org/browse/MODFEE-93) in mod-feesfines). Changing the due date for 200 loans still takes about four minutes to complete, but at least the changes succeed. This PR also slightly adjusts how the <ChangeDueDate> modals are presented, in order to give users a little more feedback when they click "Save and close." (At present, the pre-operation modal just sits there until the operation is complete, so users probably think the system is ignoring them!)


lib/ChangeDueDateDialog/ChangeDueDate.js
```diff
@@ -159,17 +160,27 @@ class ChangeDueDate extends React.Component {
...
-
-    Promise.all(promises)
-      .then(results => this.props.onDueDateChanged(results));
+      .map(loan => this.performLoanChange(loan).catch(e => e))
+
+      // Previously, this function used Promise.all() to execute these loan changes
+      // in parallel. But this was overloading the backend, where the transactions involved
+      // are complex, and causing many of the requests to time out with 500 errors. We're
+      // switching to sequential execution here in an attempt to keep the backend happy.
+      .reduce(
```
mkuklis, line 177: @Baroquem I'm glad this works! It maybe just me but it's very hard to follow what is happening here. Could this be replaced with just a simple for loop and async/await? 

For example:

````js
const loans = this.getUpdatedLoans();

for (let i = 0; i < loans.length; i++) {
   try {
     const result = await this.performLoanChange(loans[i]);
     results.push(result);
   }
   catch (e) {}
}

this.props.onDueDateChanged(results); 

````

mkuklis (approved):
Nice work @Baroquem!

zburke (approved):

SergiyVSergiyenko (approved):

lib/ChangeDueDateDialog/ChangeDueDate.js
Baroquem: @mkuklis, thanks for the suggestion. It is easier to read! I like it.

PR #927, opened 2020-09-25
STSMACOM-435: Hide labels for Custom Fields default checkbox

BogdanDenis:
## Purpose
- Hide labels for Custom Fields default checkbox, but leave them accessible via screen readers


Vladyslav-Velytskyi (approved):

zburke (approved):
I think this looks good but I defer to @JohnC-80 to confirm that this passes a11y muster.

JohnC-80 (approved):

PR #928, opened 2020-09-28
STSMACOM-441: Fix bug in ChangeDueDateDialog that did not allow data …

SergiyVSergiyenko:
https://issues.folio.org/browse/STSMACOM-441

### Purpose
The issue described in the ticket occurs after calling the `ChangeDueDate` modal again within one session.
It turned out that in the `getDerivedStateFromProps` method the data update occurs if the `ChangeDueDate` modal becomes open (the `props.open` property is responsible for this) and **the `state.open` value is set to `false`**. After updating the data, the `state.open` value is explicitly set to `true` (to avoid an infinite loop). And this value remains until the end of the session.
Consequently, the next calling of the `ChangeDueDateDialog` component (for example, after renewal loan) doesn't update the data (including the current loan).

### Approach
This PR addresses this by explicitly setting `state.open` to `false` when closing the modal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 * Added `headerProps` property to `NotesSmartAccordion` and `NotesAccordion`. Refs STSMACOM-439.
 * Execute `<ChangeDueDateDialog>` PUT requests in sequence. Refs UIU-1789.
 * Add accessibility tests for Custom Fields. Refs STSMACOM-435.
+* Fix bug in `<ChangeDueDateDialog>` that did not allow data to be updated. Fixes STSMACOM-441.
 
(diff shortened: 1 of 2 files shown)


mkuklis (approved):

Baroquem (commented):
@SergiyVSergiyenko Thanks for looking at this. It's a little confusing to have both `props.open` and `state.open` in play. It looks like `state.open` is only used in `getDerivedStateFromProps`; maybe it could be renamed to make its purpose clearer?

SergiyVSergiyenko:
Oh, thank you. That's a sensible remark.

K-Felk (approved):

Baroquem (approved):
Looks good to me!

PR #932, opened 2020-10-02
UIU-1495: Show an error message for Aged to lost items

Baroquem:
This is an addendum to https://github.com/folio-org/stripes-smart-components/pull/907, which added support for Aged to lost items to `ChangeDueDateDialog`. I forgot to add an error message for when the due date change attempt fails.


mkuklis (approved):

SergiyVSergiyenko (approved):

zburke (approved):
LGTM, @Baroquem. While this is a ui-users bug, since a portion of the fix is in stripes-smart-components, can you please clone the ticket into STSMACOM, or create a new one there so we can track which release includes this work?

Thanks!

PR #934, opened 2020-10-05
UIEH-963: Fix invalid ARIA attribute order in EditableList

BogdanDenis:
## Description
1. Remove unneeded `role="row"` attribute in `<EditableList>` item wrappers. It is unnecessary to add this attribute in `<EditableList>`, because `<MultiColumnList>` already adds `role="row"` to a wrapper element of each row
2. Add id to note type select header to be referenced as a label


Vladyslav-Velytskyi (approved):

JohnC-80 (approved):
Thank you!!

PR #942, opened 2020-10-14
STSMACOM-446: Update styling for EditableList component

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/STCOM-266

**Note:** This ticket was originally called `STCOM-266` but got moved to stripes-smart-components.

## Changes
- Updated row color of EditableListForm to alternate on odd/even rows (similar to MultiColumnList)
- Flipped the positioning of the cancel/save buttons and changed the button style of the "Save"-button to "primary"
- Added localized strings for cancel/save buttons

## Before
image

## After
image


lib/EditableList/EditableListForm.js
```diff
@@ -402,26 +402,27 @@ class EditableListForm extends React.Component {
...
+          </Button>
+          <Button
+            buttonStyle="primary"
+            disabled={pristine || submitting || invalid}
+            marginBottom0
+            id={`clickable-save-${this.testingId}-${item.rowIndex}`}
+            onClick={() => this.onSave(fields, item.rowIndex)}
+            {...(typeof actionProps.save === 'function' ? actionProps.save(item) : {})}
+          >
+            <FormattedMessage id="stripes-smart-components.save" />
```
zburke, line 423: Use `stripes-core.button.save` and `stripes-core.button.cancel` instead of defining these values locally. True, this problem is already epidemic, but let's not make it worse :)

rasmuswoelk: @zburke Ah! Thanks – I'll make the change.

We might want to make a page with a complete overview of universal" translations somewhere. Maybe in the Storybook. I made this for another project: https://framework.share-vde.dev/?path=/story/utils-translations--available-translations

zburke: Good thinking! I have an open story to consolidate our i18n documentation; perhaps this could be a part of that.

zburke (approved):

PR #949, opened 2020-10-30
STSMACOM-452 correctly retrieve related requests

zburke:
Previously, open requests were incorrectly retrieved without
qualification, rather than related only to items currently on loan. The
current approach (see #572, #575) based on manually inspecting state and
manually firing requests in `getDerivedStateFromProps` differs from how
connected components typically fetch data. That's fine; commentary in
PR #575 indicates this was deliberate due to the peculiarities of how
state is managed in the modal.

Here are the changes now in place:

* the `openRequests` query does not run automatically (by setting
`fetch: false` in the manifest)
* the results of the `loans` query, returned in a promise, are fed
directly to `getOpenRequestsQuery()` because, although they will also be
provided in `props.resources.loans`, `getOpenRequestsQuery()` is not
triggered based on a lifecycle method and therefore needs direct access
to the result from the promise.

I think the bug crept in because a manifest's `params` prop can be
implemented as a function that returns either a query string or,
importantly, `null` to indicate that required props are missing and the
fetch should not fire. In the previous implementation here,
`getOpenRequestsQuery` would return `null` under some circumstances and
a CQL query in others, but as this was not part of a `params` function,
returning `null` did not prevent the fetch from firing; it merely
prevented it from including the correct CQL query in the fetch URL.

In addition to running the unrestricted fetch, the correctly qualified
fetch would never run because `fetchData` would be called from `gDSFP`
only once (because state, rather than a change in props, is used as the
trigger), meaning the result of the `loans` query (which provides the
query for the `openRequests` query) was never used.

Whew.

Refs [STSMACOM-452](https://issues.folio.org/browse/STSMACOM-452)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 # Change history for stripes-smart-components
 
 ## 5.1.0 IN PROGRESS
-* Increase returnted Note Types per response limit. Refs STSMACOM-449.
 
+* Increase returned Note Types per response limit. Refs STSMACOM-449.
 * Do not execute search automatically when query index changes. Fixes STSMACOM-350.
 * `<CollapseFilterPaneButton>` must pass a string to `<ToolTip>`. Refs STSMACOM-447.
+* Correctly retrieve related requests when changing a loan's due date. Refs STSMACOM-452.
 
 ## [5.0.0](https://github.com/folio-org/stripes-smart-components/tree/v5.0.0) (2020-10-06)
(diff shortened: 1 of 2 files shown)


ViktorSoroka07 (commented):
Yeah, the main point that returning null was not preventing the request. I think this happened because previously it was based on the params function but after refactoring to #575 this no longer made sense. Great explanation @zburke, without it it would have a decent amount of time to figure it out.

mkuklis (approved):
Nice work @zburke!

PR #951, opened 2020-11-04
STSMACOM-454 allow escaping to be disabled

zburke:
`makeQueryFunction` now accepts another argument, `escape` (default:
`true`) to allow callers to turn off escaping of quote and backslash
values in CQL queries. This is necessary for features like
ui-inventory's "Query search" search-type that submits a fully-formed
CQL query.

Refs [STSMACOM-454](https://issues.folio.org/browse/STSMACOM-454), [UIIN-1319](https://issues.folio.org/browse/UIIN-1319).


mkuklis (approved):
Very nice @zburke. 👏

Baroquem (approved):

SergiyVSergiyenko (approved):

ViktorSoroka07 (approved):

PR #958, opened 2020-11-12
STSMACOM-457: Increase record limit for tags query

mkuklis:
https://issues.folio.org/browse/STSMACOM-457

The 100 limit started causing some issues in `bugfest-honeysuckle` were there are currently over 100 added tags.

The problem happens when only first 100 tags are fetched. When the new tag is being added which already exists in the db but it's not prefetched the frontend will treat it as a brand `new` tag and the POST request will be issued in order to create it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Correctly retrieve related requests when changing a loan's due date. Refs STSMACOM-452.
 * Allow configurable escaping in `makeQueryFunction`. Refs STSMACOM-454.
 * Show `callout` in `EntryManager` component. Refs STSMACOM-456.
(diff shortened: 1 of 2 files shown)


Baroquem (approved):

zburke (approved):
Your explanation, and fix, make sense. 

Out of curiosity, however, what happens when you run this request for 10k tags in the Honeysuckle environment: does the module tip over in the backend or handle things OK? Can the frontend actually cope with that much data?

SergiyVSergiyenko (approved):

mkuklis:
@zburke if there are more than 10k tags we will run into the same issue as with 100 currently.

mkuklis:
@zburke your question about the performance is also a valid one. I don't understand why we are pulling all these tags. All this should be pushed to the backend.

PR #959, opened 2020-11-16
STCOM-782: Return promise from createRecord in SearchAndSort

mkuklis:
https://issues.folio.org/browse/STCOM-782

It looks like ui-inventory is doing a correct thing already while submitting a form by checking the `submitting` flag:

https://github.com/folio-org/ui-inventory/blob/24a9d2cc4ad1a5156e9f9c0c21472008a98be053/src/edit/InstanceForm.js#L231

https://final-form.org/docs/final-form/types/FormState#submitting

The problem is that the `submitting` is always set to `false` because the submission handler doesn't currently receive a promise. 

This is causing "double submission" problems with instance form in ui-inventory:

https://issues.folio.org/browse/UIIN-1339
https://issues.folio.org/browse/UIIN-1340

There is some additional work which was also required to address it in ui-inventory

https://github.com/folio-org/ui-inventory/pull/1219

This PR takes care of the creation promise configured via `SearchAndSort`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Allow configurable escaping in `makeQueryFunction`. Refs STSMACOM-454.
 * Show `callout` in `EntryManager` component. Refs STSMACOM-456.
 * Increase record limit for tags query in `<Tags>`. Fixes STSMACOM-457.
+* Return promise from `createRecord` to make `submitting` work correctly. Fixes STCOM-782.
 
 ## [5.0.0](https://github.com/folio-org/stripes-smart-components/tree/v5.0.0) (2020-10-06)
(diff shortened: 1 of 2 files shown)


SergiyVSergiyenko:
Looks like a good catch, @mkuklis!
Does this have anything to do with the UIIN-1339 and UIIN-1340? If so, link please them in the description. And also a link to the task. Thank you.

SergiyVSergiyenko (approved):

mkuklis:
Thanks @SergiyVSergiyenko. I added more details to the description.

Baroquem (approved):

zburke (approved):
🕵️ ❤️ 👏

PR #963, opened 2020-11-24
UITAG-37 fix filtered tags sorting

aliaksei-chumakou:
first we show items that start with filtered text, sort them, and then we show items that contain filtered text but don't start with it.
https://issues.folio.org/browse/UITAG-37

### DIFF
--- a/lib/Tags/TagsForm.js
+++ b/lib/Tags/TagsForm.js
@@ -5,6 +5,8 @@ import { isEqual, difference, sortBy, noop } from 'lodash';
 
 import { MultiSelection } from '@folio/stripes-components';
 
+const localeCompareCB = (a, b) => a.value.localeCompare(b.value);
+
 class TagsForm extends React.Component {
   static propTypes = {
     entityTags: PropTypes.arrayOf(PropTypes.string),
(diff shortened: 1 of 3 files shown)


uladislausamets (approved):

lib/Tags/TagsForm.js
```diff
@@ -54,14 +56,16 @@ class TagsForm extends React.Component {
...
+    const filterLowercased = filter.toLowerCase();
+
+    const renderedItems = [
+      ...list
+        .filter(({ value }) => value.toLowerCase().startsWith(filterLowercased))
+        .sort(localeCompareCB),
+      ...list
+        .filter(({ value }) => !value.toLowerCase().startsWith(filterLowercased) && value.toLowerCase().includes(filterLowercased))
+        .sort(localeCompareCB),
+    ];
```
axelhunn, line 68: Shouldn't old variant be more effective?

aliaksei-chumakou: firstly, Old variant doesn't work right. so it might be more effective from memory or time consumption perspective, but it doesn't help.
secondly, I don't think that two filters instead of one makes so big difference

axelhunn (approved):

AndreiShumski2 (approved):

PR #966, opened 2020-11-25
UUIN-308 Add closedByDefault prop to TagsAccordion

AndreiShumski2:
Add `closedByDefault` prop to `<TagsAccordion />` for managing open/close accordion status by default

### DIFF
--- a/lib/Tags/TagsAccordion.js
+++ b/lib/Tags/TagsAccordion.js
@@ -12,7 +12,7 @@ import {
 
 import Tags from './Tags';
 
-const TagsAccordion = React.memo(({ link, getEntity, getEntityTags, entityTagsPath }) => (
+const TagsAccordion = React.memo(({ link, getEntity, getEntityTags, entityTagsPath, id }) => (
   <Tags link={link} getEntity={getEntity} getEntityTags={getEntityTags} entityTagsPath={entityTagsPath}>
     {({
       entityTags,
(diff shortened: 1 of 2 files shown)


uladislausamets (approved):

lib/Tags/TagsAccordion.js
```diff
@@ -32,6 +32,7 @@ const TagsAccordion = React.memo(({ link, getEntity, getEntityTags, entityTagsPa
             <FormattedNumber value={entityTags.length} />
           </Badge>
         )}
+        id={id}
```
aliaksei-chumakou, line 35: Have you tried to use `closedByDefault` prop? IMHO, it's much more clear solution

AndreiShumski2: Maybe it's better to add `{...rest}` props to the accordion to have the possibility to pass other props?

aliaksei-chumakou: I think it's a common requirement - to show accordion opened initially if some tags are selected. Maybe you can check `entityTags` for length

lib/Tags/TagsAccordion.js
```diff
@@ -32,6 +32,7 @@ const TagsAccordion = React.memo(({ link, getEntity, getEntityTags, entityTagsPa
             <FormattedNumber value={entityTags.length} />
           </Badge>
         )}
+        closedByDefault={!entityTags.length}
```
aliaksei-chumakou, line 35: to be on the safe side I would put `!entityTags.length`
```suggestion
        closedByDefault={!entityTags?.length}
```

aliaksei-chumakou (approved):

PR #967, opened 2020-11-26
STSMACOM-466: Assign/Unassign Notes Modal: Support searching Notes by note title and note details

BogdanDenis:
## Description
- Change Note search parameter from `title` to `search`
- Updated `notes` interface from `1.0` to `2.0`

## Related PRs
https://github.com/folio-org/ui-notes/pull/134
https://github.com/folio-org/ui-users/pull/1587

## Issue
[STSMACOM-466](https://issues.folio.org/browse/STSMACOM-466)


ViktorSoroka07 (approved):

Vladyslav-Velytskyi (approved):

package.json
```diff
@@ -15,7 +15,7 @@
   },
   "stripes": {
     "okapiInterfaces": {
-      "notes": "1.0",
+      "notes": "2.0",
```
zburke, line 18: Bumping the major version of an interface requires a major version change to this module, and hence a major version change to `@folio/stripes` as well. That's fine, but we need to announce this on #stripes-updates in order to give people time to plan for it, and we should mention it at #stripes-architecture so that folks are aware they can publish other breaking changes this quarter. The following UI repos in `platform-complete#snapshot` will also need to be updated:

```
$ grep -r '"tags"' ./*/package.json
./eholdings/package.json:      "tags": "1.0",
./finance/package.json:      "tags": "1.0",
./inventory/package.json:      "tags": "1.0"
./invoice/package.json:      "tags": "1.0",
./orders/package.json:      "tags": "1.0",
./organizations/package.json:      "tags": "1.0",
./plugin-find-organization/package.json:      "tags": "1.0"
./receiving/package.json:      "tags": "1.0",
./stripes-smart-components/package.json:      "tags": "1.0",
./tags/package.json:      "tags": "1.0"
```

We'll also need to coordinate this change with the folks maintaining `mod-notes` (@dmtkachenko) in order to keep the nightly build process running. If there are conflicts between the module descriptor generated by the stripes build and the interface versions required in other module, the folio-snapshot build will fail.

zburke (changes requested):
We need to coordinate the major-version change in this interface across many repositories in the project. Please do not merge this PR until that is organized.

package.json
BogdanDenis: Thanks for looking into this @zburke! I'm not sure why you've searched for `tags` instead of `notes` though...
Searching for `notes` gives:
```
$ grep -r '"notes"' ./*/package.json
./notes/package.json:      "notes": "1.0"
./stripes-smart-components/package.json:      "notes": "1.0",
./users/package.json:      "notes": "1.0"
```
Is there anything that is needed from my side apart from maybe creating PRs in these modules?

zburke: >  I'm not sure why you've searched for `tags` instead of `notes` though...

Because I got confused 🤦 Thanks for catching that. 

Your scope of work looks correctly. Are there any backend modules that need to coordinate similarly?

BogdanDenis: @zburke I'm not sure, @dmtkachenko do you know if there is any back-end work needed for this?

BogdanDenis: @zburke I've created PRs in `ui-users` and `ui-notes` with this version change as well
https://github.com/folio-org/ui-users/pull/1587
https://github.com/folio-org/ui-notes/pull/134

BogdanDenis: @zburke I've asked @dmtkachenko and he said that there's no back-end work required for this.

zburke: Thanks for following up, @BogdanDenis. Let me follow up with @JohnC-80 about when we want to merge this. We discussed it yesterday at #stripes-architecture but didn't nail down a date. I'm hoping/thinking early next week.

zburke: Let's plan to merge this tomorrow, Tuesday 2020-12-08. I'm going to coordinate it with a major version change in stripes-components as well, so I'll need to merge that change first and then will work my way up the hierarchy of deps within stripes.

aliaksei-chumakou: @zburke are you sure that `okapiInterfaces` from this is worth managing and it participates in interfaces checks? I suspect that only modules from stripes.config are involved in that and there is no stripes-smart-components... Maybe we need to include that interface dep in every modue that uses Notes component?

zburke: `@folio/stripes` includes `@folio/stripes-smart-components` and _is_ an explicit part of a platform's `stripes.config.js`. There is some legit debate about which module's `package.json` should express the interfaces consumed by SSC: here in SSC ([where the manifest is defined](https://github.com/folio-org/stripes-smart-components/blob/d6a3d9fcde84fdf958e9abd89d356a2e8626a0ef/lib/Notes/NoteCreatePage/NoteCreatePage.js#L47-L57)) or in the apps that consume this component? There are good arguments to be made on both sides:

1. If an app consumes some SSC components but not all of them, generating a module descriptor from SSC will include spurious dependencies. Obviously, this is bad. 
1. Static inspection of an app will not reveal okapi endpoints consumed only via a component imported from SSC, which could lead to omission of this interface from the module descriptor altogether. Obviously, this is bad. 

At the moment, we're in camp 1.

aliaksei-chumakou: thanks for clarification! then this should work!

zburke: Option 3: separate all components that define their manifests and connect their own components into their own repos. I don't love the repository proliferation, but honestly it feels like The Right Way, or at least the only way to avoid both 1. and 2. above. 

This would limit SSC to things like `<ControlledVocab>` and `<SearchAndSort>` where a UI module is still responsible for defining the manifest it passes in, or `<ClipCopy>` which consumes a context generated by stripes-core.

PR #973, opened 2020-12-17
SPECULATIVE: allow <EntryList> lines to be rendered by a nominated component

MikeTaylor:
As things stand, `<EntryList>` items are always rendered only as the value of a nominated field (often `name`). Sometimes it would be nice to show something more elaborate, e.g. the name and URL fields together. Something like the change outlined here should do it in a general and straightforward way. Thoughts?

**NOTE.** Do not merge this. I have not even tried to run it. It's a sketch.

### DIFF
--- a/lib/EntryManager/EntrySelector.js
+++ b/lib/EntryManager/EntrySelector.js
@@ -318,6 +318,7 @@ class EntrySelector extends React.Component {
       nameKey,
       addMenu,
       paneSetWidth
+      entryComponent
     } = this.props;
 
     const rows = this.filteredRows(contentData);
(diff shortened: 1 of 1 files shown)


mkuklis (approved):
Nice idea @MikeTaylor!

lib/EntryManager/EntrySelector.js
```diff
@@ -318,6 +318,7 @@ class EntrySelector extends React.Component {
       nameKey,
       addMenu,
       paneSetWidth
+      entryComponent
```
doytch, line 321:
```suggestion
      entryComponent: EntryComponent
```

In JSX, a lower-case tag is treated literally, so we need to rename this to start with a capital when destructuring.

doytch (commented):
I like this idea...ran into it myself when writing the Settings app for the Courses edge module.

MikeTaylor:
OK, so this is probably worth working up into a proper PR.

PR #983, opened 2021-01-12
STSMACOM-473: Refactor redux-form context in EmbeddedAddressForm

mkuklis:
https://issues.folio.org/browse/STSMACOM-473

The previous way of accessing reduxForm via `contextTypes` was not supported any more so this uses `ReduxFormContext` instead.

### DIFF
--- a/lib/AddressFieldGroup/AddressEdit/EmbeddedAddressForm.js
+++ b/lib/AddressFieldGroup/AddressEdit/EmbeddedAddressForm.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Field } from 'redux-form';
+import { Field, ReduxFormContext } from 'redux-form';
 import findIndex from 'lodash/findIndex';
 import cloneDeep from 'lodash/cloneDeep';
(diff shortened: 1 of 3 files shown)


lib/AddressFieldGroup/AddressEdit/EmbeddedAddressForm.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Field } from 'redux-form';
+import { Field, ReduxFormContext } from 'redux-form';
```
zburke, line 3: You can zap `ReduxFormContext` here now that you moved it into the wrapper.

zburke (approved):
Clear, simple, functional. Nicely done! 👏 👏

doytch (approved):

lib/AddressFieldGroup/AddressEdit/EmbeddedAddressForm.js
mkuklis: Thanks for catching this!

PR #985, opened 2021-01-15
Lint and console annoyances

MikeTaylor:


MikeTaylor:
Anyone know why "continuous-integration/jenkins/branch" would fail when "continuous-integration/jenkins/pr-merge" succeeds?

Is this merely one of those intermittent random failures where you have to keep trying the tests over and over until you get lucky and they happen to pass?

If so, is there a way to re-run just the failed test, so I don't risk losing my precious pass on the other test?

...

And is all this _really_ the most effectively to fix a couple of trivial lint errors?

doytch:
> If so, is there a way to re-run just the failed test, so I don't risk losing my precious pass on the other test?

If you hit the `Details` button to the right of the check itself on Github, it'll take you to this page, where if you hit Login, it'll log in via your GH creds. At that point you should hopefully see the Refresh icon that allows you to rerun that pipeline.

Screen Shot 2021-01-15 at 2 21 10 PM

MikeTaylor:
Brilliant, thanks @doytch!

PR #986, opened 2021-01-15
STSMACOM-474 react 17

zburke:
Increment `react` to `v17`.

Refs [STSMACOM-474](https://issues.folio.org/browse/STSMACOM-474), [STRIPES-722](https://issues.folio.org/browse/STRIPES-722)


zburke:
As noted in STRIPES-722 and in [other stripes-* PRs](https://github.com/folio-org/stripes-core/pull/993), we're not ready for react 17 right now.

PR #997, opened 2021-02-03
STSMACOM-468: A global way of implementing the column selector feature

rasmuswoelk:
## Tickets
https://issues.folio.org/browse/STSMACOM-468

## Changes
- Added `<ColumnManager>` component that can be useful to ease the process of implementing toggleable columns for  e.g. a `<MultiColumnList>`
- Added tests and documentation for `<ColumnManager>`
- Implemented  `<ColumnManager>` into `<SearchAndSort>`. This feature is opt-in and it will require a change in the individual modules to enable it. See more below.

## Usage
```js
    import { ColumnManager } from '@stripes/folio/smart-components';

    const columnMapping = {
        status: 'Status',
        name: 'Name',
        barcode: 'Barcode',
        username: 'Username',
        email: 'Email'
    };

    <ColumnManager
        id="users-list-columns" // Required
        columnMapping={columnMapping} // Required
        excludeKeys={['name']} // Exclude these keys from being toggleable
    >
        {({ visibleColumns, renderColumnsMenu, toggleColumn }) => {
            // Render UI
        }}
    </ColumnManager>
```

## Usage with SearchAndSort
The `<ColumnManager>` is implemented directly into the `<SearchAndSort>` but it will require a few small changes to enable the feature.

1. Implement the UI that is passed down to `actionMenu` via. the `renderColumnsMenu`-prop. [See example](https://github.com/folio-org/ui-inventory/compare/STSMACOM-468?expand=1#diff-0399c7f70f7645c1cfd8ef8c3b7f167b29a8dc385e86b129f0f46ac896a02144R320).
2. Remove the `visibleColumns`-prop from the `<SearchAndSort>` since it's now handled by the `<ColumnManager>`

## Considerations
- I decided to go with the render-prop pattern after discussing potential approaches with the Stripes Force team. However, I'm considering exporting it as a hook as well – e.g. `useColumnManager` to make the implementation even simpler when implementing it in functional components. The render-prop component could then be a wrapper component for the hook and be used for implementation in class components.
- I'm considering implementing this directly into the `<SearchAndSort>`-component to enable this functionality straight out of the box for modules that still use this component.

## Notes
- The visible columns are currently persisted using `sessionStorage` but could easily be switched out with `localStorage` later on – or changed via. a prop.
- The component is placed in `stripes-smart-components` because we'll most likely update the persistence logic to be handled on a user-level in the future.

## Implementation
Here is an example of the implementation in `ui-users`:
https://github.com/folio-org/ui-users/pull/1634/files

And in `ui-inventory` and `ui-requests` which uses `<SearchAndSort>`:
https://github.com/folio-org/ui-inventory/compare/STSMACOM-468?expand=1
https://github.com/folio-org/ui-requests/compare/STSMACOM-468?expand=1


Vladyslav-Velytskyi (approved):

ryandberger (approved):
In general this looks good to me. I'm okay with the sessionStorage approach and appreciate the flexibility to easily migrate to localStorage if need be. Nice work!

zburke (approved):
This looks great! I'm really pleased to have a consistent way to implement this functionality. Nice work.

PR #1003, opened 2021-02-09
STSMACOM-482: Upgrade ControlledVocab to final form

mkuklis:
This PR introduces a new prop called `formType` (which by default is set to `redux-form`) which can be used to switch `ControlledVocab` to use `final-form`.

https://issues.folio.org/browse/STSMACOM-482

There are couple other alternative approaches we could take here:

1. Expose a new component via `stripes-smart-components` called `ControlledVocabFF` or something similar.
2. Do not provide  backwards compatibility and just switch completely to final-form and bump a major version of stripes-smart-components.

I'm not sure which one is the best. The current approach allows us to move gradually. For example update ui-users, check if everything works, etc.

### DIFF
--- a/lib/ConfigManager/ConfigFinalForm.js
+++ b/lib/ConfigManager/ConfigFinalForm.js
@@ -1,0 +1,11 @@
+import React from 'react';
+
+import stripesFinalForm from '@folio/stripes-final-form';
+
+import ConfigForm from './ConfigForm';
+
+export default stripesFinalForm({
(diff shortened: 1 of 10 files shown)


Baroquem (approved):

SergiyVSergiyenko:
In my opinion, it's a decent approach.

SergiyVSergiyenko (approved):

zburke (approved):

PR #1012, opened 2021-02-22
STSMACOM-487: Notes list - Default descending date sort is not working 

BogdanDenis:
## Description
Apply default sorting of Notes by updated date when component is first loaded

## Issues
[STSMACOM-487](https://issues.folio.org/browse/STSMACOM-487)

### DIFF
--- a/lib/CustomFields/components/FieldAccordion/edit-fields/shared-fields/OptionsField/tests/OptionsField-test.js
+++ b/lib/CustomFields/components/FieldAccordion/edit-fields/shared-fields/OptionsField/tests/OptionsField-test.js
@@ -115,11 +115,6 @@ describe('OptionsField', () => {
     describe('and reaching a max number of options', () => {
       beforeEach(async () => {
         await optionsField.addOption();
-        a11yResults = await axe.run();
-      });
-
-      it('should not have any a11y issues', () => {
-        expect(a11yResults.violations).to.be.empty;
       });
 
       it('should hide Add option button', () => {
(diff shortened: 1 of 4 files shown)


Yuliia-Dovhal (approved):

Vladyslav-Velytskyi (approved):

lib/CustomFields/components/FieldAccordion/edit-fields/shared-fields/OptionsField/tests/OptionsField-test.js
```diff
@@ -115,11 +115,6 @@ describe('OptionsField', () => {
     describe('and reaching a max number of options', () => {
       beforeEach(async () => {
         await optionsField.addOption();
-        a11yResults = await axe.run();
```
mkuklis, line 118: @BogdanDenis was this giving you some trouble?

mkuklis (approved):

lib/CustomFields/components/FieldAccordion/edit-fields/shared-fields/OptionsField/tests/OptionsField-test.js
BogdanDenis: Yes, sometimes tests would fail on this step. And we actually shouldn't need a11y checks here because when maximum number of options is reached we're hiding a single button and nothing else - so there should be any a11y issues

mkuklis: Thanks @BogdanDenis . That makes sense. I ran into similar issues with axe too.

PR #1019, opened 2021-03-02
STSMACOM-490 export missing ConfigManager components

zburke:
`ConfigForm` was replaced with `ConfigReduxForm` (and a parallel
component `ConfigFinalForm`) but (a) neither was exported and (b)
`ConfigForm` continued to be exported but now is an entirely different
component.

Rather than try to restore things as they were since use of `ConfigForm`
appears to be confined to a couple of pages in ui-developer, this PR
simply exports `ConfigReduxForm` and `ConfigFinalForm`, restoring the
old behavior but under a new name. That's good enough, and small enough,
and gosh darn it, people should like it.

Refs [STSMACOM-490](https://issues.folio.org/browse/STSMACOM-490), [UID-72](https://issues.folio.org/browse/UID-72)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-smart-components
 
+## 6.0.1 (IN PROGRESS)
+
+* Export `ConfigReduxForm`, `ConfigFinalForm`. Refs STSMACOM-490.
+
 ## [6.0.0](https://github.com/folio-org/stripes-smart-components/tree/v6.0.0) (2021-02-25)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v5.0.0...v6.0.0)
 
--- a/index.js
+++ b/index.js
@@ -9,6 +9,8 @@ export { default as ClipCopy } from './lib/ClipCopy';
 
 export { default as ConfigManager } from './lib/ConfigManager';
 export { default as ConfigForm } from './lib/ConfigManager/ConfigForm';
+export { default as ConfigReduxForm } from './lib/ConfigManager/ConfigReduxForm';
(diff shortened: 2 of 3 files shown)


SergiyVSergiyenko (approved):

Baroquem (approved):

MikeTaylor (approved):
Looks good to me!

(But what actually is `<ConfigForm>` now, then?)

zburke:
@MikeTaylor, TBH, there may no longer be a compelling reason to export `ConfigForm`. `ConfigManager` provides CRUD handlers and `ConfigForm` is the form itself, inside a `Pane`, with a nice footer. That `Pane` gets wrapped `stripesForm` (or `stripesFinalForm`) in `ConfigReduxForm` or `ConfigFinalForm`, respectively. 

The reason to export `ConfigReduxForm` and `ConfigFinalForm` is so you can have a form that provides its own handlers in place of the CRUD defaults, as we do in ui-developer where session settings and token information needs to be manipulated but is not persisted via a request to Okapi. From that angle, you could sensibly argue that there is nothing config-manager-y about `ConfigReduxForm` any longer. That's true, but it does package up a bunch of boilerplate code. Maybe the right/long-term thing is to move those components up a level to something like `BoilerplateForm` and have `ConfigManager` delegate to `BoilerplateForm/ReduxFormWrapper` and `BoilerplateForm/FinalFormWrapper`. 

This maaaaybe would make their purpose more clear, but given it would just be a name change, nothing else, I'm not sure it is worth it.

MikeTaylor:
Agreed, this is not worth agonizing over too much :-)

(I was just curious, not asking for more work to be done.)

PR #1026, opened 2021-03-19
STSMACOM-479: Adding a new custom field focus

Yuliia-Dovhal:
## Purpose:
Adding a new custom field focus (Settings > Users > Custom Fields)
Related task: [STSMACOM-479](https://issues.folio.org/browse/STSMACOM-479)

## Screenshot:
Screenshot 2021-03-19 at 17 03 21
Screenshot 2021-03-22 at 15 09 07

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v6.0.0...v6.0.1)
 
 * Export `ConfigReduxForm`, `ConfigFinalForm`. Refs STSMACOM-490.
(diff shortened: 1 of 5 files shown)


BogdanDenis (approved):
Looks good to me. Let's just ask @mkuklis about the end of line style change. `.editorconfig` has `lf` as proper end of line symbols, but the files actually have crlf endings. Should we keep end of line changes or revert them to how they were?

mkuklis:
@BogdanDenis since we see LF` in the config let's just keep that.

mkuklis (approved):
Nice work @Yuliia-Dovhal!

Vladyslav-Velytskyi (approved):

PR #1028, opened 2021-03-23
STSMACOM-478: Make the move custom field icon a button.

BogdanDenis:
## Description
- Make Custom Fields drag and drop icon a button to improve accessibility
- Bump version of `react-beautiful-dnd` to `13.1.0`. This is needed because new version includes some bugfixes and accessibility improvements that we need.
There were no breaking changes that could affect our existing implementation so we can safely do the update. Here's the changelog of the library https://github.com/atlassian/react-beautiful-dnd/releases

## Issues
[STSMACOM-478](https://issues.folio.org/browse/STSMACOM-478)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-smart-components
 
+## [6.0.2](IN PROGRESS)
+
+* Make the move custom field icon a button. Refs STSMACOM-478.
+
 ## [6.0.1](https://github.com/folio-org/stripes-smart-components/tree/v6.0.1) (2021-03-03)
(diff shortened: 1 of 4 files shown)


BogdanDenis:
@zburke I'm assuming that changing the major version of `react-beautiful-dnd` is not considered as a breaking change in `stripes-smart-components` because the API of the `<FieldAccordion>` component hasn't changed. Is that correct?

Vladyslav-Velytskyi (approved):

mkuklis (approved):

zburke (approved):

Yuliia-Dovhal (approved):

PR #1044, opened 2021-04-17
STSMACOM-500: Fix MultiSelect when using CustomField via final form

mkuklis:
This PR uses `OnChange` handler via `react-final-form-listeners'` in order to correctly listen to changes from `MultiSelect`

https://issues.folio.org/browse/STSMACOM-500

@BogdanDenis would be possible for you to double check if things are working on your end with this setup?


BogdanDenis:
@mkuklis thank you for helping with investigating and fixing this bug!

BogdanDenis (approved):

PR #1048, opened 2021-04-26
STSMACOM-502 hooks approach for ColumnManager

NikitaSedyx:
purpose:
https://issues.folio.org/browse/STSMACOM-502
many modules are built with hooks, it would be nice to have one more way to configure columns manager

approach:
move visibility control to hook, define ColumnsMenu as a separate component, expose them + reuse in ColumnManager to support composition approach

initial tests are passed

documentation is updated

### DIFF
--- a/lib/ColumnManager/ColumnManager.js
+++ b/lib/ColumnManager/ColumnManager.js
@@ -30,9 +30,15 @@ const ColumnManager = ({ id, columnMapping, children, excludeKeys }) => {
     />
   ), [prefixId, columnMapping, visibleColumns, nonToggleableColumns, toggleColumn]);
 
+  /**
+   * @deprecated Will be deleted in version 7.0. Use renderColumnsMenu.
+   */
+  const renderCheckboxes = renderColumnsMenu;
+
   const renderProps = {
     visibleColumns,
     renderColumnsMenu,
+    renderCheckboxes,
     toggleColumn
   };
 
--- a/lib/ColumnManager/ColumnManager.js
+++ b/lib/ColumnManager/ColumnManager.js
@@ -2,71 +2,17 @@
  * ColumnManager
  */
 
-import React, { useState, useCallback, useMemo, useRef, memo } from 'react';
+import React, { useMemo, useRef, memo } from 'react';
 import PropTypes from 'prop-types';
-import { useIntl } from 'react-intl';
-import { Checkbox, MenuSection } from '@folio/stripes-components';
 
-const STORAGE = sessionStorage;
+import useColumnManager from './useColumnManager';
+import ColumnManagerMenu from './ColumnManagerMenu';
 
 const ColumnManager = ({ id, columnMapping, children, excludeKeys }) => {
-  const intl = useIntl();
-  const nonToggleableColumns = useRef(excludeKeys).current;
   const prefixId = `column-manager-${id}`;
-  const storageKey = `${prefixId}-storage`;
 
-  // The column mapping object determines the default visible columns and order
-  const defaultColumnKeys = Object.keys(columnMapping);
-
-  const initialVisibleColumns = useMemo(() => {
-    const stored = STORAGE.getItem(storageKey);
-    return stored ? JSON.parse(stored) : defaultColumnKeys;
-  }, [storageKey, defaultColumnKeys]);
-
-  const [visibleColumns, setVisibleColumns] = useState(initialVisibleColumns);
-
-  const orderedVisibleColumns = useMemo(() => {
-    const visible = new Set(visibleColumns);
-    return defaultColumnKeys.filter(key => visible.has(key));
-  }, [defaultColumnKeys, visibleColumns]);
-
-  /**
-   * Toggle column visibility
-   */
-  const toggleColumn = useCallback(key => {
-    setVisibleColumns(currVisibleColumns => {
-      const nextVisibleColumns = currVisibleColumns.includes(key) ?
-        currVisibleColumns.filter(k => key !== k) :
-        [...currVisibleColumns, key];
-
-      STORAGE.setItem(storageKey, JSON.stringify(nextVisibleColumns));
-
-      return nextVisibleColumns;
-    });
-  }, [storageKey]);
-
-  /**
-   * Render column checkboxes for toggling columns
-   */
-
-  const renderCheckboxes = useMemo(() => {
-    const nonToggleableSet = new Set(nonToggleableColumns);
-
-    return defaultColumnKeys
-      .filter(key => !nonToggleableSet.has(key))
-      .map(key => (
-        <Checkbox
-          data-test-column-manager-checkbox={key}
-          key={key}
-          name={key}
-          label={columnMapping[key]}
-          id={`${prefixId}-column-checkbox-${key}`}
-          checked={visibleColumns.includes(key)}
-          value={key}
-          onChange={() => toggleColumn(key)}
-        />
-      ));
-  }, [nonToggleableColumns, visibleColumns, toggleColumn, prefixId, defaultColumnKeys, columnMapping]);
+  const nonToggleableColumns = useRef(excludeKeys).current;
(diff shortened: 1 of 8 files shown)


aliaksei-chumakou (approved):

lib/ColumnManager/ColumnManager.js
```diff
@@ -75,19 +21,18 @@ const ColumnManager = ({ id, columnMapping, children, excludeKeys }) => {
...
+      excludeColumns={nonToggleableColumns}
+      toggleColumn={toggleColumn}
+    />
+  ), [prefixId, columnMapping, visibleColumns, nonToggleableColumns, toggleColumn]);
 
   const renderProps = {
-    visibleColumns: orderedVisibleColumns,
+    visibleColumns,
     renderColumnsMenu,
-    renderCheckboxes,
```
NikitaSedyx, line 90: as I see nobody uses this, and it looks like breaking change, should we update major version now?

NikitaSedyx:
smth wrong with sonar coverage report, it shows 0 for repo https://sonarcloud.io/dashboard?id=org.folio%3Astripes-smart-components

lib/ColumnManager/useColumnManager.js
```diff
@@ -0,0 +1,44 @@
...
+  const toggleColumn = useCallback(key => {
+    setVisibleColumns(currVisibleColumns => {
+      const nextVisibleColumns = currVisibleColumns.includes(key)
+        ? currVisibleColumns.filter(k => key !== k)
+        : [...currVisibleColumns, key];
+
+      STORAGE.setItem(storageKey, JSON.stringify(nextVisibleColumns));
+
+      return nextVisibleColumns;
+    });
```
axelhunn, line 35: We can use `Set` all the way here
```suggestion
    setVisibleColumns(currVisibleColumns => {
      const nextVisibleColumns = new Set(currVisibleColumns);
      const isDeleted = nextVisibleColumns.delete(key);

      if (!isDeleted) nextVisibleColumns.add(key);

      try {
        STORAGE.setItem(storageKey, JSON.stringify(nextVisibleColumns));
      } finally {
        return nextVisibleColumns;
      }
    });
```

lib/ColumnManager/readme.md
```diff
@@ -79,13 +79,56 @@ In the example below, we wrap the `<ColumnManager>` around the results pane and
...
+        <>
+            {/* Other menu sections */}
+            <ColumnManagerMenu
+                prefix="users"
+                columnMapping={columnMapping}
+                visibleColumns={visibleColumns}
+                toggleColumn={toggleColumn}
+            />
+        </>
+    );
```
axelhunn, line 111: Can I propose the API
```suggestion
    const columnManager = useColumnManager({ prefix: 'users', columnMapping });

    const actionMenu = () => (
        <>
            {/* Other menu sections */}
            <ColumnManagerMenu {...columnManager} />
        </>
    );
```

axelhunn (approved):

lib/ColumnManager/readme.md
NikitaSedyx: props are different with return value

lib/ColumnManager/useColumnManager.js
NikitaSedyx: I just moved initial logic to hook, + add/delete looks a bit complicated for me to compare with init version + IMO no reason to try catch here

lib/ColumnManager/useColumnManager.js
```diff
@@ -0,0 +1,44 @@
...
+const useColumnManager = (prefix, columnMapping) => {
+  const storageKey = `${prefix}-storage`;
+
+  // The column mapping object determines the default visible columns and order
+  const defaultColumnKeys = Object.keys(columnMapping);
+
+  const [visibleColumns, setVisibleColumns] = useState(() => {
+    const stored = STORAGE.getItem(storageKey);
+
+    return stored ? JSON.parse(stored) : defaultColumnKeys;
```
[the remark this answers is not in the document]
NikitaSedyx: it was initial implementation + I don't think it's a case for try catch

lib/ColumnManager/readme.md
axelhunn: Yes, the point is to adjust the return value too.
So we could use `prefix` and `columnMapping` just once as initializing values.

lib/ColumnManager/useColumnManager.js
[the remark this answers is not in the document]
axelhunn: I don't insist it's critical here, but...

"various browsers offer settings that disable localStorage. So a browser may support localStorage, but not make it available to the scripts on the page"
https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability

JSON.parse "throws a SyntaxError exception if the string to parse is not valid JSON"
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions

lib/ColumnManager/readme.md
NikitaSedyx: 1. prefix values are different
2. why should we return the same object from args when it can be used without? related to `columnMapping`
3. `columnManager` is a new object each render (at least with current implementation) and useMemo/useCallback won't make sense with such usage

axelhunn: Yes, we could use define `storageKey` as `${prefix}-columns-storage` in the hook (instead of just `${prefix}-storage`. The benefit is having just one prefix.

The point of returning the same `columnMapping` is just to make API surface cleaner - one entry point for all init is the hook.
And I wouldn't worry about `columnManager` being a new object. It's a common pattern for hooks, which is processed by React just fine, I believe.

mkuklis (approved):
Very nice work @NikitaSedyx!

lib/ColumnManager/readme.md
NikitaSedyx: `It's a common pattern for hooks` - destructing is used mostly everywhere, with it you suggestion doesn't make sense

otherwise memoized object is returned

axelhunn: For my suggestion I kept in mind the FinalForm `useField` API, where the whole bunch of `input` props is passed - and easily used - as an object:

```js
const { input } = useField(...);
//...
<TextField {...input} />
```

I also dare say there's maybe no need in memoizing the returned object.
At least there's not much of the object memoization here, only the individual useCallbacks:
https://github.com/final-form/react-final-form/blob/cf64402a407ee08264c2348a035ddb718a058082/src/useField.js#L133-L243

axelhunn: Or maybe, even more bold proposal, the "[hook returning component](https://dev.to/droopytersen/new-react-hooks-pattern-return-a-component-31bh)" pattern you mentioned recently is suitable here:
```js
const ColumnManager = useColumnManager({ prefix: 'users', columnMapping });

const actionMenu = () => (
  <>
    {/* Other menu sections */}
    <ColumnManager.Menu />
  </>

//..

<MultiColumnList
  id="list-users"
  visibleColumns={ColumnManager.columns}
  //...
/>
```

lib/ColumnManager/ColumnManager.js
NikitaSedyx: @zburke please advice

zburke: I'm inclined not to issue a breaking change for this. 

If you think it is worth removing because it isn't being used, let's  label it deprecated, issue a console warning, and remove it in the next release cycle that contains other breaking changes. A major release is disruptive. I don't think it's worth the churn just to formally remove something that, as you said, perhaps nobody is using anyway. 

How does that sound to you?

NikitaSedyx: sounds good to me, I'll try to reuse existing code to not introduce smth else for this prop

PR #1050, opened 2021-04-27
STSMACOM-498: View Notes Record: Numbered / Bulleted Lists are not retained after editing/saving it

BogdanDenis:
## Description
Fix issue when adding nested lists to note details and saving the lists are collapsed into one.
To fix this apply styles from React Quill library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Execute validation manually when record is being saved. Fixes STSMACOM-496.
 * Fix `<MultiSelect>` when using `<CustomField>` via final form. Fixes STSMACOM-500.
 * buggy @rehooks/local-storage 2.4.1 must be avoided. Refs STSMACOM-501.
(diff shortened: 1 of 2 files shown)


Vladyslav-Velytskyi (approved):

zburke (approved):

zburke:
> Also, disabled nested-interactive axe rule to fix failing tests. This was caused by a recent release of axe-core v4.2.0 that introduced this new rule (https://dequeuniversity.com/rules/axe/4.2/nested-interactive?application=axeAPI).
Even though tabindex="-1" can be used to disable keyboard navigation to an element (and is often used to create accessible popups etc), that rule thinks it creates an empty tab stop 😒

I am _really_ happy to learn about this and impressed you figured it out, BUT this kind of change really belongs in a separate PR with a separate Jira. I know this seems petty, but it's really helpful when trying to keep track of the work.

PR #1052, opened 2021-04-29
STSMACOM-489: Create/Edit Note | Added Display note as a pop-up field

BogdanDenis:
## Description
Added Display note as a pop-up field to Notes

## Issues
[STSMACOM-489](https://issues.folio.org/browse/STSMACOM-489)

## Screenshots
image


zburke (commented):
It doesn't make any sense to me that this component should have any knowledge of ui-users or ui-checkout. Those packages depend on this one, so for this to refer to them feels inside out. 

Maybe "display as popup" should be a prop provided by the app to this component?

BogdanDenis:
> It doesn't make any sense to me that this component should have any knowledge of ui-users or ui-checkout. Those packages depend on this one, so for this to refer to them feels inside out.
> 
> Maybe "display as popup" should be a prop provided by the app to this component?

There's a `showDisplayAsPopupOptions` prop which is provided by the Users app. But we can't do the same with actual Note properties (`popUpOnUser`, `popUpOnCheckOut`).
We don't expect that any other apps will have this functionality - this is only Users and Checkout specific. Plus, to support this feature and not complicate things too much we've enabled additional properties in mod-notes. So it's possible to save notes with arbitrary property names - that's why passing property names may cause unexpected behavior.

zburke:
I still don't understand why this component should have any knowledge of ui-users or ui-checkout. Those packages depend on this one, so for this to refer to them feels inside out.

I don't understand how "enabl[ing] additional properties in mod-notes" is in any way relevant to the API stripes-smart-components presents to UI apps.

BogdanDenis:
> I still don't understand why this component should have any knowledge of ui-users or ui-checkout. Those packages depend on this one, so for this to refer to them feels inside out.
> 
> I don't understand how "enabl[ing] additional properties in mod-notes" is in any way relevant to the API stripes-smart-components presents to UI apps.

If we need to have these properties outside of stripes-smart-components then we'll need to pass those names from ui-users as a prop to Notes. I'm just concerned that people will start passing any properties to Notes when it's just these two that are properly supported.

zburke (approved):

PR #1054, opened 2021-05-07
STSMACOM-506 avoid calling setState after unmount

zburke:
A Promise may return after a component has unmounted. If it does so, it
must not call `setState()`, which is a memory leak reported in the
console.

Refs [STSMACOM-506](https://issues.folio.org/browse/STSMACOM-506)


BogdanDenis:
Thanks for catching and fixing this issue @zburke !

BogdanDenis (approved):

Baroquem (approved):

Vladyslav-Velytskyi (approved):

PR #1061, opened 2021-05-24
STSMACOM-510 Update "remote-storage-mappings" interface

uladislausamets:
## Purpose

https://issues.folio.org/browse/STSMACOM-510
## Pre-Merge Checklist
Before merging this PR, please go through the following list and take appropriate actions.
- [ ] I've added appropriate record to the CHANGELOG.md
- Does this PR meet or exceed the expected quality standards?
  - [ ] Code coverage on new code is 80% or greater
  - [ ] Duplications on new code is 3% or less
  - [ ] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] If any API-related changes - okapi interfaces and permissions are reviewed/changed correspondingly
  - [ ] There are no breaking changes in this PR.

If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
       "password-validator": "1.0"
     },
     "optionalOkapiInterfaces": {
-      "remote-storage-mappings": "1.0"
+      "remote-storage-mappings": "2.0"
     }
   },
   "scripts": {


axelhunn (approved):

AndreiShumski2 (approved):

zburke:
Hmmmm, this exclusively upgrades the interface from `1.0` to `2.0` rather than `1.0 2.0`. If this were a straight interface dependency, that would be a breaking change, which would be a breaking change for stripes as well, and hence to all UI applications. 

We need to think about whether a major-version update to an optional interface constitutes a breaking change. @AndreiShumski2, @axelhunn, what do you think? I'm inclined to say yes since this effectively removes support, even if it was optional support, for this interface. 

If that's the case, we should revert this and implement it in R3 instead, when we anticipate a new major version of stripes. Alternatively, maybe we can provide `1.0 2.0` support given the parts of the API touched by this code.

axelhunn:
Will this format work - `1.0 2.0`.
I don't see it documented

axelhunn:
We have no breaking changes in UI for now and will happily support both 1 & 2 versions

zburke:
Yeah, `1.0 2.0` should work. I'll try to dig up documentation for why we use `1.0 2.0` instead of the semver standard `^1.0 || ^2.0`, but this syntax is [definitely in use throughout FOLIO already](https://github.com/folio-org/ui-inventory/blob/41da2c347d50623599f362dfad3cfa93f9c5e71b/package.json#L77).

axelhunn:
OK, we'll change it `1.0 2.0` today

PR #1076, opened 2021-06-14
STSMACOM-518 results list not reflecting selection on return from edit.

JohnC-80:
# Problem
After editing and saving a record, returning to the results list will not have the currently seleted item highlighted.

## Approach
Make use of MultiColumnList's `isSelected` prop so that objects containing a subset of like values can be used for criteria.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Add `NotePopupModal` component. Refs STSMACOM-509.
 * Always show `<NotePopupModal>` when popup notes exist. Refs STSMACOM-515.
 * Add `autofocusSearchField` prop to SearchAndSort. fixes STSMACOM-516
+* Use `isSelected` prop of MCL in SearchAndSort to highlight selected item with fewer criteria properties. fixes STSMACOM-518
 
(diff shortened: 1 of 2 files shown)


mkuklis (approved):

zburke:
Super glad you fixed this, @JohnC-80, but can you explain how you figured this out? How the right props for MCL's `isSelected` function are configured is ~a bit~ completely opaque to me.

JohnC-80:
@zburke So there's two props that work together in MCL - the first one is `selectedRow` it provides the 'criteria' object for whether or not the selected style is applied - then the `isSelected` prop runs a function over each row to ultimately determine the style... `isSelected` was added later as an inversion for `selectedRow` - `isSelected` can actually be used solo, just has to return a boolean - it's very useful for checkbox-selectable rows (there's an example in stripes-components storybook.) The default function for isSelected is a call to `isEqual`. In inventory's case, when an item was clicked, it was passed the entire object for `selectedRow` (passing `isEqual` ) if the page was merely routed to, the `selectedRow` object only contained the id - so it wouldn't pass an `isEqual` check since it lacked all the other keys. Good ol' `lodash` came through with the `isMatch` call to match an object containing a subset of properties.

PR #1080, opened 2021-06-22
STSMACOM-521 results selection bugfix

JohnC-80:
Resolves issue with all items styled as selected after closing the selected item.

## Problem
When the item's detail pane is dismissed, the selected item is currently set to an empty object (not undefined or null)
This empty object causes `isMatch` to return true (since both items are objects, it's considered a match.)

## Approach
Nullify selectedItem when the detail pane is dismissed. Adjust `isSelected` function to handle the null value (return false)


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -1141,12 +1141,17 @@ class SearchAndSort extends React.Component {
...
+                return isMatch(item, criteria);
+              }
+              return false;
+            }}
           />
         )}
       </FormattedMessage>
     );
-  }
+  }aaaaaa
```
mkuklis, line 1154: @JohnC-80 I like this :)

mkuklis (approved):

lib/SearchAndSort/SearchAndSort.js
JohnC-80: Sometimes I just have to get it all out.

PR #1093, opened 2021-07-12
STSMACOM-470: ViewCustomFieldsRecord must never return an empty element (<>)

mariiaschv:
## Description
`ViewCustomFieldsRecord` doesn't disappear even if no custom-fields are present.
If an accordion contains no content, the closed accordion shows a 0 badge.
If an accordion contains no content, the opened accordion shows a "No custom fields found" message.

## Screenshots
image
image

## Issues
[STSMACOM-470](https://issues.folio.org/browse/STSMACOM-470)
Also, change in ui-users was required ([UIU-2210](https://issues.folio.org/browse/UIU-2210))

### DIFF
--- a/lib/CustomFields/pages/ViewCustomFieldsRecord/ViewCustomFieldsRecord.js
+++ b/lib/CustomFields/pages/ViewCustomFieldsRecord/ViewCustomFieldsRecord.js
@@ -84,8 +84,6 @@ const ViewCustomFieldsRecord = ({
     sectionTitleFetchFailed,
   } = useSectionTitleFetch(okapi, backendModuleName.toUpperCase());
 
-  const CUSTOM_FIELDS = 'Custom fields';
-
   const { calloutRef } = useLoadingErrorCallout(customFieldsFetchFailed || sectionTitleFetchFailed);
   const columnWidth = rowShapes / columnCount;
 
@@ -96,7 +94,7 @@ const ViewCustomFieldsRecord = ({
     ? (<Badge>{visibleCustomFields.length}</Badge>)
     : (<Icon icon="spinner-ellipsis" width="10px" />);
 
-  const customFieldsAccordionTitle = sectionTitle.value || CUSTOM_FIELDS;
+  const customFieldsAccordionTitle = sectionTitle.value || <FormattedMessage id="ui-users.custom.customFields" />;
 
   const findOptionLabelById = customField => optionId => {
     return customField.selectField.options.values.find(option => option.id === optionId).value;
@@ -182,10 +180,7 @@ const ViewCustomFieldsRecord = ({
               }
             </Row>
           ))
-          : <FormattedMessage
-            id="ui-users.noItemFound"
-            values={{ item: CUSTOM_FIELDS.toLowerCase() }}
-          />
+          : <FormattedMessage id="ui-users.custom.noCustomFieldsFound" />
         }
       </Accordion>
       <Callout ref={calloutRef} />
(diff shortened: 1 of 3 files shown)


lib/CustomFields/pages/ViewCustomFieldsRecord/ViewCustomFieldsRecord.js
```diff
@@ -82,13 +84,20 @@ const ViewCustomFieldsRecord = ({
     sectionTitleFetchFailed,
   } = useSectionTitleFetch(okapi, backendModuleName.toUpperCase());
 
+  const CUSTOM_FIELDS = 'Custom fields';
```
Vladyslav-Velytskyi, line 87: Should it be added to translation file?

BogdanDenis (changes requested):

lib/CustomFields/pages/ViewCustomFieldsRecord/ViewCustomFieldsRecord.js
```diff
@@ -142,18 +149,19 @@ const ViewCustomFieldsRecord = ({
...
+        label={sectionTitleLoaded
+          ? customFieldsAccordionTitle
+          : <Icon data-test-custom-fields-loading-icon icon="spinner-ellipsis" />
+        }
+        displayWhenClosed={displayWhenClosed}
+        data-test-custom-fields-view-accordion
+      >
+        {customFieldsLoaded && formattedCustomFields.length
+          ? formattedCustomFields.map((row, i) => (
             <Row key={i}>
```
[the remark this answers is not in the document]
BogdanDenis: `formattedCustomFields` is an array of arrays. It represents array of rows which don't any properties that can be used as key, so we can ignore this here

Vladyslav-Velytskyi (approved):

lib/CustomFields/pages/ViewCustomFieldsRecord/ViewCustomFieldsRecord.js
```diff
@@ -59,6 +61,8 @@ const propTypes = {
 
 const defaultProps = {
   columnCount: 4,
+  customFieldsLabel: 'stripes-smart-components.customFields',
```
BogdanDenis, line 64: Usually in Folio labels that are passed as props are either `<FormattedMessage` components or already formatted strings. Passing and id of a translation might be confusing to users of component.
Please change default value to `<FormattedMessage id="stripes-smart-components.customFields" />` and simply render this prop

BogdanDenis (changes requested):

lib/CustomFields/pages/ViewCustomFieldsRecord/ViewCustomFieldsRecord.js
Vladyslav-Velytskyi: @BogdanDenis I'm not sure that it will be working, because `<FormattedMessage>` is the react component and we cannot use it in `defaultProps`, outside of the component

Vladyslav-Velytskyi: I think that better to rename this prop in `customFieldLabelId`. Does it make sense?

BogdanDenis: @Vladyslav-Velytskyi there's an example of `<FormattedMessage>` in default props - `<InfoPopover>` in `stripes-components`

BogdanDenis (approved):

PR #1099, opened 2021-07-23
Document canCreate prop of EditableList

skomorokh:
It is in fact possible to disable record creation on this component, but that functionality was undocumented... until now.

### DIFF
--- a/lib/EditableList/readme.md
+++ b/lib/EditableList/readme.md
@@ -31,6 +31,7 @@ Use the `EditableList` component in your jsx
 ### Configuration (props)
 Name | type | description | default | required
 --- | --- | --- | --- | ---
+canCreate | boolean | Determines if the '+ New' button is active. | | no
 contentData | array of objects | Array of objects to be rendered as list items. | | yes
 nameKey | string | The key that uniquely names listed objects: defaults to 'name'. | | no
 onCreate | function | Callback for creating new list items. | | no


lib/EditableList/readme.md
```diff
@@ -31,6 +31,7 @@ Use the `EditableList` component in your jsx
 ### Configuration (props)
 Name | type | description | default | required
 --- | --- | --- | --- | ---
+canCreate | boolean | Determines if the '+ New' button is active. | | no
```
MikeTaylor, line 34: What about setting the "default" column to `false`?

MikeTaylor (approved):
Thanks for this good-citzen moment.

lib/EditableList/readme.md
skomorokh: I'm pretty sure it's undefined if not present so only falsy not false and that column reflects that by being empty for many optional properties.

skomorokh: ...hm, but actually it's true now that I see it in the table. Another PR I guess....

MikeTaylor: Leave it how it is — your previous argument was convincing.

skomorokh: Except that I was confused while making that previous argument and it turns out to be explicitly defaulted to true so it was most honest to be explicit about that default and indeed consistent with another column further down documented as true. So I'd gone ahead with another PR to add that.

MikeTaylor: Yep, saw it, it's good.

PR #1114, opened 2021-08-19
UXPROD-3148: Adding logic for translating library-defined controlled vocabularies

attia-alshareef:
## Description
Each FOLIO installation has lists of controlled vocabulary terms (for instance, “Formats” and “Resource types” in Inventory, “Patron Groups” and “Refund Reasons” in Users — see FOLIO-2802 for a more complete list).  In installations that use more than one language, these terms do not display translated values when the locale of the session is changed.

## Use case(s)
More than one component has been monitored to control the library-defined controlled vocabularies, such as:
* ControlledVocab component.
* EntryManager component.

...This PR is for a ControlledVocab component, and more PRs for all the appropriate components will be added later.

## Workflow
For the complete translation workflow, see [Workflow](https://github.com/folio-org/ui-translations/blob/master/README.md#workflow)

## Jira Related Issues Links
https://issues.folio.org/browse/UXPROD-3148 <br />
https://issues.folio.org/browse/FOLIO-3258 <br />
https://issues.folio.org/browse/UXPROD-515 <br />
https://issues.folio.org/browse/FOLIO-2802

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -1,15 +1,27 @@
+/* eslint-disable react/sort-prop-types */
 import React from 'react';
 import PropTypes from 'prop-types';
-import { FormattedDate, FormattedMessage } from 'react-intl';
+import { FormattedDate, FormattedMessage, injectIntl } from 'react-intl';
 import { Link } from 'react-router-dom';
 import { isEqual, uniqueId, pickBy, noop } from 'lodash';
 
-import { Button, Callout, Col, ConfirmationModal, Modal, Pane, Paneset, Row, Loading } from '@folio/stripes-components';
+import {
+  Button,
+  Callout,
+  Col,
+  ConfirmationModal,
+  Modal,
+  Pane,
+  Paneset,
+  Row,
+  Loading
+} from '@folio/stripes-components';
 import SafeHTMLMessage from '@folio/react-intl-safe-html';
 
 import EditableList from '../EditableList';
 import css from './ControlledVocab.css';
 import makeRefdataActuatorsBoundTo from './actuators-refdata';
+import TranslationsActionsMenu from './TranslationsActionsMenu/TranslationsActionsMenu';
 
 class ControlledVocab extends React.Component {
   static manifest = Object.freeze({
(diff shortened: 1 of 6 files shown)


lib/ControlledVocab/ControlledVocab.js
```diff
@@ -51,22 +64,32 @@ class ControlledVocab extends React.Component {
...
-          },
+          }
         }
       }
     },
     updaterIds: {},
+    // Translated library-defined controlled vocabularies [UXPROD-3148]
+    translations: {
+      type: 'okapi',
+      path: 'translations',
```
julianladisch, line 75: I disagree with this proposed architecture.

The translations of controlled vocabulary should not be stored in a central translation module.
Instead the translations should live in the same back-end module where the non-translated controlled vocabulary lives.

This allows to have self-contained modules with as few inter-module dependencies as possible. In a micro-services-like architecture modules should be as independent as possible.

This matches the Technical Council's decision that back-end modules should handle i18n/l10n of messages: https://wiki.folio.org/display/DD/Handle+i18n+where+messages+are+generated

Does it have advantages that the translations of controlled vocabulary live in a different module?

attia-alshareef: There are many advantages that push us to save translations in a different module that I will explain in the translation app docs soon, but as a quick answer I will mention some of the risks and challenges that you will face if we are saving translations in the same back-end module:

1- It will require many changes for all back-and-modules that have controlled vocabulary. (Our solution does not require any changes at the level of back-and-modules)

2- Folio supports more than 26 languages ​​so far, which means that we are in the process of complicated multi-lingual schema and dramatic changes on the back-end architecture level are needed.

3- The amount of changes required on the stripes app to display the translated strings ​​instead of the original values ​​will be very large (our solution does not require this size of changes because all translations are fetched from one place and then merged with the default translations which are handled by react -intl )

About the Technical Council's decision that back-end modules should handle i18n/l10n of messages:
I think this is a different case because it is concerned with the error and validation messages generated by developers in the backend and I think they will be static translations (I don't know what is the approach the community follow-up to solve this issue).

zburke: This is an important discussion but I don't think this PR is the right place for it. 

@attia-alshareef you and your team have done some impressive work here but before we can start merging PRs, we need to make sure that the specific shape of the feature (l10n of runtime values) is acceptable to the Product Council and that its technical implementation passes muster with the Tech Leads (or Tech Council, if the folks at Tech Leads decide they need to boot it up a level for official approval). We are in the process of organizing that now. 

In any case, this UI work should not be merged until the other services that support it -- i.e. new backend modules, or changes to the interfaces of existing modules -- is already in place.

PR #1136, opened 2021-10-28
STSMACOM-546 lock to axe-core 4.3.3 to avoid test failures

zburke:
`axe-core` `4.3.4` causes test failures here on the master branch that `4.3.3` does not. 

From https://github.com/dequelabs/axe-core/issues/3241: 
>  it seems we released something from develop that shouldn't have been there yet. 

We should look at this again in a few weeks and hope to revert this PR. Locking directly onto a patch version is not great. 

Refs [STSMACOM-546](https://issues.folio.org/browse/STSMACOM-546)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add disabled prop for `<LocationLookup>` component.
 * Config Manager | Apply baseline keyboard shortcuts. Refs STSMACOM-544.
+* Lock to `axe-core` `4.3.3`; `4.3.4` causes test failures in CI/CD.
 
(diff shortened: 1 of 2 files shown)


WilcoFiers:
@zburke Axe-core 4.3.5 was released, in which we've disabled the test for using non-global ARIA attributes on elements with no role. This will be reintroduced in 4.4.0, just to give a heads up for that. But it wasn't meant to go into a patch.

zburke:
Thanks for the heads-up, @WilcoFiers. It was extra work of no benefit to you to comment here and explain the situation, but I really appreciate that you took the time to do it. Class act!

PR #1139, opened 2021-11-08
Change next release version from 7.0.1 to 7.1.0 to avoid confusion

BogdanDenis:
## Description
Change future release version to `7.1.0` so that no one thinks that these changes require release for Kiwi Bugfix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change history for stripes-smart-components
 
-## [7.0.1] (IN PROGRESS)
+## [7.1.0] (IN PROGRESS)
 
 * Add disabled prop for `<LocationLookup>` component.
 * Config Manager | Apply baseline keyboard shortcuts. Refs STSMACOM-544.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/stripes-smart-components",
-  "version": "7.0.0",
+  "version": "7.1.0",
   "description": "Connected Stripes components",
   "repository": "folio-org/stripes-smart-components",
   "sideEffects": [


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Change history for stripes-smart-components
 
-## [7.0.1] (IN PROGRESS)
+## [7.1.0] (IN PROGRESS)
```
AndreiShumski2, line 3: IMO it's better to remove the version at all - `## (IN PROGRESS)`. And put release version only during new release creation.

zburke (commented):
I _do_ need to do for _all_ the stripes-* repos, but it needs to happen in order from the bottom of the hierarchy (stripes-components) up to the top (stripes). Sorry for the delay on that. I'll try to get that done, including this PR, Monday or Tuesday.

BogdanDenis:
> I _do_ need to do for _all_ the stripes-* repos, but it needs to happen in order from the bottom of the hierarchy (stripes-components) up to the top (stripes). Sorry for the delay on that. I'll try to get that done, including this PR, Monday or Tuesday.

Thanks, Zak
While we're on topic of module versions - I think we should also increment `eslint-config-stripes` to `6.1.0`. I saw that you reverted changes with webpack upgrade and with that revert the version went from `6.1.0` to `6.0.0`. But the `latest` still points to `6.1.0` which causes errors in GitHub actions tests

zburke:
Oooof, thanks for bringing the bad eslint-config-stripes releases to my attention. I think I got that sorted out (at least for npm-folioci; I still need to do the same for npm-folio), but some tests are still failing here and in stripes-core. The fun never stops! I'll merge this as soon as I figure out why these tests started dying.

zburke:
@BogdanDenis , at this point I think this PR is blocked on broken tests (which @NikitaSedyx has fixed as part of #1140), not on `eslint-config-stripes` problems.

CHANGELOG.md
NikitaSedyx: in acq we keep it's always in progress at the top and add release version when do releases

https://github.com/folio-org/ui-orders/blob/master/CHANGELOG.md#in-progress

NikitaSedyx (approved):

NikitaSedyx:
@BogdanDenis test fixes are merged so your request can be unblocked

PR #1141, opened 2021-11-15
STSMACOM-541 restore react 17-related focus tests

ryandberger:
- Fulfills [STSMACOM-541](https://issues.folio.org/browse/STSMACOM-541)
- Updated interactors to use new `text-field` interactor in `stripes-testing` which uses `focusout` instead of `blur` to be compatible with React 17.
- Also made use of `dispatchFocusout` from `stripes-testing` for other element types outside of `text-field`.


JohnC-80:
@zburke `.has` is an assertion built into the new interactors... they throw if they don't succeed in finding the element matching the criteria. They are meant to be used this way and replace other types of assertion/dependencies.

In cases where the assertion isn't DOM-rendered content, `converge(fn())` can be used - which will throw if `fn` doesn't return `true` within a timeout. Will make note to document this if it isn't already...

mkuklis (approved):
Great work @ryandberger!

zburke (approved):
Thanks for the explanation of `.has()`, @JohnC-80. That makes perfect sense, but now we have a single file that uses two different conventions for assertions. Am I just being uselessly ornery here? It feels like half the file is like 
```
if (kAndRForever) {
}
``` 
and half is like 
```
if (javaGonnaHateHateHateHateHateHateHateHateHate)
{
}
```
It's not wrong, but it's not ideal. 

Anyway, no objections to this PR that objectively improves things by increasing coverage, just something to think about as we refactor other BTOG interactors and tests.

JohnC-80:
@zburke I agree that we do ultimately want consistency. I believe we will soon achieve that in future work - we're soon slated to get smart-components upgraded to using new bigtest interactors.

JohnC-80 (approved):

PR #1146, opened 2021-12-02
STSMACOM-613: Use <PersistedPaneset> in <SearchAndSort>

Baroquem:
https://issues.folio.org/browse/STSMACOM-613

This originated as a request to use `<PersistedPaneset>` to remember the resized widths of the three top-level panes in ui-inventory and to track whether the filter pane should be visible or not. Since those elements are defined in `<SearchAndSort>`, the work was done there.


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -45,6 +44,7 @@ import {
   AppIcon,
   IntlConsumer,
 } from '@folio/stripes-core';
+import { PersistedPaneset } from '@folio/stripes-smart-components';
```
zburke, line 47: This needs to be `from '../../PersistedPaneset'`; the current package already is `@folio/stripes-smart-components`.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -250,9 +250,13 @@ class SearchAndSort extends React.Component {
       stripes,
     } = this.props;
 
+    // Use local storage to remember whether the filter pane has been toggled off or not.
+    this.filterPaneVisibilityKey = '@folio/inventory/inventoryFilterPaneVisibility';
```
mkuklis, line 254: @Baroquem I think this will have to be different for each ui module right?
I wonder if we could use a `withNamespace` HOC from stripes-core: 

https://github.com/folio-org/stripes-core/blob/0fcafe70b8a25525ef488ab6e7e800c148799e4d/src/components/Namespace/withNamespace.js#L9 

to get a unique key for each module.

You could try to wrap the entire `SearchAndSort` with `withNamespace` and you should get a `namespace` prop as a string.

zburke, line 254: I don't think the `@folio/inventory/...` key belongs at this level. Can this be configured dynamically in a similar way to how `appId` is constructed below?

mkuklis (approved):
Great work @Baroquem! I left one comment which I think we will need to address before this can be merged.

zburke (changes requested):
There are some config glitches here that need to be resolved, but this looks promising.

lib/SearchAndSort/SearchAndSort.js
Baroquem: Yes, of course you're both right -- this shouldn't be hard-coded here. Thanks for the `withNamepsace` tip, Michal; that's exactly what I needed!

Baroquem: Yes, you're right. This should look better now.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -132,6 +133,7 @@ class SearchAndSort extends React.Component {
     module: PropTypes.shape({ // values specified by the ModulesContext
       displayName: PropTypes.node, // human-readable
     }),
+    namespace: PropTypes.string.isRequired,
```
zburke, line 136: A new required prop is a breaking change. Do you think there's a way to handle this dynamically?

mkuklis: @zburke I don't think this is a breaking change because the consumers don't have to pass it down or worry about it. It's handled by the SAS internally.

zburke: 🤦 `withNamespace`. Thanks for setting me straight, @mkuklis.

zburke (approved):
This has arrived in a really nice state at last! Great work!

PR #1160, opened 2021-12-26
STSMACOM-617 Settings : Move focus to second pane

vashjs:
https://issues.folio.org/browse/STSMACOM-617

## Problem
In the current implementation, when you click on the link in the "settings" panel, the focus does not switch to the newly opened one. 

current

## Solution
- Add an onClick prop that makes the element focusable
- When rendering a block, focus on the first element of the first section. 

fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Update internal state in SASQ when sort changes. Fixes STSMACOM-614.
 * Add preferred name to the Proxy Modal. Fixes STSMACOM-615.
 * Lint
+* Settings : Move focus to second pane. Refs STSMACOM-617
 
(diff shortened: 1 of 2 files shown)


JohnC-80:
`autofocus` tends to get leaned on often for these focus-management cases... it works okay for this, but the gotchas come with any subsequent items having an `autofocus` attribute applied and the case where focus jumps prematurely due to the first link being present on initial render (this could happen if the settings page is directly linked to)
Also - what if the subsequent page doesn't contain a nav list?

Stripes-components exports some utilities that can be useful for obtaining focusable elements: 
https://github.com/folio-org/stripes-components/blob/master/util/README.md#getfocusableelements

We could probably use it to send move focus in a more generic way.

JohnC-80 (approved):

vashjs:
https://github.com/folio-org/stripes-components/pull/1712 created instead

PR #1163, opened 2021-12-30
STSMACOM-610 UserName interactor

ryandberger:
- Added `UserName` interactor and tests.
- Small spelling correction to `PasswordValidationFieldInteractor`

Refs [STSMACOM-610](https://issues.folio.org/browse/STSMACOM-610)

### DIFF
--- a/lib/UserName/tests/UserName-test.js
+++ b/lib/UserName/tests/UserName-test.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { describe, beforeEach, it } from '@bigtest/mocha';
+import { describe, beforeEach, it } from 'mocha';
 import { expect } from 'chai';
 
 import {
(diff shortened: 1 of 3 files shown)


lib/UserName/tests/UserName-test.js
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import { describe, beforeEach, it } from '@bigtest/mocha';
```
zburke, line 2: Can we import directly from `mocha` instead?

lib/UserName/tests/interactor.js
```diff
@@ -0,0 +1,6 @@
+import { interactor } from '@bigtest/interactor';
+import { Label as LabelInteractor } from '@folio/stripes-testing';
+
+export default interactor(class UserNameInteractor {
+  Label = new LabelInteractor();
```
zburke, line 5: `Label` or `label`? I think the latter since this is effectively an instance property, and that seems to be the convention in other interactors in this module.

ryandberger: Agree, going with `label`.

zburke (approved):

PR #1164, opened 2022-01-03
STSMACOM-619 correctly specify LayoutHeader props

zburke:
Use the correct shape for the objects in the `actions` prop for
`LayoutHeader`. Discovered while reviewing _entirely_ unrelated changes
in #1163.

Refs [STSMACOM-619](https://issues.folio.org/browse/STSMACOM-619), [STCOM-914](https://issues.folio.org/browse/STCOM-914)


ryandberger (approved):

zburke:
This bug is not a bug. https://github.com/folio-org/stripes-components/pull/1697 is a bug and https://github.com/folio-org/stripes-components/pull/1699 fixes it.

PR #1166, opened 2022-01-05
Upgrade entry manager tests

JohnC-80:
Upgrading these to new interactors exported from `@folio/stripes-testing`.

I can imagine that we could find a way to *still make an entry manager interactor work for those entry manager that are rolled from scratch at the ui-module level...
A nav list, a button, callout are mostly the built-in interactors that are needed, along with some special ones to cover your `[data-test-whatever]` (or possibly better) containing elements.

Require a merge request on stripes-testing.


zburke (approved):
If you

PR #1168, opened 2022-01-11
STSMACOM-620: Add additional prop in SearchAndSort

UladzislauKutarkin:
https://issues.folio.org/browse/STSMACOM-620

*Problem*

We need to have the ability to change paneSubTitle when props change and change the text in the button in SearchFiled. All that needs to be done in the scope of the Browse call numbers feature.

*Solution*

Simple add additional props in SearchAndSort components.

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -90,10 +90,7 @@ class SearchAndSort extends React.Component {
     columnWidths: PropTypes.object,
     createRecordPath: PropTypes.string,
     customPaneSub: PropTypes.node,
-    customPaneSubText: PropTypes.oneOfType([
-      PropTypes.string,
-      PropTypes.element
-    ]),
+    customPaneSubText: PropTypes.node,
     detailProps: PropTypes.object,
     disableFilters: PropTypes.object,
     disableRecordCreation: PropTypes.bool,
@@ -215,7 +212,7 @@ class SearchAndSort extends React.Component {
     resultsVirtualize: PropTypes.bool,
     searchableIndexes: PropTypes.arrayOf(PropTypes.object),
     searchableIndexesPlaceholder: PropTypes.string,
-    searchFieldButtonLabel: PropTypes.string,
+    searchFieldButtonLabel: PropTypes.node,
     selectedIndex: PropTypes.string, // whether to auto-show the details record when a search returns a single row
     showSingleResult: PropTypes.bool,
     sortableColumns: PropTypes.arrayOf(PropTypes.string),
@@ -1064,7 +1061,7 @@ class SearchAndSort extends React.Component {
           disabled={!searchTerm}
           data-test-search-and-sort-submit
         >
-          <FormattedMessage id={searchFieldButtonLabel || 'stripes-smart-components.search'} />
+          {searchFieldButtonLabel || <FormattedMessage id="stripes-smart-components.search" />}
         </Button>
         {this.renderResetButton()}
         {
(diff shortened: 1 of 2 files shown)


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -90,6 +90,7 @@ class SearchAndSort extends React.Component {
     columnWidths: PropTypes.object,
     createRecordPath: PropTypes.string,
     customPaneSub: PropTypes.node,
+    customPaneSubText: PropTypes.string,
```
zburke, line 93: Are you 100% confident this should be `string`? To accept a `<FormattedMessage>` I would have expected `node`, or perhaps `oneOfType([PropTypes.string, PropTypes.element])`.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -211,6 +212,7 @@ class SearchAndSort extends React.Component {
     resultsVirtualize: PropTypes.bool,
     searchableIndexes: PropTypes.arrayOf(PropTypes.object),
     searchableIndexesPlaceholder: PropTypes.string,
+    searchFiledButtonLabel: PropTypes.string,
```
zburke, line 215: Typo, I think: `s/Filed/Field/`.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -1058,7 +1060,7 @@ class SearchAndSort extends React.Component {
           disabled={!searchTerm}
           data-test-search-and-sort-submit
         >
-          <FormattedMessage id="stripes-smart-components.search" />
+          <FormattedMessage id={this.props.searchFiledButtonLabel || 'stripes-smart-components.search'} />
```
zburke, line 1063: All other `this.props` values are destructured above beginning on line 1008/1010. Please follow the established convention. And as above, fix the typo: `s/Filed/Field/`.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -1184,6 +1187,9 @@ class SearchAndSort extends React.Component {
             <FormattedMessage id={messageKey} values={{ count: source.totalCount() }} />
           )}
           {!isSourceLoaded && (
+            customPaneSubText?
```
zburke, line 1190: Run lint before pushing up your branch to avoid these kinds of errors.

zburke (changes requested):

lib/SearchAndSort/SearchAndSort.js
UladzislauKutarkin: What do you mean` s/Filed/Field/`?

UladzislauKutarkin: Thanks a lot, my little mistake

lib/SearchAndSort/SearchAndSort.js
UladzislauKutarkin: Done.

lib/SearchAndSort/SearchAndSort.js
UladzislauKutarkin: I add oneOfType.

UladzislauKutarkin: You are right, It's better to use node to accept a `<FormattedMessage>`. So, I change a logic a little bit and now we can use node.

lib/SearchAndSort/SearchAndSort.js
UladzislauKutarkin: Done.

JohnC-80 (approved):
Looks good. Thanks for addressing Zak's changes.

zburke (approved):

PR #1169, opened 2022-01-11
STSMACOM-621 Enable accessible result lists.

JohnC-80:
Problem: built-in result lists have many `axe` errors due to the block anchors that SearchAndSort renders. Changing ui-modules to render rows requires a design step of rendering the detail record link at the cell level of results - this means that certain built-ins of SearchAndSort need to be suppressed or inverted so that apps can supply some of their own logic.

With this comes some added props.
- resultsRowFormatter - enables replacement of SearchAndSort's built-in row formatter.
- hasRowClickHandlers - boolean to suppress row-level click handling.
-  resultRowIsSelected - function to determine whether or not selected styles are applied to a result item.
-  onDismissDetail - allows module-level control of the callback for dismissing the detail pane.

Refs [STSMACOM-621](https://issues.folio.org/browse/STSMACOM-621), [STCOM-904](https://issues.folio.org/browse/STCOM-904)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Update internal state in SASQ when sort changes. Fixes STSMACOM-614.
 * Add preferred name to the Proxy Modal. Fixes STSMACOM-615.
 * Lint
+* Introduce `useCustomFields` hook. Refs STSMACOM-622.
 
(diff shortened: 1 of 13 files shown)


zburke (approved):
I feel like the new props here are not consistently named, but the old props certainly aren't either. Maybe at some point we should rename them all to a bunch of consistent things, and we can deprecate the old ones in a future breaking release.

PR #1173, opened 2022-01-20
STSMACOM-612 Notes > Handle /note-links response when there are no notes

BogdanDenis:
## Description
Handle a case when `notes` property in `note-links/` response is undefined when there are no assigned notes

## Issues
[STSMACOM-612](https://issues.folio.org/browse/STSMACOM-612)

### DIFF
--- a/lib/Notes/NotesSmartAccordion/tests/notes-smart-accordion-test.js
+++ b/lib/Notes/NotesSmartAccordion/tests/notes-smart-accordion-test.js
@@ -15,25 +15,27 @@ describe('Notes smart accordion', () => {
     scenarios: ['create-notes-with-note-types']
   });
 
-  beforeEach(async () => {
-    const domainName = 'dummy';
+  const domainName = 'dummy';
 
+  const renderPage = () => mount(
+    <div style={{ maxWidth: '500px', marginLeft: '20px' }}>
+      <NotesSmartAccordion
+        id="notes-accordion"
+        open
+        onToggle={() => {}}
+        domainName={domainName}
+        entityName="Cool Provider"
+        entityType="provider"
+        entityId="providerId"
+        pathToNoteCreate={`${domainName}/notes/new`}
+        pathToNoteDetails={`${domainName}/notes`}
+      />
+    </div>
+  );
+
+  beforeEach(async () => {
     // Provider is an example of entity. Any another entity type can be used. e.g. User, Agreement...
-    await mount(
-      <div style={{ maxWidth: '500px', marginLeft: '20px' }}>
-        <NotesSmartAccordion
-          id="notes-accordion"
-          open
-          onToggle={() => {}}
-          domainName={domainName}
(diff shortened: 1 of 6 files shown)


lib/Notes/NotesSmartAccordion/tests/notes-smart-accordion-test.js
```diff
@@ -15,25 +15,27 @@ describe('Notes smart accordion', () => {
...
+        entityType="provider"
+        entityId="providerId"
+        pathToNoteCreate={`${domainName}/notes/new`}
+        pathToNoteDetails={`${domainName}/notes`}
+      />
+    </div>
+  );
 
+  beforeEach(async () => {
     // Provider is an example of entity. Any another entity type can be used. e.g. User, Agreement...
```
mariiaschv, line 37: Maybe it's better to move this comment to be above `renderPage()`?

mariiaschv (approved):

zburke (changes requested):
The work here is solid, but I think it's the wrong work because I do not agree with the removal of the `notes` property in the API response, hence I reject this change on general principle. 

1. Changing the shape of a response in this manner should be be a breaking change in the interface version but there is no such change here. In order to avoid making this a breaking change to stripes which would require changes to _all_ UI repositories, we would have to implement this change by supporting both the current shape (`notes` is present) and the new shape (`notes` may not be present) so that both "`notes` is absent" and "`notes` is present but empty" would be handled the same way. 
2. Changing the shape of a response in this manner is consistent with _every other API that returns a collection_ in FOLIO. While this change may make sense in the limited scope of notes, it does not make sense in the larger ecosystem of FOLIO APIs where endpoints that return single records return 404s if no record matches the ID and endpoints that return lists return empty lists if there are no records that match the query.

Svitlana-Korol (approved):

BogdanDenis:
> The work here is solid, but I think it's the wrong work because I do not agree with the removal of the `notes` property in the API response, hence I reject this change on general principle.
> 
> 1. Changing the shape of a response in this manner should be be a breaking change in the interface version but there is no such change here. In order to avoid making this a breaking change to stripes which would require changes to _all_ UI repositories, we would have to implement this change by supporting both the current shape (`notes` is present) and the new shape (`notes` may not be present) so that both "`notes` is absent" and "`notes` is present but empty" would be handled the same way.
> 2. Changing the shape of a response in this manner is consistent with _every other API that returns a collection_ in FOLIO. While this change may make sense in the limited scope of notes, it does not make sense in the larger ecosystem of FOLIO APIs where endpoints that return single records return 404s if no record matches the ID and endpoints that return lists return empty lists if there are no records that match the query.

Thanks for explaining your points clearly. Unfortunately I don't have details as to why we're removing `notes` property if the collection is empty. @psmagin could you provide more details on this point please?

BogdanDenis:
Discussed with @psmagin, this work won't be needed, closing the PR

PR #1178, opened 2022-02-02
STSMACOM-549 UI-Users bombs out when adding new loan owner.

JohnC-80:
### Pro'lem: 
If no initial values were provided, the lifecycle for syncing items would bomb out since `isEqual` doesn't do well at handling `undefined` parameters (just returns false in those cases) and so the underlying logic would attempt to access a property of the undefined value.

### Refresher: 
`EditableList` uses an internal array to track the editing state of items that has to be kept in sync with its props.

### Approach:
Sure, we could have just done a check for the existence of `props.initialValues` and called it a day... but no, them UNSAFE lifecycles gotta go. Moved the `buildStatusArray` function outside of the component, set up the old gDSFP function to handle async initial values, and made sure to sync things appropriately when items were edited, re-building the status array where necessary.

2 birds with one stone: fix the bug, remove a deprecated lifecycle.

Tests look good, interaction seems good. Hopefully no other stones show up....

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Lint
 * Introduce `useCustomFields` hook. Refs STSMACOM-622.
 * Add `paneTitleRef` prop to the `Settings` component. Refs STSMACOM-623.
+* Fix issue with EditableList crashing when a new item is added. Fixes STSMACOM-549.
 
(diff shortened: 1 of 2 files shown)


ryandberger (approved):
Awesome! Nice work knocking out the old lifecycle with this fix!

PR #1188, opened 2022-02-18
STSMACOM-609 refactor ViewMetaData tests

ryandberger:
- Updated `ViewMetaData-test` and `NoteViewPage-test`


zburke (approved):
OMG, 25 commits for changes to 3 files 😆 😆 😱 . It's so nice to see passing tests sailing through CI here! The only question I have is why only some of the `it` functions are `async`. I did not notice an obvious pattern and it would be nice to if/when that's truly necessary.

ryandberger:
> OMG, 25 commits for changes to 3 files 😆 😆 😱 . It's so nice to see passing tests sailing through CI here! The only question I have is why only some of the `it` functions are `async`. I did not notice an obvious pattern and it would be nice to if/when that's truly necessary.

Good questions! It's been awhile now, and I assume I added that to make sure there are no timing issues, but if only some of them have `async` it might actually be unnecessary! I'll try removing and verify if they can go without `async`.

JohnC-80:
@zburke That's kindof how experimentation goes - especially when everything passes in local dev, but not in CI ,haha - trying every little thing to make it work - pointing at branches from other repos, tweaking all the little things. 
I might have been responsible for some of those async calls - part of some of the feedback from frontside a while ago when we first had these problems was that interactions and assertions should both be async.
I don't believe awaiting assertions is always necessary - but it's probably one thing to *try if false positives are happening.

zburke:
I'm glad we got this merged! Please choose the "Squash and merge" option in the future. The general rules for merging UI PRs are:
* Never force-push to a branch; that destroys the history on the PR and can making following the conversation really hard, especially if comments refer to commits that no longer exist. 
* Always squash when you merge; this keeps the history tidy on the main branch.
* Always put the Jira slug first on the commit subject so it's easy to see when browsing through history. GitHub truncates long subjects, and "UIU-123 restores last name of user in ..." is more helpful than "restores last name of user in details ..."

PR #1195, opened 2022-02-22
STSMACOM-636 Add aria-label for checkboxes in `<ChangeDueDateDialog>` component

OleksandrHladchenko1:
In scope of UIU-1686 it is necessary to add aria-label for checkboxes in `<ChangeDueDateDialog>` component.

* #checkbox-468 (in users > view > loans > open loans > loan detail > change due date modal)

Refs: https://issues.folio.org/browse/UIU-1686


lib/ChangeDueDateDialog/LoanList.js
```diff
@@ -76,6 +76,7 @@ const LoanList = (props) => {
       columnMapping={{
         selected: <Checkbox
           name="selected-all"
+          aria-label="Select all"
```
zburke, line 79: `Select all` needs to be a `<FormattedMessage />` instead of a hard-coded string.

lib/ChangeDueDateDialog/LoanList.js
```diff
@@ -91,6 +92,7 @@ const LoanList = (props) => {
       formatter={{
         selected: loan => <Checkbox
           name={`selected-${loan.id}`}
+          aria-label={`select-${loan.id}`}
```
zburke, line 95: Please talk to the PO and find out of `loan.id` is the best option here. Reading out a UUID won't be great; we might need a function/translation that concatenates instance-title, barcode, due-date, etc. Note that `aria-label` needs to be a string, not a component, so you'll need `useIntl` and `formatMessage()`.

zburke (changes requested):
Thanks for picking up this a11y story! We need to make sure it does i18n correctly too. This is a great start!

zburke (approved):
LGTM. Thank you!

PR #1198, opened 2022-02-23
STSMACOM-636: Add aria-label for checkboxes in `<ChangeDueDateDialog>` component

OleksandrHladchenko1:
In scope of UIU-1686 it is necessary to add `aria-label` for checkboxes in `<ChangeDueDateDialog>` component.

* #checkbox-468 (in users > view > loans > open loans > loan detail > change due date modal)

Refs: https://issues.folio.org/browse/UIU-1686

Previous PR: https://github.com/folio-org/stripes-smart-components/pull/1195

### DIFF
--- a/lib/ChangeDueDateDialog/LoanList.js
+++ b/lib/ChangeDueDateDialog/LoanList.js
@@ -54,14 +54,14 @@ const LoanList = (props) => {
   const { formatMessage } = useIntl();
 
   const visibleColumns = [
-    'alertDetails',
-    'title',
-    'itemStatus',
-    'currentDueDate',
-    'requests',
-    'barcode',
-    'effectiveCallNumber',
-    'loanPolicy',
+    'loanListAlertDetails',
+    'loanListTitle',
+    'loanListItemStatus',
+    'loanListCurrentDueDate',
+    'loanListRequests',
+    'loanListBarcode',
+    'loanListEffectiveCallNumber',
+    'loanListLoanPolicy',
   ];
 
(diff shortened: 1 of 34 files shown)


lib/ChangeDueDateDialog/LoanList.js
```diff
@@ -66,6 +69,12 @@ const LoanList = (props) => {
 
   const rowUpdater = (rowData) => requestCounts[rowData.itemId];
 
+  const setCheckboxLabel = (id) => {
+    const currentLoan = loans.find(loan => loan.id === id).item;
+
+    return `checkbox-${kebabCase(currentLoan.title)}-${currentLoan.barcode}`;
```
zburke, line 75: Since this will be read aloud by a screen reader, we shouldn't hard-code "checkbox" in there. @JohnC-80, is there any particular guidance about including an `input` element's type in its `aria-label` value? It seems to me that either we can omit "checkbox" (and likely the dashes, and kebabCase too; this is a label to be read aloud to a human; not an identifier to be used for a computer), or we at least need to pull it through react-intl and `formatMessage()`.

zburke (changes requested):
The intention here is good but the implementation is not quite right.

lib/ChangeDueDateDialog/LoanList.js
mariia-aloshyna: Hi @zburke Thank you for the review! Updated the PR. Could you take a look, please?

zburke (approved):
Thank you!

PR #1199, opened 2022-02-28
STSMACOM-633 Translate `Actions` column header for EditableList

JohnC-80:
Refs [STSMACOM-633](https://issues.folio.org/browse/STSMACOM-633)


mkuklis (approved):

zburke (changes requested):
lint, then it's good to go!

zburke (approved):

PR #1200, opened 2022-03-04
STSMACOM-637: Do not push to history if the url did not change

mkuklis:
https://issues.folio.org/browse/STSMACOM-637

I noticed this bug when working on lazy loading. The ui-courses would just go into an infinite loop and keep re-rendering and reloading all resources. This happens because `SearchAndSortQuery` execs `history.push` for the same url as the current location url.


zburke (approved):
Sweet job debugging this!

JohnC-80 (approved):

Baroquem (approved):

PR #1201, opened 2022-03-08
STSMACOM-638 tests must not inspect NoValue's DOM structure

zburke:
How a component renders in the DOM is considered private and thus is
subject to change without warning. Nonetheless, some tests inspected
the DOM to assess whether a `<NoValue>` component had been rendered, and
when its internal state changed these tests began to fail.

Assess the presence of `<NoValue>` via its interactor.

Refs [STSMACOM-638](https://issues.folio.org/browse/STSMACOM-638), [STCOM-936](https://issues.folio.org/browse/STCOM-936)


mkuklis (approved):
❤️

PR #1203, opened 2022-03-10
STSMACOM-639 correctly provide default sortby value

zburke:
As noted by helpful comments from @miketaylor in this file, react-redux
does not receive defaultProps values, and hence can't pass them onto
stripes-connect for substitution in a manifest. T

The situation we ended up in was that our default value was being
ignored, causing evaluation of the manifest to fail and thus the query
would never fire.

CC: @andrewish

Refs [STSMACOM-639](https://issues.folio.org/browse/STSMACOM-639)


mkuklis (approved):

JohnC-80 (approved):

MikeTaylor (approved):
I was going to suggest that we also update the documentation to mention that `sortby` now has a default value ... until I realised there is no documentation. Clearly this should be merged as-is, but perhaps it's worth filing an issue to document `<ControlledVocab>`?

Anyway, proceed! Merge like the wind!

PR #1204, opened 2022-03-15
STSMACOM-640 Applying a date range filter clears other filters

usavkov-epam:
## Purpose:
https://issues.folio.org/browse/STSMACOM-640

`DateRangeFilter` component uses `React.memo` and check equality of only `selectedValues` prop. But we are passing a `onChange` callback to the `DateRangeFilter` component, which may depend on other parameters and, accordingly, the reference to the actual `onChange` callback may change. As a result, when the `applyFilter` function is called, `onChange` is used from the previous render, it causes problems in some applications.

## Approach:

- Check `onChange` prop equality on `React.memo` comparison function;
- Replace `applyFilter` function from `useEffect` hook to prevent infinite rerenders;


zburke (approved):
LGTM! Thanks for the clear explanation of the problem in the description. 

Please squash when you merge to keep the commit history linear.

NikitaSedyx (approved):

AndreiShumski2 (approved):

PR #1220, opened 2022-04-14
Add pane id to ControlledVocab, Settings components

JohnC-80:
https://issues.folio.org/browse/STSMACOM-652


lib/ControlledVocab/ControlledVocab.js
```diff
@@ -431,6 +431,7 @@ class ControlledVocab extends React.Component {
           paneTitle={this.props.label}
           paneTitleRef={this.paneTitleRef}
           onMount={this.handlePaneFocus}
+          id="controlled-vocab-pane"
```
zburke, line 434: Given this is the only `<Pane>` in this `<Paneset>` why do we need an `id` here? Is it just "better safe than sorry?" I'm fine with that; just trying to understand.

zburke (approved):

PR #1227, opened 2022-05-11
STSMACOM-519: EditableList: `actions` column title should be capitalized and translated

Godlevskyi:
# Purpose 
Added translation for 'Actions' column.

# Refs
[STSMACOM-519](https://issues.folio.org/browse/STSMACOM-519)

# Screenshot

Снимок экрана 2022-05-11 в 11 18 02

### DIFF
--- a/lib/EditableList/EditableListForm.js
+++ b/lib/EditableList/EditableListForm.js
@@ -199,6 +199,7 @@ class EditableListForm extends React.Component {
     this.getColumnWidths = this.getColumnWidths.bind(this);
     this.getVisibleColumns = this.getVisibleColumns.bind(this);
     this.getReadOnlyColumns = this.getReadOnlyColumns.bind(this);
+    this.getColumnMapping = this.getColumnMapping.bind(this);
 
     if (this.props.id) {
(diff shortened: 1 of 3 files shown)


lib/EditableList/EditableListForm.js
```diff
@@ -374,6 +375,19 @@ class EditableListForm extends React.Component {
...
+    const {
+      editable,
+      columnMapping,
+    } = this.props;
+
+    if (editable) {
+      columnMapping.actions = <FormattedMessage id="stripes-smart-components.actions" />;
+    }
+
+    return columnMapping;
```
zburke, line 388: pedantic/bikeshedding: I generally like to treat `props` as immutable (I know `props.columnMapping` is an object reference, and we're not changing that ref, so it technically isn't being mutated, but still...). Strictly adhering to that, you could do something like: 
```
const map = {...columnMapping};
if (editable) {
  map.actions = <FormattedMessage id="stripes-smart-components.actions" />;
}

return map;
```

Godlevskyi: Thanks, @zburke, corrected.

PR #1229, opened 2022-05-13
STSMACOM-662 Add optional `persist` prop to <ColumnManager>

MikeTaylor:
If present and true, this persists the selection of columns into subsequent sessions.

Fixes STSMACOM-662.

(I can't wait around for the Grand Unified Theory in STCOR-601, I need to get this done ASAP and with minimal disruption.)


MikeTaylor:
Can we please get this merged before there's another CI-free translations update, and I have to once more go round the dance where I have to go round and round re-starting the tests until I roll a six and they all pass?

PR #1231, opened 2022-05-16
STSMACOM-663: Search option dropdown resets value on next pick

ODanyor:
## Description
All next picks of search option reset it's value to the initial option value(Keyword (title, contributor, identifier))

## Approach
Resets _locallyChangedQueryIndex_ on tab switches if _qindex_ is not provided by url otherwise acts as usual.

## Issues
https://issues.folio.org/browse/STSMACOM-663


lib/SearchAndSort/SearchAndSort.js
```diff
@@ -353,9 +357,15 @@ class SearchAndSort extends React.Component {
...
-    }
+    const resetLocalQIndex = () => {
+      if (this.state.locallyChangedQueryIndex && !qindex) {
+        this.setState({ locallyChangedQueryIndex: '' });
+      }
+    };
+
+    if (currSegment && nextSegment) {
+      if (currSegment !== nextSegment) resetLocalQIndex();
+    } else resetLocalQIndex();
```
mkuklis, line 368: I think we can remove the `else` here. With the `else` in place this was still not working for me correctly.

Steps to reproduce:

1. Go to inventory
2. Choose `Contributor`
3. Choose `Subject`

After choosing `Subject` the drop down would just go back to `Keyword ...`.

mkuklis (commented):
@ODanyor this was a good find and a good approach. Nice work! I left couple comments.

lib/SearchAndSort/SearchAndSort.js
ODanyor: Sorry, my fault. I think just need to replace `&&` with `||`. Yeah, but the logic here is to make it work as usual in case of if there is no `segment` param usage. And `else` is responsible for that here.

And I suppose, combining these two conditions will not make it work as expected. Since the idea is, do nothing if there is a `segment` param and it's not changed yet.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -353,9 +357,15 @@ class SearchAndSort extends React.Component {
 
     // Reset local qindex if the qindex is not present in the url.
     // This happens when the search segment changes.
-    if (this.state.locallyChangedQueryIndex && !qindex) {
-      this.setState({ locallyChangedQueryIndex: '' });
-    }
+    const resetLocalQIndex = () => {
```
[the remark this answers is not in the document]
ODanyor: > @ODanyor this was a good find and a good approach. Nice work! I left couple comments.

Yeah,  @BogdanDenis made a little investigation and found the issue line)

lib/SearchAndSort/SearchAndSort.js
ODanyor: Yes, we can leave only this condition and cut the rest, It will perfectly work. But can we consider the cases with no segment value in url? If it's possible, of course

mkuklis: @ODanyor the previous implementation was only concerned about reseting it when the segment was changed. If there is no change I don't think we want to do anything right? Have you tried that with the `else`? It did not work for me as expected.

ODanyor: Ok, in this case I will leave only `if (currSegment !== nextSegment)`. It also works for me

mkuklis (approved):
👏

PR #1232, opened 2022-05-17
STSMACOM-537 remove stripes-logger

zburke:
`stripes-logger` is not directly instantiated here, only leveraged via
the `stripes` object passed in via props. Hence, no dependency is
necessary.

Refs [STSMACOM-537](https://issues.folio.org/browse/STSMACOM-537)


zburke:
... of course it's not that simple. Because {{@folio/stripes-logger}} is a peer-dependency within the stripes-* packages, we have to provide it here as a dev-dep because yarn doesn't satisfy peers. Re-opening, alas. The next time we decided to publish a breaking change to {{@folio/stripes}} itself, we can make this change across the stripes-* repos.

PR #1238, opened 2022-05-27
STSMACOM-646 consolidate to one testing root.

JohnC-80:
Due to varying approaches in mounting components for testing (some using the local version of 'mount' or 'setupApplication' vs the version from stripes-components) multiple elements were being rendered into the DOM, which meant tests were leaking *something into the others.

### Approach
Use `stripes-smart-components'` own setupApplication functions to render tests that were previously using `stripes-components`'  `mount()` call.

Additionally, there seemed to be some inconsistent failures due possibly to stacking beforeEach callls that would set up stripes-config in different ways - by default, `setupApplication()` would set up a beforeEach call that would create a 'dummy' app, and set the component to return `null` - which would render nothing in the component area.

In each failure you could consistently see (via added console output) that the original setup application `() => null` had been what was rendered instead of the intended, 'mounted' component.


zburke (approved):
👏 👏 👏 ❤️

mkuklis (approved):
👏

PR #1244, opened 2022-06-07
STSMACOM-668 ControlledVocab - optimistic locking

AndreiShumski2:
Purpose
https://issues.folio.org/browse/STSMACOM-668
Display appropriate error message for optimistic locking

### DIFF
--- a/lib/EditableList/processBadResponse.js
+++ b/lib/EditableList/processBadResponse.js
@@ -78,5 +78,12 @@ export default async function processBadResponse(response, defaultErrorCode) {
     }
   }
 
+  if (response.status === 409) {
+    return {
+      fieldErrors: {},
+      commonErrors: [<FormattedMessage id="stripes-smart-components.error.conflict" />],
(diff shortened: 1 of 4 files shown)


usavkov-epam (approved):

lib/EditableList/processBadResponse.js
```diff
@@ -78,5 +78,12 @@ export default async function processBadResponse(response, defaultErrorCode) {
     }
   }
 
+  if (response.status === 409) {
+    return {
```
mkuklis, line 82: @AndreiShumski2 could you use `getDefaultErrorMessage('conflict')` here instead?

mkuklis (approved):

lib/EditableList/processBadResponse.js
zburke: That's a nice idea, and a simpler way to get the same effect. These functions could really use some comments!

PR #1249, opened 2022-06-13
STSMACOM-671: Search term isn't persisted in SearchAndSort

Javokhir12:
## Purpose
- In `ui-data-import`, We are saving `searchTerm` to `sessionStorage` and pass it to `<SearchAndSort />` as `query` property of `parentResources` prop. But the problem is that `search box` is getting its initial value from local state `locallyChangedSearchTerm`. The logic is comparing it with `undefined` which is always false because `locallyChangedSearchTerm` has initial value of `""` and never gets initial value from `props`.
- Also when refresh the page `searchTerm` will be lost but searched results are persisted because of url query string.
## Refs
- https://issues.folio.org/browse/STSMACOM-671

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change history for stripes-smart-components
 
+## 7.3.0 IN PROGRESS
+* Use query from props as initial value for SearchAndSort search input instead of local state. Refs STSMACOM-671.
+
 ## [7.2.0](https://github.com/folio-org/stripes-smart-components/tree/v7.2.0) (2022-06-14)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v7.1.0...v7.2.0)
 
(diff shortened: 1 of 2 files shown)


Javokhir12:
@mkuklis in #586, Did you mean browser back button? if so I tested these changes in data-import, it is working as expected.

mkuklis:
@Javokhir12 yes I just mostly want to make sure that by fixing this we don't introduce another issue.

The steps to reproduce it were described here:

https://issues.folio.org/browse/STSMACOM-232

Javokhir12:
> @Javokhir12 yes I just mostly want to make sure that by fixing this we don't introduce another issue.
> 
> The steps to reproduce it were described here:
> 
> https://issues.folio.org/browse/STSMACOM-232
> 
> Would you mind trying your change with ui-inventory?

@mkuklis Thanks. I will test it and post my findings here.

Javokhir12:
@mkuklis Here it is. I also tried to go back with browser buttons and it is working.

https://user-images.githubusercontent.com/89512426/173413435-8bd998d1-2c34-4a2c-8856-ad30840372b1.mp4

PR #1251, opened 2022-06-20
STSMACOM-672 - If two patrons in a row have pop-up notes, all should appear

Terala-Priyanka:
Issue : If two patrons in a row have pop-up notes, all should appear
https://issues.folio.org/browse/STSMACOM-672


BogdanDenis (approved):

mariiaschv (approved):

Svitlana-Korol (approved):

zburke (commented):
The current PR description amounts to "fix bug". Please explain either in comments or in an updated PR description how the code changes resolve the bug described in the ticket. 

Without that, in order to review your changes, I basically have to re-troubleshoot the bug in order to understand what causes the problem and then assess how your change fixes it.

Terala-Priyanka:
@zburke 

1. As per the existing code, patron's notes are fetched and handled when the ref "isNotesLoading" has a value of false.
2. On page load, ref is initiated to false and thus the first patron notes are fetched and handled. At that point of time this ref is assigned true.
3. In order to resolve the issue, the ref value had to be assigned false when the last note is either deleted or closed.
4. Also I have added some clean up code when the last note of patron is either deleted or closed.

PR #1253, opened 2022-06-21
STSMACOM-676 Reset SearchAndSort query going back in history

BogdanDenis:
## Description
Fix issue when `<SearchAndSort>` query doesn't return to old value, when you go back in browser history
Component's `locallyChangedSearchTerm` state value is prioritized over query parameter in url, so we need to update it when url value changes

## Screenshots

https://user-images.githubusercontent.com/19309423/174809563-e0ebd28f-f65e-4ed2-9583-70021aab912a.mp4

## Issues
[STSMACOM-676](https://issues.folio.org/browse/STSMACOM-676)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 7.3.0 IN PROGRESS
 * Search term isn't persisted in SearchAndSort. Refs STSMACOM-671.
+* Fix `<SearchAndSort>` query doesn't reset when clicking back in browser. Fixes STSMACOM-676.
 
 ## [7.2.0](https://github.com/folio-org/stripes-smart-components/tree/v7.2.0) (2022-06-14)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v7.1.0...v7.2.0)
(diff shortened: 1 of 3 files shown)


Terala-Priyanka (approved):

Svitlana-Korol (approved):

JohnC-80 (approved):

JohnC-80:
The only case where the search field's value wouldn't need to be synced with the query parameter is that of a plugin using SearchAndSort... 

Most of the "find-x" plugins that I'm aware of use SearchAndSortQuery, but this change would break any that use SearchAndSort - and don't depend directly on the URL for their values unless the query prop were being passed in some other way.

BogdanDenis:
> The only case where the search field's value wouldn't need to be synced with the query parameter is that of a plugin using SearchAndSort since plugins don't affect the browser query directly and usually go through stripes-connect.
> 
> Most of the "find-x" plugins that I'm aware of use SearchAndSortQuery, but this change would break any that use SearchAndSort - and don't depend directly on the URL for their values unless the query prop were being passed in some other way.

You're right, that's a missed case.
I added a new prop `syncQueryWithUrl` that will update local state from url if it's set to `true`

mkuklis (approved):

JohnC-80 (approved):

PR #1255, opened 2022-06-23
STSMACOM-678 - SASQ should provide way to apply initial query on mount.

JohnC-80:
## Approach

provide a boolean prop `setQueryOnMount` - false by default - that can be used to call the querySetter, so that any initial query state can be applied to the application's query.


EthanFreestone (approved):
Cheers @JohnC-80 this is exactly what we needed!

PR #1264, opened 2022-07-13
STSMACOM-684: new prop for SearchAndSort - 'validateSearchOnSubmit'

ODanyor:
Added a new prop for `SearchAndSort` component to validate search value on submit.


BogdanDenis (approved):

mkuklis:
@ODanyor what is the purpose of this prop? I'm just curious what is being validated during the search?
Should the search stop if the validation fails?

ODanyor:
> @ODanyor what is the purpose of this prop? I'm just curious what is being validated during the search? Should the search stop if the validation fails?

The purpose of the prop is to add some custom validation over the search query. If the validation fails, the search submission is prevented.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Mark `initialValues` as immutable in `<LocationForm>`. Fixes STSMACOM-679.
 * Fix `<SearchAndSort>` query doesn't reset when clicking back in browser. Fixes STSMACOM-676.
 * Copy query into search field when navigating back and forth. Fixes STSMACOM-681.
+* CLONE - browse contributors with special characters shows incomplete error message. Refs STSMACOM-684.
```
zburke, line 9: Describe the solution (what changed), not the problem. This message parrots the problem from the ticket, which is totally disconnected from the implementation here, i.e. this change has nothing to do with special characters or error messages. Instead, describe the solution, e.g. "Accept `validateSearchOnSubmit` prop to prevent invalid searches"

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -235,6 +235,7 @@ class SearchAndSort extends React.Component {
     }).isRequired,
     syncQueryWithUrl: PropTypes.bool,
     title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+    validateSearchOnSubmit: PropTypes.func, // should return bool. It prevents search submission on false
```
zburke, line 238: `validate...` is one of those tricky function names that doesn't clearly indicate how a function behaves: does it return true/false, or return valid input/null, or return void/throw? 

The comment here perfectly clarifies this (thank you!) so you don't need to change anything. Just be careful with this construction, and think about if there is a way you could name things so that you don't have to explain them with a comment.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -504,7 +505,12 @@ class SearchAndSort extends React.Component {
       locallyChangedSearchTerm,
       locallyChangedQueryIndex,
     } = this.state;
+    const { validateSearchOnSubmit } = this.props;
     const qindex = locallyChangedQueryIndex || this.props.selectedIndex;
+    const isSearchValid = validateSearchOnSubmit?.(locallyChangedSearchTerm);
+    const shouldPreventSearch = validateSearchOnSubmit && !isSearchValid;
```
zburke, line 511: It's a little awkward to check the presence of `validateSearchOnSubmit` twice here, and since you never use `shouldPreventSearch` again, storing it seems superfluous. Maybe check all the conditions at once?
```
if (validateSearchOnSubmit && !validateSearchOnSubmit(locallyChangedSearchTerm) {
  return;
}
```

zburke (changes requested):
Please clean up the CHANGELOG description so it reflects the change that happened instead of the bug that was submitted. 

Despite my comments on the code itself, what you have is fine and does not need to be altered. Just keep these notes in mind for future PRs.

lib/SearchAndSort/SearchAndSort.js
mkuklis: I was just going to say exactly the same thing here.

ODanyor: Looks great! thanks

zburke (approved):

PR #1270, opened 2022-07-27
STSMACOM-689: Failed prop type: Invalid prop `regExpForQuery` of type `regexp` supplied to `SearchAndSort`, expected `string`.

BogdanDenis:
## Description
Fix `regExpForQuery` prop type to expect instance of RegExp class

Removed unused `this.resultsList` and `onClearSearch` to fix code smells. `this.connectedViewRecord` code smell is a false positive as it's actually used in code

## Issues
[STSMACOM-689](https://issues.folio.org/browse/STSMACOM-689)


Dmytro-Melnyshyn (approved):

zburke (commented):
Does the defaultProps value need to be changed too? It's an empty string at present but I think it should be null or `new RegExp()`.

BogdanDenis:
> Does the defaultProps value need to be changed too? It's an empty string at present but I think it should be null or `new RegExp()`.

That's a good point - set default value to `null`

zburke (approved):

PR #1280, opened 2022-08-19
STSMACOM-453 - refactor ControlledVocab to accept translation keys as…

Terala-Priyanka:
## Purpose
refactor ControlledVocab to accept translation keys as props and issue a console deprecation warning
Issue: [STSMACOM-453](https://issues.folio.org/browse/STSMACOM-453)

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -139,11 +139,20 @@ class ControlledVocab extends React.Component {
     /*
       sample translation prop
       {
-        "cannotDeleteHeader": "ui-users.cv.foo.cannotDeleteHeader",
-        "cannotDeleteMessage": "ui-users.cv.foo.cannotDeleteMessage",
+        cannotDeleteTermHeader: "Cannot delete {type}",
+        cannotDeleteTermMessage: "This {type} cannot be deleted, as it is in use by one or more records.",
+        deleteEntry: "Delete {entry}",
+        termDeleted: "The {type} <b>{term}</b> was successfully <b>deleted</b>",
+        termWillBeDeleted: "The {type} <b>{term}</b> will be <b>deleted.</b>"
       }
     */
-    translations: PropTypes.object,
+    translations: PropTypes.shape({
+      cannotDeleteTermHeader: PropTypes.string,
+      cannotDeleteTermMessage: PropTypes.string,
+      deleteEntry: PropTypes.string,
+      termDeleted: PropTypes.string,
+      termWillBeDeleted: PropTypes.string,
+    }),
     /*
     * Allows for custom field validation. The function is called for each record (row) and should
     * return an empty object for no errors, or an object where the keys are the field names and
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -139,11 +139,11 @@ class ControlledVocab extends React.Component {
     /*
       sample translation prop
       {
-        cannotDeleteTermHeader: "Cannot delete {type}",
-        cannotDeleteTermMessage: "This {type} cannot be deleted, as it is in use by one or more records.",
-        deleteEntry: "Delete {entry}",
-        termDeleted: "The {type} <b>{term}</b> was successfully <b>deleted</b>",
-        termWillBeDeleted: "The {type} <b>{term}</b> will be <b>deleted.</b>"
+        cannotDeleteTermHeader: "Cannot delete patron group",
+        cannotDeleteTermMessage: "This patron group cannot be deleted, as it is in use by one or more records.",
+        deleteEntry: "Delete patron group",
+        termDeleted: "The patron group <b>{term}</b> was successfully <b>deleted</b>",
+        termWillBeDeleted: "The patron group <b>{term}</b> will be <b>deleted.</b>"
       }
     */
     translations: PropTypes.shape({
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 import { Link } from 'react-router-dom';
 import { isEqual, uniqueId, pickBy, noop } from 'lodash';
+import { deprecated } from 'prop-types-extra';
 
 import {
   Button,
@@ -83,7 +84,10 @@ class ControlledVocab extends React.Component {
     id: PropTypes.string,
     itemTemplate: PropTypes.object,
     label: PropTypes.node.isRequired,
-    labelSingular: PropTypes.node.isRequired,
+    labelSingular: deprecated(
+      PropTypes.node,
+      'Provide translation prop instead'
+    ),
     limitParam: PropTypes.string,
     listFormLabel: PropTypes.node,
     listSuppressor: PropTypes.func,
@@ -132,6 +136,14 @@ class ControlledVocab extends React.Component {
       connect: PropTypes.func.isRequired,
       hasPerm: PropTypes.func.isRequired
     }).isRequired,
+    /*
+      sample translation prop
+      {
+        "cannotDeleteHeader": "ui-users.cv.foo.cannotDeleteHeader",
+        "cannotDeleteMessage": "ui-users.cv.foo.cannotDeleteMessage",
+      }
+    */
+    translations: PropTypes.object,
     /*
     * Allows for custom field validation. The function is called for each record (row) and should
     * return an empty object for no errors, or an object where the keys are the field names and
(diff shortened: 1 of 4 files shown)


lib/ControlledVocab/ControlledVocab.js
```diff
@@ -499,7 +534,11 @@ class ControlledVocab extends React.Component {
           <ConfirmationModal
             id="delete-controlled-vocab-entry-confirmation"
             open={this.state.showConfirmDialog}
-            heading={<FormattedMessage id="stripes-core.button.deleteEntry" values={{ entry: type }} />}
+            heading={
+              deleteEntry ?
+                <FormattedMessage id={deleteEntry} /> :
```
NikitaSedyx, line 539: should we provide `values` prop here?

Terala-Priyanka: I think it is not needed when using the translations prop.

zburke: I don't think we need the `values` prop any longer because `deleteEntry` can correspond to sth like `Delete patron group` or `Delete material type`. 

Ofc, we do still need it for `stripes-core.button.deleteEntry` because it looks like `Delete {entry}`.

zburke (changes requested):
This implementation looks perfect! 

My only request is to document the new `translations` prop instead of just providing a few sample keys:
* enumerate all keys, either via a comment or by providing `PropTypes.shape({ ... })` instead of just `PropTypes.object`
* provide sample values to provide guidance to the implementors about what phrases they need to be providing. You can probably get good examples from the existing `stripes-smart-components.cv.[...]` fields, e.g. 
  * `termDeleted`: "The patron group <b>{term}</b> was successfully <b>deleted</b>"

It would be fine to do this in a comment in the prop-types where you have the "sample translation prop" comment now. Even more awesome, create a README.md and stick 'em in there. Given there is 0 documentation for this component at present :sob: this clearly stretches the scope of the ticket, so it's fine if you want to just add comments and move on.

NikitaSedyx (approved):
apart from @zburke request to provide a doc

Terala-Priyanka:
@zburke 
I have added the comments as per your review.
I think I need to gain more insight into CV component before writing a well defined readme.
I would be happy to pick a new ticket for this.

lib/ControlledVocab/ControlledVocab.js
```diff
@@ -132,6 +136,23 @@ class ControlledVocab extends React.Component {
...
       hasPerm: PropTypes.func.isRequired
     }).isRequired,
+    /*
+      sample translation prop
+      {
+        cannotDeleteTermHeader: "Cannot delete {type}",
+        cannotDeleteTermMessage: "This {type} cannot be deleted, as it is in use by one or more records.",
+        deleteEntry: "Delete {entry}",
+        termDeleted: "The {type} <b>{term}</b> was successfully <b>deleted</b>",
+        termWillBeDeleted: "The {type} <b>{term}</b> will be <b>deleted.</b>"
```
zburke, line 146: This is a great start, but it's confusing because these examples use placeholders for `type` and `term`, but `{type}` won't be a variable in this implementation (e.g. for [`cannotDeleteTermHeader`](https://github.com/folio-org/stripes-smart-components/blob/666c28b204dcbe402bcacc5e7a80e0cb5a98e03d/lib/ControlledVocab/ControlledVocab.js#L393)) while `{term}` still is (e.g.  for [`termDeleted`](https://github.com/folio-org/stripes-smart-components/blob/666c28b204dcbe402bcacc5e7a80e0cb5a98e03d/lib/ControlledVocab/ControlledVocab.js#L331-L336)). I think we need something like this:
```
cannotDeleteTermHeader: "Cannot delete patron group",
cannotDeleteTermMessage: "This patron group cannot be deleted, as it is in use by one or more records.",
deleteEntry: "Delete patron group",
termDeleted: "The patron group <b>{term}</b> was successfully <b>deleted</b>",
termWillBeDeleted: "The patron group <b>{term}</b> will be <b>deleted.</b>"
```
Sorry I wasn't more specific in my initial comment. I clearly didn't think through the details until now.

zburke (changes requested):
`PropTypes.shape` is perfect but the example prop needs to be tweaked so it's clear which props take substitutions and which do not.

zburke:
@julianladisch, will this implementation work for German? I am pretty confident it meets the needs specified in [STSMACOM-453](https://issues.folio.org/browse/STSMACOM-453) but I am not a native German speaker so I would love to get your eyes on this!

lib/ControlledVocab/ControlledVocab.js
```diff
@@ -499,7 +543,11 @@ class ControlledVocab extends React.Component {
           <ConfirmationModal
             id="delete-controlled-vocab-entry-confirmation"
             open={this.state.showConfirmDialog}
-            heading={<FormattedMessage id="stripes-core.button.deleteEntry" values={{ entry: type }} />}
+            heading={
+              deleteEntry ?
+                <FormattedMessage id={deleteEntry} /> :
+                <FormattedMessage id="stripes-core.button.deleteEntry" values={{ entry: type }} />
+            }
```
julianladisch, line 550: Can you add more tests to bump the code coverage of new code from 58.3% to at least 80%?
https://sonarcloud.io/component_measures?metric=new_coverage&selected=org.folio%3Astripes-smart-components%3Alib%2FControlledVocab%2FControlledVocab.js&view=list&pullRequest=1280&id=org.folio%3Astripes-smart-components

julianladisch: @zburke The pull request looks promising and works for German.

lib/ControlledVocab/ControlledVocab.js
```diff
@@ -132,6 +136,23 @@ class ControlledVocab extends React.Component {
...
+    /*
+      sample translation prop
+      {
+        cannotDeleteTermHeader: "Cannot delete patron group",
+        cannotDeleteTermMessage: "This patron group cannot be deleted, as it is in use by one or more records.",
+        deleteEntry: "Delete patron group",
+        termDeleted: "The patron group <b>{term}</b> was successfully <b>deleted</b>",
+        termWillBeDeleted: "The patron group <b>{term}</b> will be <b>deleted.</b>"
+      }
+    */
```
julianladisch, line 148: I expect documentation in the readme.md following the examples in https://github.com/folio-org/stripes-smart-components/tree/master/lib/ConfigManager and https://github.com/folio-org/stripes-components#links-to-documentation-of-specific-components-and-utilities

zburke: Agreed, per-component README files would be great, @julianladisch. If we had such documentation already, then it would be easy to add this new prop to it, and we'd be done! 

But given there is no documentation whatsoever, let's handle that task in a separate ticket so we don't explode the scope and point estimate of this ticket, which are already accepted. Then, we can also assign that task to somebody who is familiar with the full scope of this component.

lib/ControlledVocab/tests/mountComponent.js
```diff
@@ -105,30 +105,61 @@ export default async function (editable, server) {
...
-      hiddenFields={['description', 'numberOfObjects']}
-      nameKey="name"
-      // columnWidths={{ 'name': 300, 'code': 50 }}
-      actionSuppressor={{ edit: onEditHandler, delete: onDeleteHandler }}
-      id="institutions"
-      sortby="name"
-      editable={editable}
-    />
-  );
+  labelSingular ?
```
NikitaSedyx, line 108: why not to prepare specific props based on condition instead of putting whole component mount under condition?

lib/ControlledVocab/tests/ControlledVocab-test.js
```diff
@@ -20,6 +20,24 @@ describe('ControlledVocab', () => {
   const cv = EditableList();
   const firstRow = EditableListRow();
 
+  describe('render CV when translations prop is provided', () => {
+    setupApplication();
+
+    // eslint-disable-next-line no-undef
+    beforeEach(() => mountComponent(false, server));
+
+    it('should have row count 5', () => cv.has({ rowCount: 5 }));
```
NikitaSedyx, line 29: what does this test do in context of testing translations used when translations prop is provided or not? IMO nothing

zburke (approved):
👏 👏 👏

lib/ControlledVocab/tests/mountComponent.js
Terala-Priyanka: Updated to handle through props.

lib/ControlledVocab/tests/ControlledVocab-test.js
Terala-Priyanka: 1. It is ensuring that the component is mounting as expected when new props are provided and labelSingular prop is not.
2. rest of the tests in the same suite, ensure that the translations are picked.

PR #1281, opened 2022-08-24
STSMACOM-324 Refactor SearchAndSort away from deprecated lifecycle.

JohnC-80:
This PR moves all logic out of the deprecated `componentWillReceiveProps` lifecycle method. The list of functionalities of the original that were changed/moved to other lifecycles:
- Check for resolved resource completion and select completed row (moved to `componentDidUpdate`)
- Inform screenreader upon arival of new search results (moved to `componentDidUpdate`)
- Show detail page when there's only a single result (moved to `componentDidUpdate`)
- Reset qindex if qindex is missing from the query string (moved to `getDerivedStateFromProps`) (this happens when the user clicks among the 'instance', 'holding', 'item' tabs in ui-inventory.
- Sync search term value with query parameter (moved to `getDerivedStateFromProps`)

In order to properly derive the state, the previous query string had to be retained in state for string comparison. SandSQuery works in a similar fashion when syncing the UI with the query string.


mkuklis:
You are a brave man @JohnC-80!

zburke (approved):
👏 finally somebody is brave enough to tackle this! I'm really pleased to see this work. I was able to manually verify in ui-inventory workspace build that all conditions still work as expected ... I think. 

The only ding: test coverage here is only mediocre to begin with and there's no change to that here. Aside from that, this looks great. I appreciate the description in the PR of what moved and where it went; that made the code changes so much clearer. Thank you!

JohnC-80:
Thanks @zburke , @mkuklis I will definitely add some tests to this PR before merging.

zburke:
@JohnC-80 , did you intend to commit `.stripesclirc` and `.vscode`? I think those are usually omitted.

mkuklis (approved):
❤️

JohnC-80:
@zburke thanks - let me ignore those guys.

zburke (changes requested):
Thanks for the additional test coverage. That's awesome to see! I have no qualms about the substance here, but there is a little cleanup to do before this gets merged:

* Let's omit `package-lock.json` until we formalize the migration from yarn to npm. 
* Lint is unhappy.

zburke (approved):

PR #1283, opened 2022-09-02
STSMACOM-696 leverage yarn.lock

BogdanDenis:
## Description
In @zburke's words:

> Begin committing yarn.lock so we can choose the cadence of our updates. This reduces the surprises introduced by dependencies and transitive dependencies at the PR-level. We will still have to deal with this issue at the platform-level (a semver contract with caret or tilde means we are claiming compatibility with any later minor or patch version, even if an earlier version is in our yarn.lock) but at least we don't have to deal with it immediately on every PR in every repository.
This should also reduces potential differences between devs working in tandem, and between devs and CI. Yay for predictability!

## Issues
[STSMACOM-696](https://issues.folio.org/browse/STSMACOM-696)


Dmytro-Melnyshyn (approved):

zburke (approved):
Thank you for picking this up and moving it forward! I hope this works out well and becomes a pattern we can apply across all repositories.

I think we should remove these lines from the workflow scripts too, agreed? 
```
      - name: Publish yarn.lock
        uses: actions/upload-artifact@v2
        if: failure()
        with:
          name: yarn.lock
          path: yarn.lock
          retention-days: 5
```

PR #1286, opened 2022-09-13
STSMACOM-699: Implement reinitialization functionality to EditableListForm component

artem-blazhko:
## Approach
In some cases when we used the `EditableListForm` component, we had situations when local property `status` is not cleared after receiving new data.
As a result, we could see unnecessary validation error messages.
Also, we had cases when edit mode was incorrectly turned on. 

In the scope of this PR, I have added a new `shouldReinitialize` property which enables reinitialization if necessary.

## Refs
[STSMACOM-699](https://issues.folio.org/browse/STSMACOM-699)


lib/EditableList/EditableListForm.js
```diff
@@ -221,6 +227,20 @@ class EditableListForm extends React.Component {
...
+    if (this.props.shouldReinitialize && this.shouldUpdateStatus(prevProps.initialValues)) {
+      this.setState({ status: buildStatusArray(this.props.initialValues.items) });
+    }
+  }
+
+  shouldUpdateStatus(previousValues) {
+    return (
+      this.props.initialValues &&
+      previousValues &&
+      this.props.initialValues !== previousValues
```
mkuklis, line 240: @artem-blazhko `intialValues` is an object with some nested structure (an array) so `!==` won't work here. You will need to use a deep comparison with https://lodash.com/docs/#isEqual for example.

artem-blazhko: @mkuklis Fixed

mkuklis (approved):
Very nice and clean approach @artem-blazhko!

Dmitriy-Litvinenko (approved):

PR #1289, opened 2022-09-29
STSMACOM-700 SearchAndSortQuery requires 2 clicks of the clear to successfully clear filters.

JohnC-80:
Jira: https://issues.folio.org/browse/STSMACOM-700

Problem: The first click of the clear filters button would trigger an update to internal query state of SASQ, which updates the locatoin query as expected, but if `syncToLocationQuery`  is enabled, derived state will include the `initialFilterState` which will trump the now-empty values from the query when deriving the state. This isn't a problem with most existing SASQ implementations, as they don't make much use of the `initial*State` props.

*Approach:* Remove the initial query state prop from the `gDSFP` call since we only want to consider the location query itself when syncing to it.

### DIFF
--- a/lib/SearchAndSort/SearchAndSortQuery.js
+++ b/lib/SearchAndSort/SearchAndSortQuery.js
@@ -204,10 +204,12 @@ class SearchAndSortQuery extends React.Component {
       if (props.location.search !== state.previousQuery) {
         nextState.previousQuery = props.location.search;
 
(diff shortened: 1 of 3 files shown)


EthanFreestone (approved):
Thank you John! Apologies again for digging out obscure bugs in the codebase, and thanks as ever for the time taken to worm them out.

mkuklis (approved):

PR #1293, opened 2022-10-12
STSMACOM-704 Add props to ControlledVocab components

vashjs:
For tickets:
https://issues.folio.org/browse/UITEN-225
https://issues.folio.org/browse/UITEN-226
https://issues.folio.org/browse/UITEN-227
https://issues.folio.org/browse/UITEN-228

It's required to hide/disable create button in "View only" mode.

Solution: add `canCreate` prop to `ControlledVocab.js` component.

### DIFF
--- a/lib/ControlledVocab/ControlledVocab.js
+++ b/lib/ControlledVocab/ControlledVocab.js
@@ -161,6 +161,7 @@ class ControlledVocab extends React.Component {
     */
     validate: PropTypes.func,
     visibleFields: PropTypes.arrayOf(PropTypes.string),
+    canCreate: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -192,6 +193,7 @@ class ControlledVocab extends React.Component {
     //  !{limitParam:-limit}
     // in the manifest above.
     actuatorType: 'rest',
+    canCreate: true
   };
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


uladislausamets (approved):

mkuklis (approved):

lib/ControlledVocab/ControlledVocab.js
```diff
@@ -485,6 +487,7 @@ class ControlledVocab extends React.Component {
               // TODO: not sure why we need this OR if there are no groups
               // Seems to load this once before the groups data from the manifest
               // is pulled in. This still causes a JS warning, but not an error
+              canCreate={this.props.canCreate}
               contentData={rows}
```
zburke, line 491: @vashjs, this new prop, `canCreate`, separates a comment block (`TODO...`) from the prop it describes (`contentData`). 

In fact, the comment refers to [an old formulation of `contentData`](https://github.com/folio-org/stripes-smart-components/commit/d08eac47509670778648d68fba7e8e6e4406f5fd#diff-0748e4d1503798efcf342204a4077ea037fc95a88d0408661cee6a2e947b3b07R89) so it should just be removed, but please take note of this situation. If the comment were still relevant then this separation would be unfortunate.

vashjs: @zburke thanks for catching this.

PR #1297, opened 2022-11-15
STSMACOM-707 retain locallyChangedSearchTerm when syncing to query parameters

JohnC-80:
Problem:
If the user entered a search term and then modified the `qindex`, the search term would be lost or changed to whatever was currently in the query parameters.

Approach:
Check if there is a locally changed search term before syncing to the query parameter value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 7.4.0 IN PROGRESS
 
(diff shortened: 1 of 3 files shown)


zburke (approved):

mkuklis (approved):
Thank you @JohnC-80 ❤️

PR #1304, opened 2022-12-02
STSMACOM-710: query-string ^7.1.2 fixing DoS CVE-2022-38900

julianladisch:
https://nvd.nist.gov/vuln/detail/CVE-2022-38900

However, bumping query-string from ^5.0.0 to ^7.1.2 causes 3 test failures in the build. This is fixed by pinning axe-core to 4.4.5, a reminder to unpin it has been created: [STSMACOM-713](https://issues.folio.org/browse/STSMACOM-713)

Refs [STSMACOM-710](https://issues.folio.org/browse/STSMACOM-710)


JohnC-80:
The test failures appear to result from an update to `axe-core` that showed up in the updated `yarn.lock` file. Failures are not related to the `query-string` dependency. I made some adjustments. 🤞

JohnC-80 (approved):

package.json
```diff
@@ -50,7 +50,7 @@
     "@folio/stripes-testing": "^4.2.0",
     "@folio/stripes-util": "^5.1.0",
     "@formatjs/cli": "^4.2.20",
-    "axe-core": "^4.3.5",
+    "axe-core": "4.4.3",
```
zburke, line 53: Please file a follow-up ticket to remove this exact-version dep. I totally understand why it's in here now, and I support it, but it's not in our best interest to depend on exact versions in the long term.

zburke (approved):
❤️ 👏 👏 👏

package.json
JohnC-80: Good call. Filed https://issues.folio.org/browse/STSMACOM-713 for the upkeep task. eslint-jsx-a11y uses the latest and greatest so we can probably expect some weirdness if that gets hoisted. I'll have to update the version in `stripes-testing` to this version as well...

PR #1321, opened 2023-02-09
STSMACOM-728 Axe errors with repeated id for Tags label.

JohnC-80:
Tags was applying both an `aria-label` and `aria-labelledby` to its rendered multiselect. This both a) is not necessary and b) not good for accessibility - in fact, it did cause a failure in axe tests (not due to multiple labeling techniques present, however - but that's another story)

Approach: removed the `aria-labelledby` so that only the applied `aria-label` is used.

Bonus: included a test suite for 'Tags'


BogdanDenis (approved):
Looks great, thank you for adding the tests too!

PR #1322, opened 2023-02-13
STSMACOM-731: POC: Only 10 proxy/sponsor relationships appear in checkout pop-up modal

artem-blazhko:
## Purpose
Add `limit` query param to be able to get more than 10 items about sponsors and proxies.

## Refs
[STSMACOM-731](https://issues.folio.org/browse/STSMACOM-731)

## Screenshots/Video rec
2023-02-21_17h22_51


Dmitriy-Litvinenko (approved):

lib/ProxyManager/ProxyManager.js
```diff
@@ -59,6 +59,8 @@ class ProxyManager extends React.Component {
         records: PropTypes.arrayOf(PropTypes.object),
       }),
     }).isRequired,
+    sponsorLimitParam: PropTypes.string,
+    proxyLimitParam: PropTypes.string,
```
zburke, line 63: Hmmm, this moves the problem, but doesn't solve it. If all we're going to do is accept a new limit prop from an app (where presumably it's hard-coded), why not just hard-code it here to 35 or 45 or what ever the less-than-4096 character limit is? That's easier -- no changes are necessary in apps like ui-checkout that call this component -- and no less effective. 

The more complicated version is to run one query with `limit=0` to get the guaranteed-accurate count, and then run queries in batches of 35-50 until we retrieve them all. I asked the PO on UICHKOUT-847 if this accurate-but-laborious process is necessary, or if just bumping the limit higher than the default (10) will be good enough.

artem-blazhko: I think the first solution that you suggested has some limitations. If we have hard-coded values in `ProxyManager` component we can have the same bug in the future. 
The second approach has some potential issues. What should we do in the case if we have a few batches of 35-50 items and one request is failed? Should we request all batches one more time? Also, in the case when we have a big number of requests we will have a request queue in the browser because browser can handle a specific number of requests at the same time. This situation slows down the interaction between user and browser.
Also, if we use any of the approaches described above it will affect already existing functionality.
@zburke what do you think about it? Could you share your thoughts?

artem-blazhko:
@zburke could you take a look at this PR?

zburke (approved):
LGTM!

PR #1328, opened 2023-03-09
STSMACOM-735: Reset the previously selected query index when there is none in the next selected segment.

Dmytro-Melnyshyn:
## Description
When we make request with the query index `Title (all)` in the `Instance` segment and then switch to `Holdings` and make request here, but we don't have the `Title (all)` in the `Holdings` segment, we will see an [error](https://issues.folio.org/secure/attachment/58884/2023-03-09_15h31_41.mp4). Therefore, we need to check if the selected segment has previously picked the query index or not.
## Issues
[STSMACOM-735](https://issues.folio.org/browse/STSMACOM-735)
## Screencast

https://user-images.githubusercontent.com/77053927/224097687-198067ff-1c51-4218-8a8a-492decab032e.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 8.1.0 IN PROGRESS
 * Add `limit` query params to `ProxyManager` component. Refs STSMACOM-731.
+* Reset the previously selected query index when there is none in the next selected segment. Fixes STSMACOM-735.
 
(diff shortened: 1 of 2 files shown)


Dmytro-Melnyshyn:
@zburke @BogdanDenis please take a look at this PR.

BogdanDenis (approved):

zburke (approved):

PR #1330, opened 2023-03-14
STSMACOM-719 Optimize `useColumnManager` to reduce redundant renders

usavkov-epam:
## Purpose
https://issues.folio.org/browse/STSMACOM-719

`useColumnManager` hook contains a variable that is a dependency in other hooks:
```
const defaultColumnKeys = Object.keys(columnMapping);
```
```
const orderedVisibleColumns = useMemo(() => {
    const visibleColumnsSet = new Set(visibleColumns);

    return defaultColumnKeys.filter(key => visibleColumnsSet.has(key));
}, [defaultColumnKeys, visibleColumns]);
```

Since each re-render creates a new array of keys, changing the value of the reference of this variable will entail a number of redundant renders of dependent components.

<details>
<summary>Preview</summary>

chrome_fkyDsMObUQ

</details>

---

Memoization in this case is a more profitable operation than calling an additional rerender of an MCL with a hundred elements.

<details>
<summary>Preview</summary>

chrome_FO7PSUc9iq

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Add `limit` query params to `ProxyManager` component. Refs STSMACOM-731.
 * Disabled action menu if user does not have any of the required permissions. Refs STSMACOM-736.
 * Reset the previously selected query index when there is none in the next selected segment. Fixes STSMACOM-735.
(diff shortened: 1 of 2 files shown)


ryandberger (approved):
What a great improvement! Thanks for handling this!

vashjs (approved):

PR #1335, opened 2023-03-21
STSMACOM-738: Checkout/User notes pop-up > Excessive pop-ups

daniil-patckin:
## Purpose
If the user has a note and you try to search for user in the check-out application, the pop-up opens again for each action.
The pop-up should only display once when using the app for a session. 

## Issues
[STSMACOM-738](https://issues.folio.org/browse/STSMACOM-738)

## Approach
In the old code, the data was fetched every time the user click the button. We need to fetch the data again, but only when the user changes.

## Screncast

https://user-images.githubusercontent.com/124583895/226565381-aeee8d60-4320-40a2-8a43-9e8fa653e5da.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Optimize `useColumnManager` to reduce redundant renders. Refs STSMACOM-719.
 * Fix a page crush if searchableIndexes prop is missing. Fixes STSMACOM-735.
 * Fix an excessive Notes pop-ups. Fixes STSMACOM-738.
+* Accept override default column visibility in ColumnManager. Fixes STSMACOM-734.
 
 ## [8.0.0](https://github.com/folio-org/stripes-smart-components/tree/v8.0.0) (2023-01-30)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v7.3.0...v8.0.0)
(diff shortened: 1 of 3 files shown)


zburke (commented):
All the explanations here describe the symptom -- excessive popups -- but there is nothing describing the solution. What was wrong with the old code, and how does the new code address it?

zburke (approved):
LGTM. This approach provides a clear and clear explanation of the problem and the solution; thank you! 

I review a lot of PRs. Having a clear statement of the problem with the code makes it way easier/faster to assess the fix. Without that, I basically have to re-troubleshoot the problem myself and then evaluate whether the change properly addresses it. That can be really time consuming.

lib/Notes/NotePopupModal/NotePopupModal.js
```diff
@@ -114,13 +116,13 @@ const NotePopupModal = ({
...
   useEffect(() => {
-    if (entityId && !isNotesLoading.current) {
+    if (entityId && entityId !== prevEntityId && !isNotesLoading.current) {
       mutator.NotePopupModal_assignedNotes.GET()
         .then(handleGetNotesRequest);
 
       isNotesLoading.current = true;
     }
-  }, [entityId, handleGetNotesRequest, mutator.NotePopupModal_assignedNotes]);
+  }, [entityId, handleGetNotesRequest, mutator.NotePopupModal_assignedNotes, prevEntityId]);
```
BogdanDenis, line 125: I think we can fix it by removing `mutator.NotePopupModal_assignedNotes` from deps. It changes every render and triggers this effect

Dmytro-Melnyshyn (approved):

PR #1340, opened 2023-03-31
STSMACOM-742 avoid experimental decorator syntax

zburke:
Decorator syntax is still experimental and not all transpilers can handle it. In particular, `esbuild-loader` does not support it (https://github.com/evanw/esbuild/issues/104) and understandably doesn't plan to until the specification settles down. esbuild-loader offers a ~50% performance bump over babel-loader, and refactoring to stick with standard feels like a fair price to pay for that.

Refs [STSMACOM-742](https://issues.folio.org/browse/STSMACOM-742), [STRWEB-76](https://issues.folio.org/browse/STRWEB-76)


zburke:
This turns out to be unnecessary as esbuild-loader is able to handle decorators alone on a line. Lucky break!

PR #1345, opened 2023-04-13
STSMACOM-745: Fix import to stripes-connect

mkuklis:
https://issues.folio.org/browse/STSMACOM-745

This should be merged after https://github.com/folio-org/stripes-connect/pull/200 is merged


zburke:
@mkuklis, https://github.com/folio-org/stripes-connect/pull/200 has been reverted off and on a few times now. Do you think we should separate the "correctly configure exports" aspect of it from the "introduce transpilation" aspect? Because I think _that_ change is likely to stick, and would allows us to move forward with this one, independent of our transpilation experiments.

mkuklis:
@zburke I created a separate PR: https://github.com/folio-org/stripes-connect/pull/208

Let's merge these two first.

zburke (approved):
LGTM; thank you for helping on this long long road of cleaning up our API surfaces!

PR #1346, opened 2023-04-13
STSMACOM-718 Provide 'count' property to FormattedMessage

JohnC-80:
[STSMACOM-718](https://issues.folio.org/browse/STSMACOM-718)

Formatted message was previously throwing an error message in the browser console that it expected a 'count' property. The rendered message was sent to screen-reader output - including screenshot of result in browser element inspector. 🎉 

image


artem-blazhko (approved):

zburke (approved):
Oooh, thank you. Little fixes like this are so nice to see.

PR #1352, opened 2023-05-12
STSMACOM-750 Use correct proptype for 'expanded' prop for EditCustomFieldRecord

JohnC-80:
https://issues.folio.org/browse/STSMACOM-750

Removes the `isRequired` check from the `expanded` prop-type.

This is problematic because if the prop is just statically applied to appease the proptypes/clean the resulting console errors, the Accordion will be stuck in that non-interactive state - stuck open or closed.


zburke (approved):

MikeTaylor (approved):
Thanks, John, appreciated.

JohnC-80:
Sorry, moved to #1353 since I renamed the branch XD

PR #1354, opened 2023-05-17
lint

zburke:
lint. that is all.


lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
```diff
@@ -55,8 +55,8 @@ const EditCustomFieldsSettings = ({
   permissions,
   history,
 }) => {
-  const makeOkapiRequest = useCallback(makeRequest(okapi)(backendModuleId), [okapi, backendModuleId]);
-  const makeTitleRequest = useCallback(makeRequest(okapi)(), [okapi]);
+  const makeOkapiRequest = useCallback((url) => makeRequest(okapi)(backendModuleId)(url), [okapi, backendModuleId]);
```
MikeTaylor, line 58: What is the intent of these changes (and similar ones below)? The before and after look equivalent to me.

lib/CustomFields/utils/useCustomFieldsFetch.js
```diff
@@ -11,8 +11,8 @@ const useCustomFieldsFetch = (okapi, backendModuleId, entityType) => {
   const [customFieldsFetchFailed, setCustomFieldsFetchFailed] = useState(false);
 
   const makeOkapiRequest = useCallback(
-    makeRequest(okapi)(backendModuleId),
-    [backendModuleId, okapi.token, okapi.tenant, okapi.url]
+    (url) => makeRequest(okapi)(backendModuleId)(url),
+    [backendModuleId, okapi]
```
MikeTaylor, line 15: This one seems to change the semantics. If the okapi object itself is the same one as before but okapi.token has changed then the callback won't fire, will it?

(We won't see that for now, since the token, tenant and url never do change.)

lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
zburke: Functionally, they are equivalent and the change here is just to appease lint. `useCallback()` expects a function and an array, but since we don't use TypeScript, it has no way to recognize that `makeRequest(...)` returns a function (actually, it's a function that returns a function that returns a function that returns a function, which honestly is a bit much if you ask me) and so it whines.

lib/CustomFields/utils/useCustomFieldsFetch.js
zburke: I tried leaving the dependency array alone and changing the call  to
```
(url) => makeRequest({ okapi: { ... } })(backendModuleId)(url),
```
but I had test failures that I could not explain. 

But ... you're right that if the dependency array is only checked with shallow equality, this changes the behavior and won't do. TBH, I didn't think hard about this, but saw the simple `[backendModuleId, okapi]` dependency array elsewhere so I just copied it. I think this behavior is likely OK, but it's worth a closer look to make sure I have not just introduced a sneaky bug. I'll change this PR to "Draft" until I can confirm this is acceptable behavior.

lib/CustomFields/pages/EditCustomFieldsSettings/EditCustomFieldsSettings.js
MikeTaylor: Ah yes, the old trick where you have to make your code _less_ clear to persuade Lint that you know what you're doing. I go back and forth on this. I think sometimes it's better just to add an `// eslint-disable-next-line name-of-rule` comment. In this case the change you've made is _probably_ better than that, but it's a tough call.

And yes, a function that returns a function that returns a function that returns a function is just bad design. Functional programming is sold to us on the basis that it's easier to reason about. This kind of behaviour does not help to convey the truth of that.

zburke:
@gurleenkaurbp , in an ideal world, I'd love to figure out why the tests behave differently when these functions receive a single `okapi` object instead of separate `token`, `tenant`, and `url` parameters. At the same time, I think Mike is right that none of these values are expected to change over the course of a section so fussing about them may not be worth the effort, especially when we have the example of

MikeTaylor:
> So yes, I think my fear of the changing-token bug is legitimate ... but at the same time, sort of not really real, since tokens never change.

Only one concern: won't this stop being true when we finally implement refresh tokens?

zburke:
Yes, it is extremely unfortunate, @MikeTaylor , that these components don't leverage the project standards for interacting with APIs (react-query, stripes-connect), thus expanding the number of changes that will be necessary to implement token rotation.

lib/CustomFields/utils/useCustomFieldsFetch.js
zburke: Thanks for pushing back on this, @MikeTaylor. I was able to adjust the miragejs config to resolve the test failures, allowing the original dependency array to stay as-is.

PR #1361, opened 2023-06-12
Optional `resourcePath` prop for `<EntryManager>`

MikeTaylor:
The `<EntryList>` component was created on the assumption that the records it manages are returned in full when a list is requested: so, for example, the list of records returned from `/path/to/endpoint` are each the same shape as the individual ones obtained by ID at paths like `/path/to/endpoint/12335`. This pattern is generally followed by most web-service endpoints provided by most FOLIO modules.

Some modules, however, return a list of summary records (for example, just ID and title), so that it's necessary to fetch each full record separately by ID. When `<EntryList>` is used to manage records on such an endpoint, the path to that service should be passed in as the `resourcePath` prop. When this is done, the full record is fetched each time a user views or edits a record.

This is required by:
* https://issues.folio.org/browse/UIHAADM-7
* https://issues.folio.org/browse/UIHAADM-8
* https://issues.folio.org/browse/UIHAADM-9

### DIFF
--- a/lib/EntryManager/ConnectedWrapper.js
+++ b/lib/EntryManager/ConnectedWrapper.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { useOkapiKy } from '@folio/stripes/core';
 import { Loading } from '@folio/stripes/components';
 
-function ConnectedDetail({ resourcePath, initialValues, underlyingComponent, ...rest }) {
+function ConnectedWrapper({ resourcePath, initialValues, underlyingComponent, ...rest }) {
   const okapiKy = useOkapiKy();
   const [record, setRecord] = useState();
   const [loading, setLoading] = useState(false);
@@ -24,7 +24,7 @@ function ConnectedDetail({ resourcePath, initialValues, underlyingComponent, ...
   return <Component initialValues={record} {...rest} />;
 }
 
-ConnectedDetail.propTypes = {
+ConnectedWrapper.propTypes = {
   resourcePath: PropTypes.string.isRequired,
   initialValues: PropTypes.shape({
     id: PropTypes.string.isRequired,
@@ -32,4 +32,4 @@ ConnectedDetail.propTypes = {
   underlyingComponent: PropTypes.func, // React component
 };
 
-export default ConnectedDetail;
+export default ConnectedWrapper;
--- a/lib/EntryManager/ConnectedWrapper.js
+++ b/lib/EntryManager/ConnectedWrapper.js
@@ -7,6 +7,13 @@ function ConnectedWrapper({ resourcePath, initialValues, underlyingComponent, ..
   const okapiKy = useOkapiKy();
   const [record, setRecord] = useState();
   const [loading, setLoading] = useState(false);
+  const Component = underlyingComponent;
+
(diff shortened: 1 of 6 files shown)


lib/EntryManager/ConnectedWrapper.js
```diff
@@ -0,0 +1,40 @@
...
+  const [record, setRecord] = useState();
+  const [loading, setLoading] = useState(false);
+  const Component = underlyingComponent;
+
+  if (!initialValues.id) {
+    // Creating a new record: pass through
+    return <Component initialValues={initialValues} {...rest} />;
+  }
+
+  if (!record) {
```
mkuklis, line 17: @MikeTaylor this should be wrapped in `useEffect`.

mkuklis:
@MikeTaylor I like the way you approached this. I think we just need to wrap the fetch in `useEffect` and this should be ready to go.

mkuklis:
@MikeTaylor I think you could get rid of the loading state but I suggested `useEffect` to avoid executing async calls directly in render. Something like this could potentially work (totally untested):
 
````js
const id = initialValues?.id;

useEffect(() => {
  const fetchRecord = async () => {
    const resp = await okapiKy(`${resourcePath}/${id}`);
    const rec = await res.json();
    setRecord(rec);
  };

  if (!record && id) {
    fetchRecord();
  }
}, [id, record]);

if (!id) {
  // Creating a new record: pass through
  return <Component initialValues={initialValues} {...rest} />;
}

if (!record) {
  return <Loading />;
}
````

**Comment by MikeTaylor** [comment]
_sigh_ As is so often the case, `useEffect` is mysteriously firing when it should not.

**Comment by MikeTaylor** [comment]
Seems like the problem is with the `okapiKy` function. ESLint wants it included in `useEffect`'s dependency list, but we seem to be getting back a different instance of that function on each render. Removing it from the list fixes the problem.

**Comment by MikeTaylor** [comment]
@mkuklis OK, how does that look?

File: lib/EntryManager/ConnectedWrapper.js [hunk 2]
```diff
@@ -0,0 +1,42 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { useOkapiKy } from '@folio/stripes/core';
```
mkuklis, line 3: @MikeTaylor given this a stripes module I think we need to use `@folio/stripes-core` here

lib/EntryManager/ConnectedWrapper.js
```diff
@@ -0,0 +1,42 @@
...
+import { Loading } from '@folio/stripes/components';
```
mkuklis, line 4: @MikeTaylor same here `@folio/stripes-components`

mkuklis (commented):
@MikeTaylor I think the errors we see are related to stripes imports. Hopefully the issues go away after you adjust it.

lib/EntryManager/ConnectedWrapper.js
MikeTaylor: Yep, I just landed on that! :-)

MikeTaylor: It was realising that even the lint phase failed that gave it away!

PR #1367, opened 2023-06-19
`<EntryManager>`: `parseInitialValues` prop works with `resourcePath`

MikeTaylor:
The `parseInitialValues` function is now applied correctly when
`resourcePath` is defined (see STSMACOM-757) so that full records are
loaded from a WSAPI.

Previously, the transformation would be applied to the brief version
of the record before the full record was loaded, so that fields that
occur only in the full version of the record were not transformed.

Also, I have documented these previously undocumented props.

Fixes STSMACOM-761.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Remove `isRequired` check from `expanded` prop on EditCustomFieldsRecord. Refs STSMACOM-750.
 * Unlock `locallyChangedSearchTerm` sync with `query` parameter. Fixes STSMACOM-754.
 * Enhance `<EntryManager>` with optional `resourcePath` prop to fetch full record. Fixes STSMACOM-757.
+* Make `<EntryManager>`s `parseInitialValues` function work with correctly when `resourcePath` is defined. Document this previously undocumented prop and its converse, `onBeforeSave`. Fixes STSMACOM-761.
 
 ## [8.0.0](https://github.com/folio-org/stripes-smart-components/tree/v8.0.0) (2023-01-30)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v7.3.0...v8.0.0)
--- a/lib/EntryManager/EntrySelector.js
+++ b/lib/EntryManager/EntrySelector.js
@@ -58,6 +58,7 @@ class EntrySelector extends React.Component {
     paneWidth: PropTypes.string.isRequired,
     parentMutator: PropTypes.object.isRequired,
     parentResources: PropTypes.object,
+    parseInitialValues: PropTypes.func.isRequired,
     resourcePath: PropTypes.string,
     permissions: PropTypes.object,
     prohibitItemDelete: PropTypes.shape({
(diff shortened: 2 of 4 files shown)


mkuklis:
@MikeTaylor I'm a bit confused by this. In the original implementation, the `parseInitialValues` (if present) was executed in EntryWrapper. I see that you moved it to a lower level `EntrySelector` and it's currently only executing when the `resourcePath` is not present.

I'm curious why was this needed. Shouldn't this be up to the implementer to decide if they want to use `resourcePath` and if so they could omit `parseInitialValues`?

MikeTaylor:
> @MikeTaylor I'm a bit confused by this. In the original implementation, the `parseInitialValues` (if present) was executed in EntryWrapper. I see that you moved it to a lower level `EntrySelector` [...]

I don't think that's quite right. The component hierarchy in `<EntryWrapper>` is a bit odd, but EntryWrapper ends up rendering both EntrySelector go get the detail view, and within EntryForm to get the editing overlay.  EntryWrapper itself handles the parsing of initialValues for the editing form that it invokes directly; and it leaves EntrySelector to do this for the full-record display that _it_ invokes.

> and it's currently only executed when the `resourcePath` is not present.

I don't think that is right, either — at least, it is certainly not intended to be:
```
const ComponentToRender = resourcePath ? ConnectedWrapper : EntryFormComponentWithParsing;
```
sets the component either directly to EntryFormComponentWithParsing, or to ConnectedWrapper; but in the latter case, EntryFormComponentWithParsing still gets invoked (and so does the parsing) because it's passed as the `underlyingComponent` prop.

mkuklis (approved):
I think this cleared things out for me:

https://github.com/folio-org/stripes-smart-components/pull/1367/commits/c07485760d4ccd30b60fb6c4a85c42651adad7df

Thanks @MikeTaylor!

MikeTaylor:
Thanks, @mkuklis, it's good to get that spurious prop-passing out!

NikitaSedyx:
in props new objects are always created, it makes components to re-render infinitely

mkuklis:
@MikeTaylor please check what @NikitaSedyx is describing.

MikeTaylor:
I assume this is considered to be the cause of https://issues.folio.org/browse/UIHAADM-73?

MikeTaylor:
@NikitaSedyx I assume you are referring to
```
const EntryFormComponentWithParsing = ({ initialValues: iv, ...rest }) => (
      <EntryFormComponent initialValues={parseInitialValues(iv)} {...rest} />
    );
```
But then why is my code that uses this working perfectly? And why is the permissions editing working perfectly on dev1?

MikeTaylor:
It's going to be really hard to debug this if I can't reproduce it!

MikeTaylor:
I'm thinking that if @NikitaSedyx's diagnosis is correct, then `useMemo` might be the medicine.

NikitaSedyx:
> dev1

I have no idea what this env is, but I can easily reproduce it locally with fresh stripes-smart-component

> But then why is my code that uses this working perfectly

talking about this part, it's form, but Details view has issues

> then useMemo might be the medicine.

yep, memoization should help unless initialValues on top level are not re-created, but I believe they are not, otherwise we could notice these rerenders before

MikeTaylor:
New PR incoming in a shortish while.

PR #1369, opened 2023-06-22
STSMACOM-762 Fix rendering loop in <EntryManager>

MikeTaylor:
This PR fixes an intermittently manifesting rendering loop which occurs in `<EntryManager>` under some circumstances, but not in others — I have not been able to determine what the trigger is.

The fix is to memoize the creation of the components that apply parsing to the values in the record to be viewed or edited. I can't use `useMemo` for this, as the components in question are class-based rather than functional; and I couldn't get loadash's `_.memoize` to work at all. So rather than continue flogging that horse, I just did it by hand.

Fixes STSMACOM-762.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Enhance `<EntryManager>` with optional `resourcePath` prop to fetch full record. Fixes STSMACOM-757.
 * Expose `<EntryForm>` and `<EntrySelector>` components for reuse. Refs STSMACOM-759.
 * Make `<EntryManager>`s `parseInitialValues` function work with correctly when `resourcePath` is defined. Document this previously undocumented prop and its converse, `onBeforeSave`. Fixes STSMACOM-761.
+* Fix intermittent rendering loop when using `<EntryManager>` with `parseInitialValues`. Fixes STSMACOM-762.
 
 ## [8.0.0](https://github.com/folio-org/stripes-smart-components/tree/v8.0.0) (2023-01-30)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v7.3.0...v8.0.0)
--- a/lib/EntryManager/EntrySelector.js
+++ b/lib/EntryManager/EntrySelector.js
@@ -22,6 +22,12 @@ import {
 } from '@folio/stripes-components';
 import ConnectedWrapper from './ConnectedWrapper';
 
+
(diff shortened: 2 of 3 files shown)


NikitaSedyx (commented):
doesn't seem like it's right way to do

I would suggest to create a separate component that can be memoized and used in both selector and form, and instead of manual memo of `detailComponent` in render you can do it in componentDidMount/DidUpdate or even constructor - do we expect that details view component is changed?

or as another alternative you can create a HoC that will wrap with useMemo detailedComponent that can be passed to EntrySelecto/Form

@zburke @JohnC-80 @mkuklis what do you think?

MikeTaylor:
Dude, we're talking about _fixing a bug_ here.

If you have an architectural preference for a different solution, feel free to file a different PR. Otherwise I think e should merge this or not merge it based on _whether it fixes the bug_.

NikitaSedyx:
fixing bug doesn't mean that we can produce anything we want

few days you gave good advice
> The first thing to do is to get the poison out. Then we can figure out how to proceed from there.

mb it's better to revert initial changes and make them in better way? revert will fix a bug as well

MikeTaylor:
I've fixed this bug, and I have to move on to other work. I'm not going to go back and do this same work all over again to better conform to a new aesthetic preference that you have developed since [yesterday](https://github.com/folio-org/stripes-smart-components/pull/1367#issuecomment-1601980399).

NikitaSedyx:
sorry, I can't approve this fix, let's wait other members to take a look

mkuklis:
@MikeTaylor I agree with @NikitaSedyx here. I think we should take a step back and review this a bit closer again before we move forward.

If the original issue was related to `parseInitialValues` Could approach similar to this work (this is based on your original PR related to `resourcePath`:

https://github.com/folio-org/stripes-smart-components/pull/1371

MikeTaylor:
OK, I'll just fork `<EntryManager>` and maintain a working copy in my application's source tree. Feel free to revert the recent changes,

mkuklis:
@MikeTaylor have you tried what I suggested? It basically keeps everything you've done in:

https://github.com/folio-org/stripes-smart-components/pull/1361

we will just have to revert: 

https://github.com/folio-org/stripes-smart-components/pull/1367

The approach with using `parseInitialValues` inside `ConnectedWrapper` seems to be working fine.

PR #1371, opened 2023-06-23
STSMACOM-757: Parse initial values in ConnectedWrapper

mkuklis:


zburke:
Replaces #1369, we hope

PR #1373, opened 2023-06-29
ERM-2922: On changing the fields to be searched, the text string is cleared from the search box

EthanFreestone:
feat: retainInternalQueryState

Created another boolean property of SASQ, retainInternalSearchState, which when `true` allows the URL to change and only overwrite what is in internal query state when explicitly set in the URL

ERM-2922


lib/SearchAndSort/SearchAndSortQuery.js
```diff
@@ -111,6 +111,7 @@ class SearchAndSortQuery extends React.Component {
     queryGetter: PropTypes.func,
     querySetter: PropTypes.func,
     queryStateReducer: PropTypes.func,
+    retainInternalQueryState: PropTypes.bool,
```
JohnC-80, line 114: Since this only affects the `search` portion of the queryState, I suggest renaming this to `retainInternalSearchState`.

JohnC-80 (changes requested):

lib/SearchAndSort/SearchAndSortQuery.js
EthanFreestone: Oh of course! Tweaked name now

Jagusti:
Hi @JohnC-80  - is there anything else you need from us to progress the review on this?  Thanks

PR #1375, opened 2023-07-05
STSMACOM-766 - disable edit button when user doesn't have permission to edit

Terala-Priyanka:
## Purpose
User Settings Permission sets: Disable editing for users with "Setting (Users): View all settings" permission

## Approach
Disable 'Edit' button when user do not have permissions to edit.

**Note:** Added unit test to test new code. Refer to test result on screenshot 2 for the same. However, coverage seems to be zero. Coverage seems to be 100% locally.

## Screenshot
behavior with diku login
image

behavior with another user with permission "Settings (Users): Can view permission sets" **only**
image

screenshot from bigtest
image

## Refs
https://issues.folio.org/browse/STSMACOM-766


zburke (approved):
LGTM. Thanks for the note about coverage reporting. I'll take a look at the GA files and see if I can figure out what's going on there.

mariia-aloshyna (approved):

PR #1377, opened 2023-07-11
STSMACOM-770 - propagate sort fields while applying filters too

Terala-Priyanka:
## Purpose
users are not sorted by display-name

## Approach
Propagate sort fields too (along with search fields and filter parameters). This will provide correct ns values to querySetter provided via props.

NOTE: Locally coverage is showing 100%. CI script is probably buggy now. I have observed the same in couple of the earlier PRs.

## Screenshots

image

image

image

screencast - https://issues.folio.org/secure/attachment/64170/image-2023-07-11-12-04-56-189.png

## Refs
https://issues.folio.org/browse/STSMACOM-770


zburke (approved):
👏 👏 👏  small change with big results. Nice work!

PR #1382, opened 2023-07-24
STSMACOM-771: Use `escapeCqlValueAllowAsterisk` instead of `escapeCqlValue` to avoid escaping the asterisk.

Dmytro-Melnyshyn:


zburke:
I've put this on hold by moving it to "draft" while we decide how to handle the changes originally implemented folio-org/stripes-util/pull/69, but rolled back folio-org/stripes-util/pull/74.

PR #1386, opened 2023-08-21
STSMACOM-756 - ControlledVocab 'Last updated' display must be robust to sparse user data.

Terala-Priyanka:
…

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * Add `<AdvancedSearch>` to `<SearchAndSort>` component. Refs STSMACOM-767.
 * *BREAKING* Upgrade `react` to `v18`. Refs STSMACOM-769.
 * Pass `labelInfo` prop to checkboxes in `<CheckboxFilter>`. Refs STSMACOM-773.
+* ControlledVocab "Last updated" display must be robust to sparse user data. Refs STSMACOM-756.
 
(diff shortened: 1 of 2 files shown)


BogdanDenis (approved):

zburke (changes requested):
Please write a test to exercise the following cases: missing first name, missing last name, missing last and first name, missing `personal` attribute altogether. 

This change resolves the situation from the ticket's description, but the link to the original Slack conversation notes, "The UI should handle the lack of all other fields without breaking." Is this change enough?

Terala-Priyanka:
> Please write a test to exercise the following cases: missing first name, missing last name, missing last and first name, missing `personal` attribute altogether.
> 
> This change resolves the situation from the ticket's description, but the link to the original Slack conversation notes, "The UI should handle the lack of all other fields without breaking." Is this change enough?

Hey @zburke 
From the code implementation, I see the important checks points to be personal object, firstName, lastName and id.
1. absence of firstName will not cause at problem.
2. lastName can never go missing. API prohibits saving a user record when there is no lastName.
3. as obvious id will always exist.
4. On top of the above reasons, UI (users UI) prohibits user creation when lastName field is not filled.

I would add the necessary unit test - the scenario when personal object itself is missing.

Terala-Priyanka:
@zburke 
I am sure test has reached the new code. I am not sure why sonar is complaining of 0% coverage. 
As discussed I have added couple of tests. Could you please re-review?

PR #1388, opened 2023-08-23
STSMACOM-775: Search And Sort Query - fire queryStateReducer events on search submit

EthanFreestone:
feat: Add SASQ event for search submit and all submit

Added internalSetState call for submit all and submit search, which does not change internal state itself, but fires an event that queryStateReducer can use to trigger changes from those events

refs: https://issues.folio.org/browse/STSMACOM-775


EthanFreestone:
Apologies @JohnC-80 I know it's busy as we ramp up towards feature freeze, but did you have thoughts on the latest iteration of these changes? Would be good to get this merged and close out our ERM tasks

JohnC-80 (approved):

PR #1404, opened 2023-10-25
STSMACOM-785 - add default parameters to onSubmit, submitAll internal handlers

JohnC-80:
[this commit](https://github.com/folio-org/stripes-smart-components/commit/d197a0) added some necessary event handling and `changeType`s to SASQ internals, but it did not account for leftover `changeType` that may have been depended upon by app code.

If SASQ has `setQueryOnMount = true` or an `initialSearchState` applied, a `submitAll()` is carried out when the component mounts. State is initialized with an `init.reset` changeType. Previously, the submitAll() call had no `changeType` applied, so `init.reset` remained - and thus, the expectant code was built.

This code adds the `changeType` as a default parameter for the methods that we would expect, with overridden `changeType`s applied for those situational cases where `submitAll()` and `submitSearch` are called - that is, at mount time, given the previously mentioned props, and upon a user-triggered reset.

### DIFF
--- a/lib/SearchAndSort/SearchAndSortQuery.js
+++ b/lib/SearchAndSort/SearchAndSortQuery.js
@@ -249,7 +249,7 @@ class SearchAndSortQuery extends React.Component {
   componentDidMount() {
     if (this.props.setQueryOnMount ||
       (this.props.syncToLocationSearch && Object.keys(this.state.searchFields).length > 0)) {
-      this.onSubmitAll(false);
+      this.onSubmitAll(false, 'init.reset');
     }
   }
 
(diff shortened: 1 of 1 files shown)


EthanFreestone (approved):
LGTM, works with our use cases

lib/SearchAndSort/SearchAndSortQuery.js
```diff
@@ -397,7 +397,7 @@ class SearchAndSortQuery extends React.Component {
   };
 
   onResetSearch = (submit = true) => {
-    const submitCallback = submit ? () => { this.onSubmitSearch(); } : noop;
+    const submitCallback = submit ? () => { this.onSubmitSearch(false, 'search.before-reset'); } : noop;
```
zburke, line 400: Would `null` be more expressive than `false` here? Hmmm, looks like we already use `false` in other places (line 252) but this seems unfortunate since the check in `onSubmitSearch` is really evaluating an object reference rather than checking a boolean.

zburke (approved):

PR #1412, opened 2023-11-07
STSMACOM-748: Show the username in the 'last updated' accordion in the Note editing pane.

Dmytro-Melnyshyn:
## Purpose
Show the `username` in the `last updated` accordion in the Note editing pane.

## Issues
[STSMACOM-748](https://issues.folio.org/browse/STSMACOM-748)

## Screenshots
image

### DIFF
--- a/lib/Notes/utils/getMetadataFormResponse.js
+++ b/lib/Notes/utils/getMetadataFormResponse.js
@@ -4,20 +4,14 @@ const getMetadataPropsFromResponse = (responseMetadata) => {
     createdByUsername: createdBy,
     updatedByUsername,
     updatedDate: lastUpdatedDate,
-    updatedByUserId,
-    createdByUserId,
   } = responseMetadata;
-
-  const lastUpdatedBy = createdByUserId === updatedByUserId
-    ? createdBy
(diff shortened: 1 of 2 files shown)


lib/Notes/utils/getMetadataFormResponse.js
```diff
@@ -2,10 +2,16 @@ const getMetadataPropsFromResponse = (responseMetadata) => {
...
     createdDate,
     createdByUsername: createdBy,
+    updatedByUsername,
     updatedDate: lastUpdatedDate,
-    updatedByUserId: lastUpdatedBy
+    updatedByUserId,
+    createdByUserId,
   } = responseMetadata;
 
+  const lastUpdatedBy = createdByUserId === updatedByUserId
```
BogdanDenis, line 11: I think we can just use `updatedByUsername` in both cases.
If ids are equal - we're passing `createdBy`, which is the same as `updatedByUsername`
If they are not - we're passing `updatedByUsername`

So we can just remove the conditional

BogdanDenis (changes requested):

BogdanDenis (approved):

PR #1421, opened 2023-12-22
STSMACOM-793: Use the default match and search option in Advanced search when they are not entered.

Dmytro-Melnyshyn:
## Purpose
Doesn't show the `startsWith` match option when a user selects the Advanced search option and doesn't enter a match and search option ([screencast](https://github.com/folio-org/stripes-components/assets/77053927/55a314c9-7395-43ac-82f5-68fd5019651b)).

## Related PRs
https://github.com/folio-org/stripes-components/pull/2189/files

## Issues
[STSMACOM-793](https://issues.folio.org/browse/STSMACOM-793)

## Screencast

https://github.com/folio-org/stripes-smart-components/assets/77053927/ed434b6a-b57d-49a6-9bcb-93a47b5802cc


lib/SearchAndSort/tests/advancedSearchQueryToRows-test.js
```diff
@@ -38,4 +38,17 @@ describe('advancedSearchQueryToRows', () => {
       });
     });
   });
+
+  describe('when query without match and search option', () => {
+    it('should parse query correctly', () => {
```
Dmitriy-Litvinenko, line 43: Could you please clarify should we rename to `should use default value for match and search option` or something like that?

Terala-Priyanka (commented):
I agree with Dmitriy's comment.

BogdanDenis (approved):

mariia-aloshyna (approved):

PR #1430, opened 2024-01-16
STSMACOM-800 Add field type `DATE_PICKER` to custom fields components.

alb3rtino:
See [STSMACOM-800](https://issues.folio.org/browse/STSMACOM-800).
UI work for [FCFIELDS-46](https://issues.folio.org/browse/FCFIELDS-46).

This PR introduces a new field type for custom fields, `DATE_PICKER`, designed to handle ISO 8601 date values. Custom fields UI components have been updated to utilize this new field type. To add and edit date values the `<Datepicker>` component from `stripes-components` is used. Tests have been updated and translations added.

The required backend work has been done in [folio-custom-fields PR #42](https://github.com/folio-org/folio-custom-fields/pull/42).

[screen-capture.webm](https://github.com/folio-org/stripes-smart-components/assets/25431636/021da625-655f-417e-bde4-9d28ed886e46)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Use the default match and search option in Advanced search when they are not entered. Refs STSMACOM-793.
 * Show successful toast notifications for Create and Edit actions in `<ControlledVocab>`. Refs STSMACOM-796.
 * `<ControlledVocab>` - last updated by column - show "System" when items are created by system user. Refs STSMACOM-797.
+* Add field type `DATE_PICKER` to custom fields components. Refs FCFIELDS-46.
 
 ## [9.0.0](https://github.com/folio-org/stripes-smart-components/tree/v9.0.0) (2023-10-11)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v8.0.0...v9.0.0)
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * Use the default match and search option in Advanced search when they are not entered. Refs STSMACOM-793.
 * Show successful toast notifications for Create and Edit actions in `<ControlledVocab>`. Refs STSMACOM-796.
 * `<ControlledVocab>` - last updated by column - show "System" when items are created by system user. Refs STSMACOM-797.
+* Add field type `DATE_PICKER` to custom fields components. Refs FCFIELDS-46.
```
BogdanDenis, line 17: Please move the original Jira ticket to STSMACOM project and update the ticket number here
No need to re-create the PR, after merging the branch should be deleted anyways
We only reference tickets from corresponding Jira projects in repositories so we can correctly track release versions.
The next released version of `stripes-smart-components` will be `9.1.0` or `10.0.0`, but FCFIELDS project might have a different next release version, and there could be a lot of confusion on what has been released, where and when

BogdanDenis (changes requested):

CHANGELOG.md
alb3rtino: Done.

artem-blazhko (approved):

PR #1438, opened 2024-02-02
[STSMACOM-806] Add `onComponentLoad` prop to `<EditCustomFieldsRecord>`

alb3rtino:
See https://issues.folio.org/browse/STSMACOM-806.

Adds a `onComponentLoad` prop to `<EditCustomFieldsRecord>` that takes a callback function.
This will allow parent components to know when the rendering of form fields is complete.
`<EditCustomFieldsRecord>` renders its `<Field>` components asynchronously after data was fetched from the `/custom-fields` API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * Upgrade `stylelint` and associated dependencies. Refs STSMACOM-803.
 * `<UserName>` must handle sparse data. Refs STSMACOM-802.
 * `<EditableList>` - added new `getReadOnlyFieldsForItem` prop to control read only fields for different items. Refs STSMACOM-801.
+* Add `onComponentLoad` prop to `<EditCustomFieldsRecord>`. Refs STSMACOM-806.
 
 ## [9.0.0](https://github.com/folio-org/stripes-smart-components/tree/v9.0.0) (2023-10-11)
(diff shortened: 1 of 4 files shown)


BogdanDenis (approved):

BogdanDenis:
Hey @alb3rtino, I don't see an issue with adding this callback, but I don't understand how a data could be manipulated from parent before custom fields is loaded. Could you give an example of this?

alb3rtino:
> Hey @alb3rtino, I don't see an issue with adding this callback, but I don't understand how data could be manipulated from parent before custom fields is loaded. Could you give an example of this?

Please see [this PR comment](https://github.com/folio-org/ui-orders/pull/1542#discussion_r1475062228) where a `<EditCustomFieldsRecord>` was [added](https://github.com/folio-org/ui-orders/blob/4c58a9cf9843ebd534813dd3ef7e90d4ce6cf36f/src/components/POLine/POLineForm.js#L610-L617). Here [populateFieldsFromTemplate()](https://github.com/folio-org/ui-orders/blob/4c58a9cf9843ebd534813dd3ef7e90d4ce6cf36f/src/components/POLine/POLineForm.js#L190-L206) may be called before the custom fields have been loaded and as a result the custom field form values don't get populated.

My idea is to wait for the custom fields to properly load before attempting to populate them. Thats what the callback is about. If you have an alternative solution for resolving this issue, I would be happy to hear about it.

ryandberger (approved):

PR #1441, opened 2024-02-06
STSMACOM-808: `<ItemEdit>` - pass the `error` to the `fieldComponents` callback.

Dmytro-Melnyshyn:
## Purpose
 `<ItemEdit>` - pass the `error` to the `fieldComponents` callback.
## Issues
[STSMACOM-808](https://folio-org.atlassian.net/browse/STSMACOM-808)

### DIFF
--- a/lib/EditableList/ItemEdit.js
+++ b/lib/EditableList/ItemEdit.js
@@ -50,7 +50,7 @@ const ItemEdit = ({
       if (Object.hasOwnProperty.call(fieldComponents, name)) {
         return (
           <div key={fieldKey} style={fieldStyle}>
-            { fieldComponents[name]({ fieldProps, fieldIndex, name, field, rowIndex }) }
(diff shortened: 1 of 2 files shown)


lib/EditableList/ItemEdit.js
```diff
@@ -50,7 +50,7 @@ const ItemEdit = ({
       if (Object.hasOwnProperty.call(fieldComponents, name)) {
         return (
           <div key={fieldKey} style={fieldStyle}>
-            { fieldComponents[name]({ fieldProps, fieldIndex, name, field, rowIndex }) }
+            { fieldComponents[name]({ fieldProps, fieldIndex, name, field, rowIndex, error }) }
```
BogdanDenis, line 53: Is there a related PR for this change? This passed object doesn't match how it's passed below
It should be added to `fieldProps`
```
error: error?.fieldErrors?.[name]
```

PR #1445, opened 2024-02-13
STSMACOM-809 EditableList - keep changed values when request fails, but clear them when clicking cancel (follow-up)

BogdanDenis:
## Description
EditableList - keep changed values when request fails, but clear them when clicking cancel
Previously I added `keepDirtyOnReinitialize` to fix form reset when requests failed, but it caused the cancel button not resetting edited values to initial values

Turns out the root cause was constantly changing `initialValues` which in turn causes form reinitialize on each render
Now `initialValues` are memoized

## Screenshots

https://github.com/folio-org/stripes-smart-components/assets/19309423/481e5597-da06-4ab4-9881-29c642f3899e

## Issues
[STSMACOM-809](https://folio-org.atlassian.net/browse/STSMACOM-809)

### DIFF
--- a/lib/EditableList/EditableList.js
+++ b/lib/EditableList/EditableList.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import sortBy from 'lodash/sortBy';
 import EditableListFinalForm from './EditableListFinalForm';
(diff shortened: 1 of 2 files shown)


Dmytro-Melnyshyn (approved):

mariia-aloshyna:
Hi @BogdanDenis it looks like because of these changes the validation stopped working and we have failed tests for this [PR](https://github.com/folio-org/ui-data-import/pull/1542)

PR #1452, opened 2024-03-11
STSMACOM-816 Safely render user-supplied markup in Notes

JohnC-80:
https://folio-org.atlassian.net/browse/STSMACOM-816

Makes use of a new component: `<CleanHTML>` which can be used in place of any element using React's `_dangerouslySetInnerHTML`

Test included. (but will fail before https://github.com/folio-org/stripes-components/pull/2239 is merged)


JohnC-80:
Closing in favor of the `dompurify` solution in https://github.com/folio-org/stripes-smart-components/pull/1453

PR #1453, opened 2024-03-11
STSMACOM-816 Safely render HTML in Note template

JohnC-80:
This approach makes use of the `dompurify` library. It requires less configuration out of the box than the previously [attempted](https://github.com/folio-org/stripes-smart-components/pull/1452) `sanitize-html`


zburke (approved):
LGTM

PR #1458, opened 2024-03-15
STSMACOM-820 Fix incorrect state calculation in <SearchAndSortQuery>

BogdanDenis:
## Description
There is a bug in `<SearchAndSortQuery>` that resets filter state when user selects a filter value from initial values.

https://github.com/folio-org/stripes-smart-components/assets/19309423/2746e8c0-6223-4307-bd74-899011d6002e

The root cause is incorrect order of operations during component state update and a little bit of state mutation
### Incorrect order
Prevously we had this bit of code to get next state:
```
nextState.changeType = changeType;
nextState.searchChanged = !isEqual(
  nextState.searchFields,
  curState.default.searchFields
);
nextState.filterChanged = !isEqual(
  nextState.filterFields,
  curState.default.filterFields
);
nextState.sortChanged = !isEqual(nextState.sortFields, curState.default.sortFields);
return queryStateReducer(curState, nextState);
```
The issue is that in `isEqual(nextState.filterFields, curState.default.filterFields)` `nextState.filterFields` is not a complete object of filter values after update. It's just an object with properties to add to state, so we cannot know if next state is the same as default state just from this object.
For that we need to calculate the full next state, and only after that can we compare it to default state.
We can call `this.setState` twice - first call to calculate selected filter and other values, and then the first call will take twose values and compare them to defaults to set flags

### State mutation
Related https://github.com/folio-org/ui-plugin-find-instance/pull/212 fixes a state mutation issue and uses `nextState.filterFields` instead of `nextState.filterChanged` for it's logic. More details in comments there

## Screenshots

https://github.com/folio-org/stripes-smart-components/assets/19309423/f39bcc80-8992-4831-9f99-db6c6d48abdc

## Issues
[STSMACOM-820](https://folio-org.atlassian.net/browse/STSMACOM-820)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 9.2.0 IN PROGRESS
 
+## [9.1.1] (IN PROGRESS)
+
+* Fix incorrect state calculation in `<SearchAndSortQuery>`. Fixes STSMACOM-820.
+
 ## [9.1.0](https://github.com/folio-org/stripes-smart-components/tree/v9.1.0) (2024-03-13)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v9.0.1...v9.1.0)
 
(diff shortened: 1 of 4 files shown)


mariia-aloshyna (approved):

Dmytro-Melnyshyn (approved):

zburke (approved):
LGTM, but you and @JohnC-80 are way more expert than I am when it comes to SASQ. Thanks so much for the detailed RCA in the PR description, and the comments explaining how/why this fix works ❤️ . I would not have been able to assess this without you walking me through it.

PR #1459, opened 2024-03-19
STSMACOM-822: Do not trigger logic for auto-opening the record's view screen in `<SearchAndSort>` if URL already contains the record ID.

Dmytro-Melnyshyn:
## Description
When there is a single search result and the third pane is open, and then the user creates a new record in a different route and when the new record is created the user is redirected to the main page (with the previously opened third pane). The third pane doesn't contain the newly created record, it contains the previously opened one. [video record](https://github.com/folio-org/stripes-smart-components/assets/77053927/5528549c-b886-467d-9262-179dcc033769) 

Root cause: 
The logic for automatically opening the record’s view screen fires even if the URL already contains the record’s view screen route. The logic must only be triggered if there is no open record.

## Approach

Add a check for a record ID in the URL. If it exists, then prevent the automatic opening of the record.

## Issues
[STSMACOM-822](https://folio-org.atlassian.net/browse/STSMACOM-822)

## Screencast

https://github.com/folio-org/stripes-smart-components/assets/77053927/56a3e40a-8db7-44e0-b22b-89599b5b2233

### DIFF
--- a/lib/SearchAndSort/SearchAndSort.js
+++ b/lib/SearchAndSort/SearchAndSort.js
@@ -411,6 +411,9 @@ class SearchAndSort extends React.Component {
     const totalCountResult = currentState.totalCount();
     if (showSingleResult &&
       totalCountResult === 1 &&
+      // Check for the absence of record ID in the URL to prevent the record from being automatically opened when
+      // a user navigates from another route to `../view/:id`. The record will be open in any case.
+      // Otherwise, the previously open record replaces the newly created one.
       !recordId &&
       (this.lastNonNullResultCount > 1 || (previousState.records()[0] !== currentState.records()[0]))) {
       this.onSelectRow(null, currentState.records()[0]);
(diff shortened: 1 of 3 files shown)


BogdanDenis (approved):

BogdanDenis:
Please add tests for this case

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -398,16 +398,20 @@ class SearchAndSort extends React.Component {
...
     const currentState = makeConnectedSource(this.props, logger);
 
+    const recordId = location.pathname.split('/')[3];
+
     // if the results list is winnowed down to a single record, display the record.
     // If there's a single hit, open the detail pane as side-effect.
     const totalCountResult = currentState.totalCount();
     if (showSingleResult &&
       totalCountResult === 1 &&
+      !recordId &&
```
zburke, line 414: Technically, this looks like the correct change here, but you must include some comments to explain what's going on and why.

zburke (approved):
This needs ~better~ documentation _in the code_, not just in the PR since it isn't at all obvious why the `!recordId` condition is present. Explain that magic number, too :) 

Once comments are in place, this LGTM.

lib/SearchAndSort/SearchAndSort.js
Dmytro-Melnyshyn: Yes, that will make it clearer, thanks @zburke!

PR #1464, opened 2024-04-01
STSMACOM-817 - Create Reusable component for profile pictures.

manvendra-s-rathore:
[STSMACOM-817](https://folio-org.atlassian.net/browse/STSMACOM-817) - Create Reusable component for profile pictures.

## Purpose 
Create a re usable component for profile pictures which will be benefited for users, check out and reading room screens etc..


lib/ProfilePicture/ProfilePicture.js
```diff
@@ -0,0 +1,46 @@
+import { useIntl } from 'react-intl';
+import PropTypes from 'prop-types';
+import { Img } from 'react-image';
+
+import { Loading } from '@folio/stripes/components';
```
JohnC-80, line 5: Since this library is part of the `@folio/stripes` collection, it can import from `@folio/stripes-components` directly instead of the path through the `stripes` package.

lib/ProfilePicture/utils/useProfilePicture.js
```diff
@@ -0,0 +1,36 @@
+import { useQuery } from 'react-query';
+
+import {
+  useNamespace,
+  useOkapiKy,
+} from '@folio/stripes/core';
```
JohnC-80, line 6: Similar to before, this can import from `@folio/stripes-core` directly.

JohnC-80 (changes requested):

lib/ProfilePicture/ProfilePicture.js
```diff
@@ -0,0 +1,47 @@
+import { useIntl } from 'react-intl';
+import PropTypes from 'prop-types';
+import { Img } from 'react-image';
+
+import { Loading } from '@folio/stripes-components';
+
+import { isAValidURL } from './utils/util';
+import useProfilePicture from './utils';
+import profilePicThumbnail from './icons/profilePicThumbnail.png';
```
zburke, line 9: Would `placeholderThumbnail` be a better name for this? That might make it easier to distinguish between the two conditions (user data is present vs user data is missing).

lib/ProfilePicture/ProfilePicture.js
```diff
@@ -0,0 +1,47 @@
...
+
+import css from './ProfilePicture.css';
+
+const ProfilePicture = ({ profilePictureLink, croppedLocalImage }) => {
+  const intl = useIntl();
+  const { isFetching, profilePictureData } = useProfilePicture({ profilePictureId: profilePictureLink });
```
zburke, line 15: Based on the comments below, it seems `profilePictureLink` (which you refer to as "Profile Picture Link" on line 18, I think?) is not necessarily a link, right? Then don't put `Link` in its name. 

The logic here seems reversed. If we have a link, we don't need the hook to query for the data, do we? Can we check that first, moving the logic up from line 23 before line 15, and then pass a `disabled` prop to the hook based on whether it's a valid URL?

zburke (changes requested):
Sorry, I found this code really hard to follow, and there are straight up errors in `package.json`: dependencies are in the wrong place and the necessary interface is missing as well as missing interface-checks in the hook. 

IIUC, you have two main conditions: you have user data (show it) or you don't (show a placeholder), and you have some nested conditions when you have user data: you get a URL in `croppedLocalImage`, or you get a URL in `profilePictureLink`, or you get an identifier in `profilePictureLink` and use the hook to retrieve a blob. Try to structure the logic this way. Likewise the tests. It would be nice if it were possible for the tests to distinguish "got a URL and showed it" from "got binary data and showed it" from "got nothing and showed a placeholder".

lib/ProfilePicture/ProfilePicture.js
manvendra-s-rathore: Hey Zak, renamed profilePictureLink to profilePictureSource and moved the logic of validating whether whether profilePictureLink is URL or not before the useProfilePicture hook, created a **isProfilePictureAUUID** prop while replaces the isAValidUUID utility function inside useProfilePicture hook.

manvendra-s-rathore: Hey Zak, renamed profilePictureLink to **profilePictureSource** and shifted the logic of validating whether profilePictureLink is URL or not before the useProfilePicture hook.
Created **isProfilePictureAUUID** prop which replaces the isAValidUUID utility function inside useProfilePicture hook.

manvendra-s-rathore:
Hey @folio-org/fe-tl-reviewers , can someone please review this PR.

Terala-Priyanka (approved):

manvendra-s-rathore:
@zburke @folio-org/fe-tl-reviewers, I'm merging this PR due to its significant dependencies. Any review comments or suggested changes can be addressed in a separate PR.

gurleenkaurbp (approved):

PR #1472, opened 2024-04-23
STSMACOM-831 `<AdvancedSearch>` Improve algorithm of splitting query string into rows

BogdanDenis:
## Description
`<AdvancedSearch>` Improve algorithm of splitting query string into rows

## Screenshots

https://github.com/folio-org/stripes-smart-components/assets/19309423/793f9336-f0dc-425d-902a-4dfb0a4c0292

## Issues
[STSMACOM-831](https://folio-org.atlassian.net/browse/STSMACOM-831)

### DIFF
--- a/lib/SearchAndSort/advancedSearchQueryToRows.js
+++ b/lib/SearchAndSort/advancedSearchQueryToRows.js
@@ -1,6 +1,6 @@
 import {
+  BOOLEAN_OPERATORS,
   MATCH_OPTIONS,
-  DEFAULT_SEARCH_OPTION,
 } from '@folio/stripes-components/lib/AdvancedSearch';
 
 const advancedSearchQueryToRows = (queryValue) => {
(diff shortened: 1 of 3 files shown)


lib/SearchAndSort/tests/advancedSearchQueryToRows-test.js
```diff
@@ -39,15 +39,41 @@ describe('advancedSearchQueryToRows', () => {
...
+  describe('when query contains boolean option and match option as part of value, but no search option', () => {
+    it('should parse query correctly', () => {
+      const query = 'keyword containsAny some text and containsAny';
 
       expect(advancedSearchQueryToRows(query)).to.deep.include({
-        query,
+        query: 'some text and containsAny',
         bool: '',
         searchOption: 'keyword',
-        match: 'containsAll',
```
Dmytro-Melnyshyn, line 50: I think that when a user selects the `Advanced search` search option, then enters any value and then opens the Advances Search modal, there should be the user input in the modal's query field.

BogdanDenis: Do you mean like this?

https://github.com/folio-org/stripes-smart-components/assets/19309423/190c443b-beba-446c-b9eb-220c123efb47

lib/SearchAndSort/advancedSearchQueryToRows.js
```diff
@@ -1,43 +1,67 @@
...
+  */
+
+  const tokens = queryValue.replace(/\s+/g, ' ').split(' ');
 
-  // split will return array of strings:
-  // ['keyword==test', 'or issn=123', ...]
-  const matches = queryValue.split(splitIntoRowsRegex).map(i => i.trim());
+  const rows = [];
+  for (let index = 0; index <= tokens.length;) {
+    const token1 = tokens[index - 1];
```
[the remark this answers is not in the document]
BogdanDenis: Yes, query string begins with search option and doesn't have boolean as a first token:
`title exactPhrase some text`
In this case `tokens[-1]` will be `undefined`, which is handled inside `if`:
`if ((isFirstTokenBoolean || !token1) && isThirdTokenMatch)`

artem-blazhko (approved):

PR #1481, opened 2024-05-15
STSMACOM-837 improve AdvancedSearch parsing algorithm to keep repeated spaces in queries

BogdanDenis:
## Description
Another improvement/fix of Advanced Search's algorithm to split a query string into array of rows
This one is to fix an issue where repeated spaces in a query value would be replaced with just one space
This was happening when splitting the string into tokens:
```
const tokens = queryValue.replace(/\s+/g, ' ').split(' ');
```
We really need to get rid of duplicate spaces when splitting the string into tokens (explanation in code comments)
But what we can do is after we've split the string and we know exactly what type each token is (boolean, search option, match option or a query) we can restore those spaces that we've lost in queries

## Screenshots

https://github.com/folio-org/stripes-smart-components/assets/19309423/28df7ef1-13c5-4c80-ae1b-97ba62a1bd40

## Issues
[STSMACOM-837](https://folio-org.atlassian.net/browse/STSMACOM-837)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * `<ControlledVocab>` - pass override headers to PUT method, new prop to hide "New" button. Refs STSMACOM-825.
 * `<SearchAndSort>` - re-position Advanced search button when search panel does not have enough space. Refs STSMACOM-830.
 * Extend `ViewMetadata` component to accept a child render function for custom rendering. Refs STSMACOM-828.
+* Improve AdvancedSearch parsing algorithm to keep repeated spaces in queries. Fixes STSMACOM-837.
 
 ## [9.1.1] (IN PROGRESS)
 
--- a/lib/SearchAndSort/advancedSearchQueryToRows.js
+++ b/lib/SearchAndSort/advancedSearchQueryToRows.js
@@ -10,56 +10,162 @@ const advancedSearchQueryToRows = (queryValue) => {
   }
 
   /*
-    Algorithm for parsing string into rows is as follows:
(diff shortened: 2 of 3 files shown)


UladzislauKutarkin (approved):

zburke (approved):
Ooof, string parsing 🤯 . Thanks for the super helpful comments and examples! Next up, escape sequences so we can correctly parse a search string about SEO like "keyword contains keyword"? 🙃

BogdanDenis:
That's a good point @zburke, I was sure that `keyword containsAll keyword` will consider the second `keyword` as a query (which is expected imo) but I added that case anyways

zburke:
I expect this will be an algorithm that gets slowly tweaked and improved over time. The hairiest part of string parsing is often figuring out what to do when you encounter a "special" character/character-sequence: do you treat it as special and perform substitution or tokenization, or is there some escape sequence that makes it un-special? 

There will always be titles like `optimizing keyword searching and SEO` or `the hobbit or there and back again` or `charlotte's web` that wreak havoc on parsing. It may be more important to know where the edges of the algorithm are than to try to cover all sequences, e.g. if you want to use keywords as values like `keyword containsAll keyword` then you have to use a query-builder that forces your query into subject/predicate/value boxes instead of trying to parse those pieces out of freeform text. 

Trying to make a string-parser that covers all cases is a fool's errand, but next steps like implementing a CFG are big steps with their own downsides. My point is just: try to keep perspective as you work on this because it's easy to get lost in the weeds. String parsing is all weeds 😆 Maybe what you have is "good enough" and you can ignore some of the annoying edge cases here because there are other avenues to handle them.

🤝

PR #1486, opened 2024-06-20
STSMACOM-778 - Supply boolean value to enabled option of useQuery

Terala-Priyanka:
[STSMACOM-778](https://folio-org.atlassian.net/browse/STSMACOM-778) - useRemoteStorageMappings doesn't validate the presence of interfaces it relies on

When "remote-storage-mappings" is not available, stripes.hasInterface returns undefined which is not being consumed by useQuery enable option as false. Hence it has to be "withRemoteStorage" has to be booleanized.
With this change, ui-inventory will not make calls to remote storage mappings API when this optional interface is not available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Extend `ViewMetadata` component to accept a child render function for custom rendering. Refs STSMACOM-828.
 * Improve AdvancedSearch parsing algorithm to keep repeated spaces in queries. Fixes STSMACOM-837.
 * Support Optimistic Locking in Tags. Refs STSMACOM-839.
+* Supply boolean value to enabled option of useQuery. STSMACOM-778.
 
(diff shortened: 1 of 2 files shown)


artem-blazhko (approved):

zburke (approved):
🤯 IIUC, we derived that the internal implementation of useQuery's `enabled` uses strict equality, er, tests for strictly false like `=== false` instead of loosely checking for falseyness. Here's the bug for folks who want to play along at home: 
```
const foo = undefined;
if (foo === false) { 
  console.log('foo is false'); 
} else { 
  console.log('foo is true'); 
}
```
I understand why folks like strict equality, but in the case of boolean comparison I think it is the wrong choice. 

Nice work troubleshooting this, @Terala-Priyanka! It's amazing that such a complex problem can have since a simple solution.

manvendra-s-rathore (approved):

PR #1496, opened 2024-07-23
STSMACOM-846: Add optional `isRequestUrlExceededLimit` property to `SearchAndSort` to return a specific error message in `StripesConnectedSource`.

Dmytro-Melnyshyn:
## Purpose:
Display a specific error message in the results pane when the request URL exceeds the limit.

## Approach
Add the `isRequestUrlExceededLimit` property to indicate whether the limit is exceeded.

## Issues
[STSMACOM-846](https://folio-org.atlassian.net/browse/STSMACOM-846)

## Screencast

https://github.com/user-attachments/assets/6a60bced-cbc3-4d86-b78e-efe99428f664

### DIFF
--- a/lib/SearchAndSort/requestUrlLimit.js
+++ b/lib/SearchAndSort/requestUrlLimit.js
@@ -1,0 +1,1 @@
+export const REQUEST_URL_LIMIT = 4121;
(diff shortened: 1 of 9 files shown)


lib/SearchAndSort/requestUrlLimit.js
```diff
@@ -0,0 +1 @@
+export const REQUEST_URL_LIMIT = 4121;
```
BogdanDenis, line 1: Is this value documented somewhere in Chrome docs?

Dmytro-Melnyshyn: This is a server limitation according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414). It also causes this error with the same length in Firefox. I took this number from the request URL length.

OleksandrHladchenko1 (approved):

PR #1512, opened 2024-09-09
STSMACOM-853 DateRangeFilter - update all calls of `validateDateRange`

BogdanDenis:
## Description
After [STSMACOM-838](https://folio-org.atlassian.net/browse/STSMACOM-838) validateDateRange function now accepts requiredFields argument, which should be passed to all calls

## Issues
[STSMACOM-853](https://folio-org.atlassian.net/browse/STSMACOM-853)

## Related PRs
https://github.com/folio-org/stripes-inventory-components/pull/74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 * Add `showSortIndicator` prop to `SearchAndSort` to display sort indicator next to the column names. STSMACOM-849.
 * Trim value to validate required field in the `ControlledVocab` component. STSMACOM-850.
 * Return a specific error message in `StripesConnectedSource`. STSMACOM-851.
-* DateRangeFilter - update all calls of `validateDateRange`. STSMACOM-853.
+* DateRangeFilter - pass `requiredFields` argument to all calls of `validateDateRange`. STSMACOM-853.
 
 ## [9.1.1] (IN PROGRESS)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * Add `showSortIndicator` prop to `SearchAndSort` to display sort indicator next to the column names. STSMACOM-849.
 * Trim value to validate required field in the `ControlledVocab` component. STSMACOM-850.
 * Return a specific error message in `StripesConnectedSource`. STSMACOM-851.
+* DateRangeFilter - update all calls of `validateDateRange`. STSMACOM-853.
 
 ## [9.1.1] (IN PROGRESS)
 
--- a/lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
+++ b/lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
@@ -50,7 +50,7 @@ const InvalidOrderError = () => (
   </span>
 );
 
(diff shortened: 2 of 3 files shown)


lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
```diff
@@ -170,10 +170,10 @@ const TheComponent = ({
 
   React.useEffect(() => {
     setFilterValue({
-      ...getInitialValidationData(selectedValues, dateFormat),
+      ...getInitialValidationData(selectedValues, dateFormat, requiredFields),
       selectedValues
     });
-  }, [selectedValues, dateFormat]);
+  }, [selectedValues, dateFormat, requiredFields]);
```
Dmytro-Melnyshyn, line 176: Let's put requiredFields.start and requiredFields.end into deps.

BogdanDenis: This is an array, so we can just add itself

Dmytro-Melnyshyn: We can also destructure the array.
```
const [startDate, endDate] = requiredFields;

useEffect(() => {
  ...
}, [selectedValues, dateFormat, startDate, endDate]);
```

BogdanDenis: In this case we will have to pass `startDate` and `endDate` as a custom array like this too:
```
useEffect(() => {
  setFilterValue({
      ...getInitialValidationData(selectedValues, dateFormat, [startDate, endDate]),
      selectedValues
    });
}, [selectedValues, dateFormat, startDate, endDate]);
```

I don't like passing it like this because we already have a `requiredFields` prop that we can use, and desctructuring it, giving new names to items and then passing the same array in this way makes it too confusing

Dmytro-Melnyshyn (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * Add `showSortIndicator` prop to `SearchAndSort` to display sort indicator next to the column names. STSMACOM-849.
 * Trim value to validate required field in the `ControlledVocab` component. STSMACOM-850.
 * Return a specific error message in `StripesConnectedSource`. STSMACOM-851.
+* DateRangeFilter - update all calls of `validateDateRange`. STSMACOM-853.
```
zburke, line 24: Update them how? This comment amounts to "change function". We don't need paragraphs and footnotes, but enough context so we don't _have_ to read the code to grok the change would be helpful.

zburke (approved):
Code changes LGTM. 

It would be nice to see a test showing that the updated functions call `validateDateRange()` with the expected arguments, i.e. with/without `requiredFields`.

mariia-aloshyna (approved):

BogdanDenis:
Hey @zburke thanks for review. I am finding it difficult to test this specific case.
It's happening when `selectedValues` prop changes, so internal state needs to be re-calculated. And this calculation previously was missing a `requiredFields` argument.
I don't know if we can re-render a component with different props in BigTest

PR #1514, opened 2024-09-12
Fix the reinitialization of useEffect dependency when the component re-renders

usavkov-epam:
After merging https://github.com/folio-org/stripes-smart-components/pull/1512 some of the tests started failing due to infinite calling of the `useEffect` inside the `<DateRangeFilter>` component.

image

The absence of explicit passing of the `requiredFields` prop causes the application of the default value, which is an array literal. Accordingly, with each render of the component, this array is created anew and the reference to this object in memory changes, which serves as an indicator for calling the `useEffect` callback.

### DIFF
--- a/lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
+++ b/lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
@@ -75,6 +75,8 @@ const arePropsEqual = (prev, next) => {
   );
 };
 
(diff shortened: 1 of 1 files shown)


alisher-epam (approved):

SerhiiNosko (approved):

zburke:
Thanks, @usavkov-epam, for the clear explanation of the bug and how this change resolves it!

PR #1517, opened 2024-09-13
STSMACOM-855: `DateRangeFilter` - add the optional `hideCalendarButton` property to hide the calendar icon button; add error message for invalid `YYYY` format.

Dmytro-Melnyshyn:
## Purpose
- add the ability to hide the calendar icon button;
- show a different error message if format is YYYY and the year is invalid.

## Approach
- the optional `hideCalendarButton` property to hide the calendar icon button;
- add an error message specificly for invalid `YYYY` format.

## Related PRs
https://github.com/folio-org/stripes-components/pull/2347

## Issues
[STSMACOM-855](https://folio-org.atlassian.net/browse/STSMACOM-855)

### DIFF
--- a/lib/SearchAndSort/components/DateRangeFilter/tests/DateRangeFilter-test.js
+++ b/lib/SearchAndSort/components/DateRangeFilter/tests/DateRangeFilter-test.js
@@ -14,6 +14,7 @@ import DATE_TYPES from '../date-types';
 const filterName = 'dateRange';
 const dateFormat = 'YYYY-MM-DD';
 const invalidMessage = 'Please enter a valid date';
+const invalidYearMessage = 'Please enter a valid year';
 const missingStartDateMessage = 'Please enter a start date';
 const missingEndDateMessage = 'Please enter an end date';
 const wrongOrderMessage = 'Start date is greater than end date';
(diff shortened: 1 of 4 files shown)


lib/SearchAndSort/components/DateRangeFilter/tests/DateRangeFilter-test.js
```diff
@@ -203,6 +204,22 @@ describe('DateRangeFilter', () => {
...
   });
 
+  describe('when YYYY format is used and the provided year is invalid', () => {
+    setupApplication({
+      component: (
+        <DateRangeFilterHarness
+          dateFormat="YYYY"
+          selectedDates={{
+            startDate: '',
+            endDate: '200a',
```
BogdanDenis, line 214: Instance dates could have non-digit characters inside them, so in our case the component should support dates like `199u` which represent years 1990-1999
Maybe we could add a new prop for validating a single date and use it inside `validateDateRange` util.

BogdanDenis (changes requested):

lib/SearchAndSort/components/DateRangeFilter/tests/DateRangeFilter-test.js
Dmytro-Melnyshyn: We only need to support digits.

BogdanDenis (approved):

mariia-aloshyna (approved):

PR #1521, opened 2024-10-04
STSMACOM-858: Display `Unknown user` when the creator was deleted (follow-up)

OleksandrHladchenko1:
[STSMACOM-858](https://folio-org.atlassian.net/browse/STSMACOM-858)

* When user created reference data and after that the user was deleted, there is no personal information. Display `updated by Unknown user` instead.


lib/ControlledVocab/ControlledVocab.js
```diff
@@ -478,6 +478,8 @@ class ControlledVocab extends React.Component {
       user = <Link to={`/users/view/${metadata.updatedByUserId}`}>{name}</Link>;
     } else if (metadata.updatedByUserId === SYSTEM_USER_ID || !metadata.updatedByUserId) {
       user = <FormattedMessage id="stripes-smart-components.system" />;
+    } else {
+      user = <FormattedMessage id="stripes-smart-components.unknownUser" />;
```
zburke, line 482: Alternatively, grab this from stcom instead: `stripes-components.metaSection.unknownUser` and then we don't need a new translation.

zburke (approved):

Dmytro-Melnyshyn (approved):

PR #1532, opened 2024-10-28
UITEST-870: Rename the `ui-notes.item.assign-unassign` permission.

Dmytro-Melnyshyn:
## Purpose
Rename the permission to meet https://github.com/folio-org/ui-notes/pull/253

## Related PRs
https://github.com/folio-org/ui-notes/pull/253

## Issues
[STSMACOM-870](https://folio-org.atlassian.net/browse/STSMACOM-870)


BogdanDenis:
Hey @zburke when renaming permissions we found that some of them are used in other modules, like `stripes-smart-components`. Renaming them would mean a breaking change because in this case a permission is a part of module's public API. Does this mean that we can't include these tickets until the next release of stripes?

zburke:
@BogdanDenis, your analysis sounds correct to me. The next (Sunflower) release of stripes will be a breaking change, so that's a good time to make those changes. 

For Ramsons, could you duplicate the permission instead of renaming it? I will check in with the Eureka folks who requested all these permission changes and see if that will work on their end, too.

BogdanDenis:
> For Ramsons, could you duplicate the permission instead of renaming it? I will check in with the Eureka folks who requested all these permission changes and see if that will work on their end, too.

Do you mean by this that we just copy the permissions but with a new name, or that we also should update places where we check for these permissions and check that a user has either the old one or the copy?

BogdanDenis (approved):

mariia-aloshyna (approved):

zburke:
@Dmytro-Melnyshyn , due to these permissions, ui-notes is effectively an optional, undocumented peer-dependency of stripes-smart-components. That's problematic for many reasons, but the particular problem here is that the new permissions names are only present in a new release of ui-notes, which means this change is equivalent to bumping the minimum version of a peer, which means this is a breaking change for stripes-smart-components. As such, we cannot accept this change.

PR #1538, opened 2024-11-15
STSMACOM-875 Use `<IfAnyPermission>` and `stripes.hasAnyPerm` to check for Notes assign/unassign perm.

BogdanDenis:
## Description
It seems that permission's `replaces` property doesn't work as it was expected to, so we have to check for both old and new permission names.
Use `<IfAnyPermission>` and `stripes.hasAnyPerm` to check for Notes assign/unassign perm.

## Screenshots

image

## Issues
[STSMACOM-875](https://folio-org.atlassian.net/browse/STSMACOM-875)

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@folio/stripes-cli": "^3.0.0",
     "@folio/stripes-components": "^12.0.0",
     "@folio/stripes-connect": "^9.0.0",
-    "@folio/stripes-core": "^10.0.0",
+    "@folio/stripes-core": "^10.2.3",
     "@folio/stripes-final-form": "^8.0.0",
     "@folio/stripes-form": "^9.0.0",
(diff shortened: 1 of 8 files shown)


Dmytro-Melnyshyn (approved):

package.json
```diff
@@ -47,7 +47,7 @@
     "@folio/stripes-cli": "^3.0.0",
     "@folio/stripes-components": "^12.0.0",
     "@folio/stripes-connect": "^9.0.0",
-    "@folio/stripes-core": "^10.0.0",
+    "@folio/stripes-core": "^10.2.0",
```
zburke, line 50: `<IfAny>` doesn't show up until `^10.2.3`

zburke (approved):
Code changes LGTM but please make sure to set the minimum stripes-core version correctly in `package.json`: `<IfAny>` requires `^10.2.3`.

PR #1540, opened 2024-11-18
STSMACOM-876 `<EditableListForm>` `state.status` initialize as `null` so it can pass `null` re-initialization checks on component update.

BogdanDenis:
## Description
Uncaught error in `setError` because `state.status` array is empty and we're trying to access an nth element when deleting a record has failed.
This is because `status.empty` is initialized in the constructor as `[]` but the re-initialization check in `getDerivedStateFromProps` checks for `!state.status`, so it's never again re-initialized (unless you add a new field which always re-initializes the state)

## Screenshots

https://github.com/user-attachments/assets/81511ca8-c338-43ad-ba41-a797134ff31c

## Issues
[STSMACOM-876](https://folio-org.atlassian.net/browse/STSMACOM-876)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Improve confirmation modal footer for `ControlledVocab` component. Refs STSMACOM-863.
 * Add the `endDateInputRef` prop to `DateRangeFilter` to access the end date element. Refs STSMACOM-859.
 * Remove unnecessary `aria-rowindex` in `ItemView` and `ItemEdit` components. Fixes STSMACOM-871.
+* Check for `active` status of `<Settings>` navigation links independently from the applied href and query string. Query string should only be applied to active link. Refs STSMACOM-837.
 * `<EditableListForm>` use `lodash.isEmpty` to check for empty `state.status` in re-initialization checks on component update. Fixes STSMACOM-876.
 
 ## [9.2.3] IN PROGRESS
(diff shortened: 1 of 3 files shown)


zburke (approved):
Oh, I hate these old, uncommented, heavily-lifecycle-dependent class components so much. 

I understand your logic: if we initialize `status` to `[]` in the constructor, gDSFP will never rebuild it because `[]` isn't falsey. Fine. But, is initializing to `null` correct since we know it's supposed to be an array, or is there a better fix in gDSFP like `if (isEmpty(this.status))`? 

I'm OK with this change. I just want to be sure it's the best/clearest change we can make that will resolve this bug.

zburke:
PS, @BogdanDenis, please make sure to move the ticket to "awaiting release" when this merges. Technically, this should go into the CHANGELOG under 9.3.0 and merge to master, and then I'll cherry pick it onto the b9.2 release branch.

BogdanDenis:
@zburke That's a good suggestion to use `isEmpty`, I've updated the PR

zburke (commented):
@BogdanDenis will this work correctly if we _only_ include the gDSFP change?

BogdanDenis:
> @BogdanDenis will this work correctly if we _only_ include the gDSFP change?

It does, I've updated the PR

OleksandrHladchenko1 (approved):

PR #1543, opened 2024-11-26
STSMACOM-878 Fetch updaters in `<ViewMetaData>` on `props.metadata` changes.

BogdanDenis:
## Description
`<ViewMetaData>` component may initially mount with `metadata.createdByUserId` and `metadata.updatedByUserId` empty, so it won't make a call to fetch updaters. But when `metadata` props later change - the component should make additional calls to fetch users.

## Screenshots

https://github.com/user-attachments/assets/4a1a1bf5-0509-464f-8142-02396ccf6ea0

## Issues
[STSMACOM-878](https://folio-org.atlassian.net/browse/STSMACOM-878)


zburke (approved):
Thank you for the refactor here. Soon we should take it a step further cut out the manifest, mutator, and resources entirely so we can excise `stripesConnect`.

PR #1546, opened 2024-12-02
STSMACOM-880: `DateRangeFilter` - set the `wrongDatesOrder` to `false` when the field value changes.

Dmytro-Melnyshyn:
## Purpose
DateRangeFilter: Don't display errors after the reset button is hit.

## Description
Previously, the `handleDateChange` function set `wrongDatesOrder` to `false` to reset this error and only compare dates order when the "Apply" button is pressed. 

## Issues
[STSMACOM-880](https://folio-org.atlassian.net/browse/STSMACOM-880)

## Screencasts

https://github.com/user-attachments/assets/b176b6f6-4455-4fc6-a653-9fa7a8af7759


lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
```diff
@@ -169,6 +169,7 @@ const TheComponent = ({
       ...filterValue,
       errors: {
         ...filterValue.errors,
+        wrongDatesOrder: false, // dates order is validated only when the "Apply" button is clicked, not when the field value is changed
```
BogdanDenis, line 172: If previously this error was shown on change and now it doesn't - then this change could affect a lot of other modules and their tests, maybe some QA test cases.
Can we follow the approach from `<DateFilter>` in `stripes-inventory-components` where we listen to reset events and clear inputs by ref?

BogdanDenis (changes requested):

lib/SearchAndSort/components/DateRangeFilter/DateRangeFilter.js
Dmytro-Melnyshyn: @BogdanDenis 
The approach in `stripes-inventory-components` just resets the value, but the state of errors is reset in the `handleDateChange` function. The problem is that the error message related to `wrongDatesOrder` doesn't disappear, and this is due to the recent change in the `handleDateChange` function. It seems that it was changed by accident, because this bug is still fixed. The `wrongDatesOrder` was always applied when the "Apply" button was hit, and there was no requirement to change this behavior.

BogdanDenis (approved):

zburke (approved):

PR #1555, opened 2025-01-15
STSMACOM-886 provide LinkedUser component

zburke:
Link to a user-details record if permissions allow, or return a plaintext name.

Mocks shamelessly copied from ui-users.

Refs [STSMACOM-886](https://folio-org.atlassian.net/browse/STSMACOM-886)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,7 @@
 {
+  "env": {
+    "jest": true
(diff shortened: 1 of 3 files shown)


ryandberger (approved):
Clean up lint, but looks good otherwise!

PR #1558, opened 2025-01-21
STSMACOM-887: Change the request URL limit for `REQUEST_URL_LIMIT` due to increase to 8192.

Dmytro-Melnyshyn:
## Description:
Okapi’s URI max length was increased from 4096 to 8192 in [OKAPI-1198](https://folio-org.atlassian.net/browse/OKAPI-1198). This requires a change to the `REQUEST_URL_LIMIT` variable.

## Issues
[STSMACOM-887](https://folio-org.atlassian.net/browse/STSMACOM-887)


zburke (approved):
It would be nice if this value were tied to the `okapi` interface somehow. We'd have to make this into a hook, but then we could look at the interface version retrieved during discovery and supply different values conditionally.

artem-blazhko (approved):

Dmytro-Melnyshyn:
> It would be nice if this value were tied to the `okapi` interface somehow. We'd have to make this into a hook, but then we could look at the interface version retrieved during discovery and supply different values conditionally.

I agree with you. I suspect they will not change this value in the future, as it's the maximum. They increased it for Okapi to be the same as Kong.

PR #1561, opened 2025-02-06
STSMACOM-893 EntryManager sort by id if sort-key is absent

zburke:
Avoid an NPE when sorting lists if the requested sort-key is missing by using `id` as the key instead. Should we check for `id`? Eh, maybe, but `<EntryManager>` is a tool for handling entries retrieved from API requests so it seems like a safe assumption.

This bug was introduced in 1a8f5450b566c0cf21626760e2ef40de5e63f8c2.

Refs [STSMACOM-893](https://folio-org.atlassian.net/browse/STSMACOM-893)

### DIFF
--- a/lib/EntryManager/EntrySelector.js
+++ b/lib/EntryManager/EntrySelector.js
@@ -22,6 +22,19 @@ import {
 } from '@folio/stripes-components';
 import ConnectedWrapper from './ConnectedWrapper';
 
+/**
+ * listSortByLowerCase
+ * Sort the list by the lower-cased value of key. If the key is not present
+ * on an item, use its `id` property instead.
+ *
+ * @param {Array} list Array of items to sort
+ * @param {string} key property of an object to sort by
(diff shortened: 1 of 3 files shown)


lib/EntryManager/EntrySelector.js
```diff
@@ -22,6 +22,19 @@ import {
...
+/**
+ * listSortByLowerCase
+ * Sort the list by the lower-cased value of key. If the key is not present
+ * on an item, use its `id` property instead.
+ *
+ * @param {Array} list Array of items to sort
+ * @param {string} key property of an object to sort by
+ * @returns list, sorted by the lower-cased value of key
+ */
+export const listSortByLowerCase = (list, key) => {
```
MikeTaylor, line 34: I assume the `export`? is just for the tests?

MikeTaylor (approved):
Looks good!

PR #1573, opened 2025-03-21
STSMACOM-890 Convert test suite to new bigtest interactors.

JohnC-80:
Overview of changes:

* `bigtest/mocha` imports are changed to `mocha` - once all test suites are changed, the `bigtest/mocha` dependency can be eliminated.
  *  `bigtest/mocha` repeatedly calls `it()` calls until they pass or the timeout is reached (converge at the 'it' level)
  *  New bigtest moves converge to interactors themselves... so their usage will converge only on tests where they are in use.
  *  New bigtest assertions can eliminate the need for `chai`, as `exists()`,`absent()`,`is()`, and `has()` assertions are contained in the interactor itself.
* New bigtest interactors for single suites can be contrived per test suite by combining imports from `@folio/stripes-testing`.
* Interaction is common among the Custom Fields suites, so those interactors were consolidated locally. These interactors can be moved to stripes-testing, as other modules that make use of Custom Fields can also use these interactors.
* exchanges `axe` for `runAxeTest` utility from `@folio/stripes-testing`

Note - skipping axe test in custom-fields-form for now - unfortunately the draggable accordions have a glaring a11y issue (nested interactive elements.)

Future work:
* larger test suites/shared interactors will be converted in separate future PR's.
* eventually remove all `bigtest/interactor` interactors from this library entirely. All things should get their necessary interactors from `@folio/stripes-testing`.


zburke (approved):
Wow, a ton of work here and just ticking 'Approve' doesn't do it justice. Thank you for this Herculean effort!

PR #1577, opened 2025-04-30
STSMACOM-902: ConfigManager - Add `userId` property to be able to retrieve user's own settings, not just tenant settings from mod-settings.

Dmytro-Melnyshyn:
## Description
`ConfigManager` - Add `userId` property to be able to retrieve user's own settings, not just tenant settings from mod-settings. This change is made specifically for https://github.com/folio-org/ui-myprofile/pull/251

## Issues
[STSMACOM-902](https://folio-org.atlassian.net/browse/STSMACOM-902)

### DIFF
--- a/lib/ConfigManager/ConfigManager.js
+++ b/lib/ConfigManager/ConfigManager.js
@@ -30,7 +30,7 @@ const getPath = (_q, _p, _r, _l, props) => {
   }
 
   return getSettingsPath({ scope, configName, userId });
-}
+};
 
 class ConfigManager extends React.Component {
   static manifest = Object.freeze({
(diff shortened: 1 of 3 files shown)


BogdanDenis (approved):

JohnC-80 (approved):

lib/ConfigManager/ConfigManager.js
```diff
@@ -13,7 +13,7 @@ class ConfigManager extends React.Component {
       type: 'okapi',
       path: (_q, _p, _r, _l, props) => (props.moduleName ?
         `configurations/entries?query=(module==${props.moduleName} and configName==${props.configName})` :
-        `settings/entries?query=(scope==${props.scope} and key==${props.configName})`
+        `settings/entries?query=(scope==${props.scope} and key==${props.configName}${props.userId ? ` and userId==${props.userId}` : ''})`
```
JohnC-80, line 16: Minor but this template is reaching a point of complexity where I would break them out into separate functions/variables with differentiated naming - `getConfigurationsPath()`... `getSettingsPath()` etc...

PR #1579, opened 2025-05-07
STSMACOM-904: Add `hideEditButton`, `interactive` and `noRowClick` props to `NotesSmartAccordion` components, add a sort icon for the list headers, and remove the padding of the ql-editor container.

Dmytro-Melnyshyn:
## Purpose
1. Be able to display notes without an edit button.
2. Be able to not redirect to note details after hitting a list item.
3. Remove a "pointer" cursor from the list items, since they will not be a link to redirect to note details.
4. Remove the padding of the ql-editor container.
5. Add a sort icon for the list headers.

## Approach
1. Add `hideEditButton` property.
2. Add `noRowClick` property.
3. Add `interactive` property to make it configurable.
4. Increase the specificity in styles to overwrite the built-in ql-editor styles.
5. Set the `showSortIndicator` property.

## Issues
[STSMACOM-904](https://folio-org.atlassian.net/browse/STSMACOM-904)

## Screenshots
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Apply wrappers with `flow()` instead of annotations. Refs STSMACOM-896.
 * `ConfigManager` - Add `userId` property to retrieve the user's own settings from mod-settings. Refs STSMACOM-902.
-* Add `hideEditButton`, `interactive` and `noRowClick` props to `NotesSmartAccordion` components, add a sort icon for the list headers, and remove the padding of the ql-editor container. Refs STSMACOM-904.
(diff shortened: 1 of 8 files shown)


JohnC-80 (approved):

zburke (approved):
It's a bit unfortunate that we already have some props with these `hideX`-style names, because that leads to inconsistency in naming/behavior with props like `interactive` (default: true) vs `hideEditButton` (default: false). As a rule, it's usually more intuitive to turn an "on-switch" off (e.g. `showEditButton: false`) than to turn an "off-switch" on (e.g. `hideEditButton: true`). You don't need to change anything here, but it's something we should all keep in mind as we design APIs in the future.

PR #1584, opened 2025-05-26
STSMACOM-790: Add `configNamePrefix` prop to custom fields components to be able to store section titles separately for different `entityType`s.

Dmytro-Melnyshyn:
## Description
When multiple instances of the `EditCustomFieldsSettings`/ `EditCustomFieldsRecord`/ `ViewCustomFieldsRecord`/ `ViewCustomFieldsSettings` UI component, each associated with a distinct `entityType`, are used within a single application, then modifying the accordion label in one component results in an unintended change to the label in all other components. This behavior contradicts the expected functionality, causing inconsistency in the labeling of accordion sections based on entityTypes.

The issue arises due to the request sent to the configurations API to store the accordion label, which utilizes a single `configName` for all entityTypes.

Using a single `configName` for multiple entityTypes, leads to a universal change in the accordion label across all instances of the `EditCustomFieldsSettings` components, regardless of their associated entityTypes.

## Approach
Use an optional `configNamePrefix` property to extend `configName`.

## Issues
[STSMACOM-790](https://folio-org.atlassian.net/browse/STSMACOM-790)

## Screencasts

https://github.com/user-attachments/assets/fa6adc9f-cb8d-4b70-83b1-ac87f53aee48

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Pass props for rendering aside content for `<EditCustomFieldsRecord>`'s accordion. Refs STSMACOM-903.
 * Add `hideEditButton`, `interactive` and `canClickRow` props to `NotesSmartAccordion` components, add a sort icon for the list headers, and remove the padding of the ql-editor container. Refs STSMACOM-904.
 * Update button label for creating new notes in NotesAccordion. Refs STSMACOM-906.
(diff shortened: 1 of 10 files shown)


zburke (approved):
LGTM. Thanks for the detailed breakdown of the problem and how this change addresses it.

PR #1586, opened 2025-06-20
STSMACOM-910: Migrate custom fields from mod-configuration to mod-settings.

Dmytro-Melnyshyn:
## Description
Provided a `scope` property, which specifies that the `mod-settings` API should be used, otherwise `mod-configuration` will be used.

## Issues
[STSMACOM-910](https://folio-org.atlassian.net/browse/STSMACOM-910)


zburke (approved):
Backwards compatibility and tests 😍 , though I'm not in love with passing `moduleName` to `useSectionTitleFetch` and then ignoring it when `scope` is defined. Maybe this is the price we pay for the compatibility? Or an indication/implication that we should always do `useHook({ props })` instead of `useHook(av0, av1, ...avN)`. 

Lint is cranky about quotes and whitespace. Please fix this in new code before you merge.

usavkov-epam (approved):

PR #1591, opened 2025-07-03
STSMACOM-914: Settings Custom Fields View - include the `Display in accordion`, `Hidden` fields, and re-order the fields.

Dmytro-Melnyshyn:
## Description
Settings Custom Fields View:
- added new property `hasDisplayInAccordionField` to show/hide the new `Display in accordion` field, which is designed to store the name/id of the accordion where the custom field will be placed. Currently, BE is done only for the mod-users, mod-orders, or any other module should do some extra work to use this field;
- help text field is now always displayed instead of when value is present;
- changed fields order to `Field label`, `Help text`, `Display in accordion`, `Hidden`, `Required`;
- checkbox doesn't have `Required` field;
- created new `DisplayInAccordion` component for `Display in accordion` field;
- created new `HiddenValue` component;
- added new `displayInAccordionOptions` property, it is a list of options `[{ value, label }]`, `value` is used to find the correct `label` and display it for the `Display in accordion` field;
- the initial `Custom fields` option label in the `Display in accordion` field will be the same as section title.

## Issues
[STSMACOM-914](https://folio-org.atlassian.net/browse/STSMACOM-914)

## Screenshots

https://github.com/user-attachments/assets/ab9fb6b7-29d0-4d57-abc0-532b85fe729e

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Migrate custom fields from mod-configuration to mod-settings. Fixes STSMACOM-910.
 * Migrate tags from mod-configuration to mod-settings. Refs STSMACOM-911.
 * `<ControlledVocab>` - added a new `onCreateFail` prop to allow parent components to specify custom error handling. Refs STSMACOM-913.
+* Settings Custom Fields View - include the "Display in accordion", "Hidden" fields, and re-order the fields. Refs STSMACOM-914.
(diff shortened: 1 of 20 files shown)


Dmytro-Melnyshyn:
@zburke I added a description and made the `DisplayInAccordion` field optional.

PR #1597, opened 2025-07-25
STSMACOM-921: `ViewCustomFieldsRecord` - add `sectionId` prop so that not all custom fields are displayed with and without accordion.

Dmytro-Melnyshyn:
## Description
Currently, the "Custom fields" accordion displays all custom fields in `ViewCustomFieldsRecord`. It should also be possible to display custom fields based on the `displayInAccordion` field. Adding the `sectionId` prop will filter out custom fields that don't match the sectionId. It should also be possible to support displaying fields without an accordion, as in some cases custom fields need to be displayed in existing accordions.

## Issues
[STSMACOM-921](https://folio-org.atlassian.net/browse/STSMACOM-921)

## Screencasts

https://github.com/user-attachments/assets/1a7d1245-270c-408b-ae86-b9ea556c2e37


zburke (approved):
LGTM. Thank you for the backwards-compatible implementation that preserves the behavior of existing impls which lack `sectionId`. 🤝

PR #1626, opened 2026-03-18
STSMACOM-946: Add a default empty object for `customFieldsValues` so that `ViewCustomFieldsRecord` can render `NoValue` when no values are provided.

Dmytro-Melnyshyn:
## Description
When `customFieldsValues` is undefined, the `ViewCustomFieldsRecord` component crashes, so we always need to pass an empty object. This PR adds a default empty object by default for `customFieldsValues`, so we no longer need to pass an empty object in all places where `ViewCustomFieldsRecord` is used.

## Issues
https://folio-org.atlassian.net/browse/STSMACOM-946


zburke (approved):
Hard to believe we didn't encounter this until now! 

Adding a default value for the prop looks like the smallest fix, but this kind of thing is often handled with optional chaining instead. Do you think there would be value in defining a pattern/BP for optional properties? Now we'll have code like this:
```
const foo = someProp.optionalField['someKey'] // no NPEs!
if (_.isEmpty(someProp.someValue))            // Oh, hi lodash
```
instead of this:
```
const foo = someProp.optionalField?.['someKey'] // also no NPEs
if (!someProp.someValue)
```
Both are fine. Default value means a small change in once place, but checking for "emptiness" is more involved. Optional chaining means null checking is easy, but is required every place where the optional property is accessed. 

To be clear, I'm approving this PR as-is. I'm just curious if you think there's value in consistently choosing a specific implementation.

mariia-aloshyna (approved):

PR #1628, opened 2026-03-25
STSMACOM-948: Improve ProxyForm layout and validation logic

Dmitriy-Litvinenko:
## Purpose
Improve ProxyForm layout and validation logic

## Approach
- Problem with wrong radio buttons disabling was fixed (part of business logic that associated with proxy/sponsor relationship).
- Improved business logic for enabling/disabling the "Continue" button.
- Updated layout: shifted the "Continue" button to the right.
- Enhanced styling and error display for disabled radio buttons.
- Minor code improvement.
- Just for information. We've identified an opportunity to improve the business logic of the current component. In the future, we plan to improve the business logic and refactor the code (as part of a separate task https://folio-org.atlassian.net/browse/STSMACOM-949). Now we have 0 code coverage and we planned to resolve this problem in future.

# Refs
https://issues.folio.org/browse/STSMACOM-948

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 * `<DateRangeFilter>` - set only the start date field as a focus ref. Fixes STSMACOM-944.
 * Add a limit of 100 for tenant rules in the `PasswordValidationField`. Fixes STSMACOM-942.
 * Add a default empty object for `customFieldsValues` so that `ViewCustomFieldsRecord` can render `NoValue` when no values are provided. Refs STSMACOM-946.
+* Improve ProxyForm layout and validation logic. Refs STSMACOM-948.
 
 ## [10.0.0](https://github.com/folio-org/stripes-smart-components/tree/v10.0.0) (2025-02-24)
 [Full Changelog](https://github.com/folio-org/stripes-smart-components/compare/v9.2.0...v10.0.0)
--- a/lib/ProxyManager/ProxyForm.css
+++ b/lib/ProxyManager/ProxyForm.css
@@ -4,10 +4,6 @@
   margin-left: 6px;
 }
 
-input[disabled] {
-  margin-left: 5px;
-}
-
 .error {
   margin-top: 1px;
   margin-left: 24px;
--- a/lib/ProxyManager/ProxyForm.js
+++ b/lib/ProxyManager/ProxyForm.js
@@ -4,7 +4,7 @@ import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import { Field, reduxForm } from 'redux-form';
 import { Button, Col, RadioButton, Row } from '@folio/stripes-components';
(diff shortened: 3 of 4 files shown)


zburke (changes requested):
I'm quite confused here. The ticket describes a bug, not a task, but there is no description here in the PR of how these changes correct the bug. I accept that the video shows a bug (the proxy is expired, but the "Who are you acting as?" dialog displays "Sponsor user record expired") but I cannot accept this code that turns logic on its head:
```
const isProxyUserExpired = isSponsorExpired(patron);
const allRadiosDisabled = isProxyUserExpired && allProxiesDisabled;
```
These are complete non-sequiturs. The names of the functions involved here are a clue that something is more deeply wrong with existing logic or data or labels. Things to explore:

1. Are `isProxyExpired` and `isSponsorExpired` written incorrectly? 
2. Is the problem that `renderDisabledRadio()` is being called incorrectly, e.g. with the first two arguments flipped? 
3. The constructor sets `proxiesMap` and `sponsorsMap`; are they correct?

I think the answer is (2) and the problem is misleading labels:
```
// current labels
renderDisabledRadio(proxy, sponsor, proxyMap)
// correct labels
renderDisabledRadio(sponsor, currentPatron, proxyMap)
```

Maybe we just need to change some variable names and some labeling, e.g. instead of "Who are you acting as?" use "Who is the borrower acting as?", instead of `isProxyUserExpired` and `isSponsorExpired(patron)` we need `isPatronExpired(patron)` and `isPatronExpired(patron)` etc.

Dmitriy-Litvinenko:
@zburke 
The pull request description has been updated. 
Some variables and functions have been renamed to clarify any misunderstandings.
Could you please review?

mariia-aloshyna (approved):

zburke (approved):
The bugs are fixed 😓 so I'm approving this, but I think there is some room for improvement in the code, even if this won't change how things work in the UI. 

I'm still struggling a bit with the naming here. If we look at the dialog, what we see is
> Patron is acting as
> [] Self
> Or acting as proxy for
> [] sponsor A
> [] sponsor B

I think a clear way to name the arguments to `renderDisabledRadio` would be to rename them from `(proxy, patron, proxiesMap)` to `(sponsor, patron, proxiesMap)`. 

Aside, that function should be lifted out of the class entirely since it's pure, or it just should access `this.proxiesMap` directly instead of receiving it as an argument. That's not your fault though; it's always been like that. 

If ProxyForm no longer needs its `sponsorOf` prop, should `ProxyManager` stop drilling `sponsorOf` through `ProxyModal`? 

The PR description "Problem with wrong radio buttons disabling was fixed (part of business logic that associated with proxy/sponsor relationship)" still does not describe the code change. Describe the problem with the radio buttons. Describe how you fixed it. Describe the business logic error around proxy and sponsor relationships.

Dmitriy-Litvinenko:
Thank you @zburke i added feedback to https://folio-org.atlassian.net/browse/STSMACOM-949

PR #1633, opened 2026-06-02
STSMACOM-953 test tightening for ProfilePicure, ViewMetaData

zburke:
* Correctly coerce `enabled` to a boolean as required by react-query
* Supply `/users` as a mirage endpoint to assist tests

`enabled` was the main thing here, causing related to tests to fail reliably such that they were disabled with the comment "works locally". They never ran locally for me. All of this came to a head when trying to publish a release branch with dependencies installed against production releases. The same tests ran fine when built against npm-folioci, which doesn't make any sense to me, but there it is. Anyway. Now they run against CI, and they run against prod. 

Refs [STSMACOM-953](https://folio-org.atlassian.net/browse/STSMACOM-953)

### DIFF
--- a/lib/ProfilePicture/tests/ProfilePicture-test.js
+++ b/lib/ProfilePicture/tests/ProfilePicture-test.js
@@ -1,6 +1,5 @@
 import React from 'react';
-import { describe, beforeEach, afterEach, it } from 'mocha';
-import sinon from 'sinon';
+import { describe, beforeEach, it } from 'mocha';
 import {
   HTML,
   runAxeTest
@@ -9,7 +8,6 @@ import {
 import {
   mount,
   setupApplication,
-  wait,
 } from '../../../tests/helpers';
 
 import ProfilePicture from '../ProfilePicture';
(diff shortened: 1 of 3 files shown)


lib/ProfilePicture/tests/ProfilePicture-test.js
```diff
@@ -30,35 +28,18 @@ describe('ProfilePicture', () => {
...
     beforeEach(async () => {
       await renderComponent({
         profilePictureLink: 'https://url-to/image.jpg'
       });
-      await wait(200);
     });
 
     it('should not have any a11y issues', () => runAxeTest());
 
     // This test fails in CI, as the element may be present, but not visible.
```
Dmitriy-Litvinenko: Could you please clarify do we need this comment?

lib/ProfilePicture/tests/ProfilePicture-test.js
```diff
@@ -30,35 +28,18 @@ describe('ProfilePicture', () => {
...
     // It's dependably visible locally.
```
Dmitriy-Litvinenko: Could you please clarify do we need this comment?

Dmitriy-Litvinenko (approved):

PR #1637, opened 2026-07-08
STSMACOM-957 Handle undefined response in processBadResponse

MikeTaylor:
Add guard clause to handle undefined response in processBadResponse function.

I ran across this while developing ui-cyclops, and running against a server that had not yet had the new permissions inserted. Obviously that's a bit of an edge-case as it's unlikely to happen on a well-managed production system. But I do think that no permissions state should ever cause an actual crash.

New jest tests cover this condition; existing BT suite allowed it to slip through.

Refs [STSMACOM-957](https://folio-org.atlassian.net/browse/STSMACOM-957)

### DIFF
--- a/lib/EditableList/processBadResponse.js
+++ b/lib/EditableList/processBadResponse.js
@@ -23,7 +23,7 @@ const getErrorMessage = (errorCode, values = {}) => {
 const handleError = (errorMessages, message, values) => {
   const updatedErrorMessages = { ...errorMessages };
 
-  if (values && values.fieldLabel) {
+  if (values?.fieldLabel) {
     if (updatedErrorMessages.fieldErrors[values.fieldLabel]) {
       updatedErrorMessages.fieldErrors[values.fieldLabel].push(message);
     } else {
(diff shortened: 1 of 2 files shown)


lib/EditableList/processBadResponse.js
```diff
@@ -68,6 +68,16 @@ const getDefaultErrorMessage = (defaultErrorCode) => {
...
 export default async function processBadResponse(response, defaultErrorCode, getCustomErrorMessages = null) {
+  // ControlledVocab rejects its delete promise with no
+  // argument when the DELETE fails (e.g. a 403), so `response` arrives here as
+  // undefined and `response.status` throws, crashing the whole app instead of
+  // surfacing the failure. Guard against a non-Response reason and fall back to
+  // the default error message. See stripes-smart-components ControlledVocab
+  // onDeleteItem -> deleteItemReject().
+  if (!response || typeof response.status !== 'number') {
+    return getDefaultErrorMessage(defaultErrorCode);
+  }
```
zburke: Alternatively, instead of adding new condition we change existing conditions that inspect `response.status`:
```
if (response.status === 422) {  // throws if response is undefined
if (response?.status === 422) { // keeps calm and carries on
```
or change the overall structure to only handle the default in one place:
```
if (response?.status) {
  if (response.status === 422) ...
  if (response.status === 409) ...
}
return getDefaultErrorMessage(defaultErrorCode);
```

zburke (approved):
Thanks for this. I appreciate the code-comments that explain why this error handling is necessary, and I *really appreciate that you took the time to fix it rather than just flagging it and leaving it for somebody else to deal with. 

The current implementation resolves the problem, but now handles the default/unknown-status case in two separate places. Consider the alternatives that only handle it once. 

I filed a corresponding ticket (assigned to you) and added it to the PR title. I hope this doesn't feel meddlesome. It provides visibility of this change for POs and makes it easier for them to track which release contains a given bug-fix.

lib/EditableList/processBadResponse.js
MikeTaylor: I'm happy with any of these approaches.

zburke: Do you want to pick your favorite and merge it, or do you want somebody else to pick up this work and merge it? I trust your judgement and have approved the PR to avoid blocking your progress.

MikeTaylor: Thanks. I think I'd prefer you to make the change you like best. It's not a showstopper for me, so if you need another day or two to think it through, that won't slow me down.

PR #1641, opened 2026-08-11
STSMACOM-962: Fixed the issue where the Last updated field erroneously displays Unknown user

Dmitriy-Litvinenko:
## Purpose
Fixed the issue where the Last updated field erroneously displays Unknown user- #1641

# Approach 
The backend has a default limit of 10 records. This commit overrides that default by introducing a constant on the frontend and passing it to the GET request and component manifest, aligning it with the required data length of 2000.

# Refs
https://issues.folio.org/browse/STSMACOM-962


zburke (approved):
Code changes LGTM 👌 . PR title should describe the solution, not the problem, e.g. "Supply missing limit param to updaters query". 

As a reviewer, I want to know "How does the code work now that this PR has been merged?" Sure, it is helpful to provide details about the bug/problem in the description and how the new code mitigates it, but the emphasis should be on describing the correctly-working code that is present now rather than the buggy code that has been removed.

vashjs (approved):

PR #1642, opened 2026-08-12
STSMACOM-961: EditCustomFieldsRecord - apply word-breaking styles to all custom field types including RadioButtonGroup labels, Select/MultiSelect options, Checkbox labels, and validation messages

Dmitriy-Litvinenko:
## Purpose
EditCustomFieldsRecord - apply word-breaking styles to all custom field types including RadioButtonGroup labels, Select/MultiSelect options, Checkbox labels, and validation messages

# Approach 
image

# Refs
https://issues.folio.org/browse/STSMACOM-961

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * `ViewCustomFieldsRecord` - make `calloutRef` available after the component is mounted to prevent a page crash when the section title fetch fails. Fixes STSMACOM-929.
 * `SearchAndSort` - Add pass through `actionMenuToggleProps` for `Pane`. Refs STSMACOM-959.
 * Update user addresses view to support empty state. Refs STSMACOM-960.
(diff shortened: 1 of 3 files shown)


zburke (changes requested):
Replace `word-break: break-word` with `word-break: normal; overflow-wrap: anywhere`. `word-break: break-word` [is deprecated](https://drafts.csswg.org/css-text/#word-break-property): 
> For compatibility with legacy content, the [`word-break`](https://drafts.csswg.org/css-text/#propdef-word-break) property also supports a deprecated `break-word` keyword. When specified, this has the same effect as `word-break: normal` and [`overflow-wrap: anywhere`](https://drafts.csswg.org/css-text/#propdef-overflow-wrap), regardless of the actual value of the `overflow-wrap` property.

zburke (approved):

vashjs (approved):

PR #1643, opened 2026-08-20
STSMACOM-963

Dmitriy-Litvinenko:
1. useState(true) → useState({}) — validity is now a map of refId → boolean instead of one shared flag.
  2. checkValid — now checks fieldValidity[customField.refId] === false (strictly false), so a required field with
  undefined validity (never changed) or true validity never shows the error.
  3. OnChange handler — now writes { [customField.refId]: values.length > 0 } into the map, leaving all other fields'
  validity untouched.


lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
```diff
@@ -322,11 +322,10 @@ const EditCustomFieldsRecord = ({
...
-                setIsValid(false);
-              } else {
-                setIsValid(true);
-              }
+              setFieldValidity({
+                ...fieldValidity,
+                [customField.refId]: values.length > 0,
+              });
             }}
           </OnChange>
```
Dmytro-Melnyshyn: Have you tried removing the `OnChange` code snippet? Why isn't `<Field required />` sufficient?

Dmitriy-Litvinenko: It's a good suggestion, but as I understand it, it won't work.

`<Field required />` in React Final Form doesn't add validation automatically — it just passes required as a prop to the underlying component. To actually validate, you'd need an explicit validate function on the `<Field>`, plus handle the fact that an empty array [] is truthy, so native HTML constraint validation won't flag it.

Additionally, the warning is rendered outside RFF's meta.error flow — checkValid reads local component state and passes the result directly as a prop, so there's no path for RFF's built-in validation to feed this display logic.

A validate-based approach would be cleaner and would also catch the case where the user submits without ever touching the field (the current OnChange approach only fires on interaction). That said, this changes keeps the change minimal and targeted, which is the right call for a bugfix.

Dmytro-Melnyshyn: The `<Field>` for `MULTISELECT` already gets `validate={validateCustomFields(customField)}`. What does `validateCustomFields` return today when a required multiselect's value is `[]`?

Dmitriy-Litvinenko: EditCustomFieldsRecord maintained a manual validation path for multiselect fields alongside the standard form validation that was already in place. This parallel path was both redundant and the source of the false "Field is required" warning reported in STSMACOM-963.

Why it is redundant:

validate={validateCustomFields(customField)} already handles the required-and-empty-array case correctly — validateCustomFields explicitly checks Array.isArray(value) ? !value.length : !value before applying the required constraint. There was never a need for a second path.

Changes:

- Remove onBlur: e => e.preventDefault() so multiselect fields are marked touched on blur and meta.error from validate flows through normally.
- Remove <OnChange>, fieldValidity state, checkValid, and getMultiSelectProps — the entire manual validation path is now unnecessary.
- required={customField.required} and validate={validateCustomFields(customField)} on <Field> are sufficient; they were already present and are now the single source of truth for multiselect validation.

Dmytro-Melnyshyn: Looks good, this is the direction I was hoping for. `react-final-form-listeners` is now unused, please drop it from `package.json` and from `yarn.lock`, and remove `reactFinalFormListeners.mock.js`.

zburke (commented):
Wow, sharp eyes @Dmytro-Melnyshyn, I'm so glad you pressed for a second look here, and thanks @Dmitriy-Litvinenko for your patience. 

The redundant validation path `getMultiSelectProps()` was added in PR #814 to address [UIU-1569](https://folio-org.atlassian.net/browse/UIU-1569). There is not a lot of detail there ("Fix validation for MultiSelect ... Able to view validation message for MultiSelect CustomField"). It's possible improvements in the last six years supply validation messages that did not exist at the time. Still, please check that the scenarios enumerated under UIU-1569 are still satisfied.

lib/CustomFields/pages/EditCustomFieldsRecord/EditCustomFieldsRecord.js
Dmitriy-Litvinenko: Done

Dmytro-Melnyshyn (approved):

